PR #1, opened 2025-02-20
Fibonacci works

keyboardDrummer:
### Changes
- Add CLI tool that can take a Java file with JVerify annotations, and attempt to verify it. Only works for the simplest of Java programs right now.
- Added Dafny as a submodule, and reference the commit from this unmerged PR https://github.com/dafny-lang/dafny/pull/6112

### Testing
- Add a success and a failure test for a fibonacci program
- Add CI workflow that:
  - Checks running the code generation creates no changes
  - Runs the junit tests

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,43 @@
...
+        working-directory: ./jverify-collection/dafny
+        run: |
+          dotnet build ./Source/Dafny.sln
+          make z3-ubuntu
+      - name: Code generation has no diff
+        working-directory: ./jverify-collection
+        run: |
+          make test-generation-nodiff
+      - name: Install Z3
+        run: |
```
alex-chew: I think you meant to write this?

```suggestion
      - name: Build Dafny
        working-directory: ./jverify-collection/dafny
        run: |
          dotnet build ./Source/Dafny.sln
      - name: Install Z3
        working-directory: ./jverify-collection/dafny
        run: |
          make z3-ubuntu
      - name: Code generation has no diff
        working-directory: ./jverify-collection
        run: |
          make test-generation-nodiff
```

.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "jverify/dafny"]
+	path = jverify/dafny
+	url = git@github.com:dafny-lang/dafny.git
```
alex-chew: Is this left over from testing?

build.gradle.kts
```diff
@@ -0,0 +1,146 @@
...
+        // Override the jvmArgs to remove unwanted options
+        jvmArgs = allUnnamedArgs
+
+        // Keep the standard I/O settings
+        standardInput = System.`in`
+        standardOutput = System.out
+    }
+}
+
+//dependencies {
```
alex-chew: nit: are these still needed?

library/src/main/java/com/aws/jverify/Erased.java
```diff
@@ -0,0 +1,11 @@
+package com.aws.jverify;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.SOURCE)  // Keep annotation at runtime
```
alex-chew: Did you mean to use `RetentionPolicy.RUNTIME` here or is the comment wrong? (And same with the other `@interface`s)

library/src/main/java/com/aws/jverify/Erased.java
```diff
@@ -0,0 +1,11 @@
...
+@Target(ElementType.METHOD)            // Can only be used on classes/interfaces
```
alex-chew, line 9: Also the comments on some of these `@Target`s appear inaccurate.

library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -0,0 +1,22 @@
+package com.aws.jverify;
+
+import java.util.function.Function;
+
+public class JVerify {
+    public static void check(boolean condition) {
+    }
+
+    public static <T> void precondition(boolean condition) {
```
alex-chew: nit: type parameter not needed?
```suggestion
    public static void precondition(boolean condition) {
```

library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    public static void invariant(boolean condition) {
+    }
+    
+    public static <T> void postcondition(Function<T, Boolean> predicate) {
+    }
+
+    public static void postcondition(Boolean predicate) {
```
alex-chew: nit: can this parameter can be primitive?

```suggestion
    public static void postcondition(boolean predicate) {
```

javaTypesGenerator/src/main/java/com/aws/jverify/generator/CSharpToJavaConverter.java
```diff
@@ -0,0 +1,420 @@
...
+    }
+
+    public List<CSharpClass> parseCSharpClasses(String csharpCode) {
+        List<CSharpClass> classes = new ArrayList<>();
+        Map<String, CSharpClass> classMap = new HashMap<>();
+
+        // Updated pattern for class declaration with generics and constraints
+        // Example: "class Specification<T> : NodeWithComputedRange where T : Node"
+        Pattern classWithBodyPattern = Pattern.compile(
+                "(abstract)?\\s*class\\s+(\\w+)\\s*(?:<([^>]+)>)?\\s*(?::\\s*(\\w+))?\\s*(?:where\\s+([^{]+))?\\s*\\{([^}]*)}",
```
alex-chew: nit: it'd be nice to split this up, and/or name the capture groups, for legibility

javaTypesGenerator/src/main/java/com/aws/jverify/generator/CSharpToJavaConverter.java
```diff
@@ -0,0 +1,420 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.*;
+import java.util.regex.*;
+import java.util.stream.Collectors;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class CSharpToJavaConverter {
+    Set<String> foundTypes = new HashSet<String>();
```
alex-chew: Is this supposed to contain the names of only parsed classes, or those of parsed enums too? Right now only the former appears to be implemented, but the variable name suggests to me that perhaps both were intended.

javaTypesGenerator/src/main/java/com/aws/jverify/generator/CSharpToJavaConverter.java
```diff
@@ -0,0 +1,420 @@
...
+        if (typeParam.isPresent()) {
+            // If it's a type parameter with a constraint, return the bounded type
+            if (typeParam.get().constraint != null) {
+                return WildcardTypeName.subtypeOf(ClassName.get("", typeParam.get().constraint));
+            }
+            // If it's an unbounded type parameter, return it as is
+            return TypeVariableName.get(field.type);
+        }
+
+        String baseType = convertCSharpTypeToJava(field.type);
```
alex-chew, line 174: nit: It took me a moment to realize that `convertCSharpTypeToJava` and `convertCSharpTypeToJavaType` are distinct; consider renaming one or both and/or documenting the intent of each

javaTypesGenerator/src/main/java/com/aws/jverify/generator/CSharpToJavaConverter.java
```diff
@@ -0,0 +1,420 @@
...
+        for (CSharpClass csharpClass : classes) {
+            if (csharpClass.parentClass != null) {
+                csharpClass.parentClassRef = classMap.get(csharpClass.parentClass);
+            }
+        }
+
+        return classes;
+    }
+
+    private TypeName convertCSharpTypeToJavaType(CSharpField field, List<TypeParameter> typeParameters) {
```
alex-chew: nit: Consider renaming `typeParameters` to `classTypeParameters` (or otherwise clarifying that they're not the *field*'s type parameters, as I initially thought)

javaTypesGenerator/src/main/java/com/aws/jverify/generator/CSharpToJavaConverter.java
```diff
@@ -0,0 +1,420 @@
...
+
+        // Get all fields including inherited ones
+        List<CSharpField> allFields = csharpClass.getAllFields();
+
+        // Add fields and prepare constructor parameters
+        for (CSharpField field : allFields) {
+            int iteration = 0;
+            for(;iteration < 5;iteration++) {
+                var newName = field.name + (iteration == 0 ? "" : iteration);
+                try {
```
alex-chew, line 237: Why is this needed?

javaTypesGenerator/src/main/java/com/aws/jverify/generator/CSharpToJavaConverter.java
```diff
@@ -0,0 +1,420 @@
...
+                                fieldType,
+                                newName,
+                                Modifier.PRIVATE, Modifier.FINAL);
+                        if (field.isNullable) {
+                            builder.addAnnotation(Nullable.class);
+                        }
+                        FieldSpec fieldSpec = builder.build();
+                        classBuilder.addField(fieldSpec);
+                    } else {
+                        var b = 3;
```
alex-chew: What's this for?

javaTypesGenerator/src/main/java/com/aws/jverify/generator/CSharpToJavaConverter.java
```diff
@@ -0,0 +1,420 @@
...
+                                .addStatement("return this.$N", newName)
+                                .build();
+                        classBuilder.addMethod(getter);
+                    }
+                    break;
+                } catch(IllegalArgumentException ignored) {
+                }
+            }
+            if (iteration == 5) {
+                throw new RuntimeException("failed");
```
alex-chew: nit: It'd be helpful to add more context to the message

verifier/src/test/java/com/aws/jverify/TestVerifier.java
```diff
@@ -0,0 +1,74 @@
...
+                "\n" +
+                "Dafny program verifier finished with 4 verified, 4 errors\n", output);
+        Assertions.assertEquals(4, exitCode);
+
+    }
+    
+    @Test
+    public void testRunThroughGradle() throws IOException, InterruptedException {
+        var process = new ProcessBuilder(
+                "../gradlew",
```
alex-chew, line 62: Just a heads up - on several occasions I've run into cache/locking errors when Gradle gets invoked in a "reentrant" manner as you're doing here (i.e. using `gradlew` to run a test from the command line, which in turn executes further `gradlew` commands), along the lines of https://github.com/gradle/gradle/issues/8750. Disabling the Gradle daemon globally will usually work around the issue, so this isn't a blocker, but it'd be nice to avoid running into that situation in the first place.

verifier/src/main/java/com/aws/jverify/verifier/Serializer.java
```diff
@@ -0,0 +1,130 @@
...
+    private void serializeValue(Object value, AnnotatedType annotatedType) {
+        var isNullable = annotatedType.getAnnotation(Nullable.class) != null;
+        if (isNullable) {
+            encoder.writeNullable(value == null);
+            if (value == null) {
+                return;
+            }
+        }
+        if (value == null) {
+            throw new RuntimeException();
```
alex-chew: nit: Consider adding more context to the empty `RuntimeException`s

verifier/src/main/java/com/aws/jverify/verifier/Main.java
```diff
@@ -0,0 +1,109 @@
...
+@Command(name = "verify", description = "Verify Java code that calls the JVerify library")
+class AppCommand implements Callable<Integer> {
+    @Parameters(description = "Java files to verify")
+    private List<Path> inputs;
+    
+    @Option(names = "--print-binary-dafny", description = "Print the binary Dafny code that is generated from Java")
+    private Path printBinaryDafny;
+
+    @Option(names = "--print-dafny", description = "Print the Dafny code that is generated from Java")
+    private Path printDafny;
```
alex-chew: nit: The descriptions should mention that the arguments' values are expected to be paths, and should also state whether each path should be a file or directory

verifier/src/main/java/com/aws/jverify/verifier/Main.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Option(names = "--dafny", description = "Location of the Dafny CLI to use. Overrides environment variable JVERIFY_DAFNY.")
+    private Path dafny;
+    
+    @Override
+    public Integer call() throws IOException {
+        Writer writer = new OutputStreamWriter(System.out);
+
+        Class<?> clazz = JVerify.class;
+        String location = URI.create(getJarLocationForClass(clazz)).getPath();
```
alex-chew: Can you explain what this bit is doing?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -0,0 +1,488 @@
...
+        List<FileStart> filesStarts = new ArrayList<>();
+        var parsed = task.parse();
+        task.analyze();
+        
+        for (var compilationUnit : parsed) {
+            this.compilationUnit = (JCTree.JCCompilationUnit)compilationUnit;
+
+            if (compilationUnit.getSourceFile().getName().contains("JVerify.java")) {
+                continue;
+            }
```
alex-chew: nit: This seems fragile but it's not a big deal right now

alex-chew (commented):
I've reviewed everything except `JavaToDafnyCompiler.java`; I'll follow up with a review of that file first thing in the morning.

.gitmodules
keyboardDrummer: Don't know why it's there. Removed it.

library/src/main/java/com/aws/jverify/Erased.java
keyboardDrummer: I think source is fine, so I've removed the confusing comments.

javaTypesGenerator/src/main/java/com/aws/jverify/generator/CSharpToJavaConverter.java
keyboardDrummer: Added a comment. It's for avoiding reserved names.

javaTypesGenerator/src/main/java/com/aws/jverify/generator/CSharpToJavaConverter.java
keyboardDrummer: Debugging 😅

verifier/src/main/java/com/aws/jverify/verifier/Main.java
keyboardDrummer: Added a comment and improved the variable names

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: Removed it. Was leftover from an earlier setup

javaTypesGenerator/src/main/java/com/aws/jverify/generator/CSharpToJavaConverter.java
```diff
@@ -138,7 +138,13 @@ public List<CSharpClass> parseCSharpClasses(String csharpCode) {
...
         // Example: "class Specification<T> : NodeWithComputedRange where T : Node"
         Pattern classWithBodyPattern = Pattern.compile(
-                "(abstract)?\\s*class\\s+(\\w+)\\s*(?:<([^>]+)>)?\\s*(?::\\s*(\\w+))?\\s*(?:where\\s+([^{]+))?\\s*\\{([^}]*)}",
+                "(abstract)?\\s*" +
+                        "class\\s+" +
+                        "(?<name>\\w+)\\s*" +
+                        "(?:<(?<typeParams>[^>]+)>)?\\s*" +
+                        "(?::\\s*(?<baseClass>\\w+))?\\s*" +
+                        "(?:where\\s+(?<constraints>[^{]+))?\\s*" +
+                        "\\{(?<body>[^}]*)}",
```
alex-chew: Thanks for splitting these up! It would be good to use the capture groups' names when retrieving their values too, but it's not a blocker.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -0,0 +1,485 @@
...
+import java.nio.file.Files;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.function.Function;
+
+public class JavaToDafnyCompiler {
+    public static final String JVERIFY_CLASS = "com.aws.jverify.JVerify";
```
alex-chew: nit:

```suggestion
    public static final String JVERIFY_CLASS = com.aws.jverify.JVerify.class.getName();
```

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
alex-chew, line 1: The `com.sun.tools.javac` package documentation says:

>This package provides a legacy entry point for the javac tool. See the [jdk.compiler](https://docs.oracle.com/en/java/javase/21/docs/api/jdk.compiler/module-summary.html) module for details on replacement APIs.

and many of its members carry the warning:

> This is NOT part of any supported API. If you write code that depends on this, you do so at your own risk. This code and its internal interfaces are subject to change or deletion without notice.

I see that some of the `jdk.compiler` module is imported (i.e. `com.sun.source.tree.*`); do the other `com.sun.tools.javac` imports not have equivalents in the newer APIs? I don't think it's a big deal for this first pass but it'd be good to minimize reliance on unstable APIs.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -0,0 +1,485 @@
...
+            var annotationsByName = annotations.stream().collect(Collectors.toMap(
+                    (JCTree.JCAnnotation a) -> a.getAnnotationType().toString(),
+                    a -> a));
+            
+            var isStatic = (method.getModifiers().flags & Flags.STATIC) == Flags.STATIC;
+            var returnType = toType(method.getReturnType());
+
+            List<Formal> ins = method.getParameters().map(jvd -> 
+                    new Formal(toOrigin(jvd), name(jvd, jvd.getName()), toType(jvd.getType()), false, true,
+            null, null, false, false, false, null));
```
alex-chew, line 124: Generating builders for the generated Dafny AST would greatly improve legibility of constructs like this throughout this file, though I don't think we should tackle that right away.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -0,0 +1,485 @@
...
+            
+            if (annotationsByName.containsKey("Pure")) {
```
alex-chew: nit: It'd be good to refer to annotation names via `.class.getSimpleName()` in case the names change (here and elsewhere). Not a blocker.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -0,0 +1,485 @@
...
+            throw new NotImplementedException("decreases");
+        }
+        return false;
+    }
+
+    private static boolean fromJVerify(Symbol.MethodSymbol methodSymbol) {
+        return methodSymbol.getEnclosingElement().getQualifiedName().contentEquals(JVERIFY_CLASS);
+    }
+
+    private Expression translateExpression(JCTree.JCExpression expression) {
```
alex-chew: Why is this distinct from `toExpr(JCExpression)`?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -0,0 +1,485 @@
...
+        throw new NotImplementedException(member.getClass().getName());
+    }
+
+    private List<Statement> processStatementWithHeader(JCTree.JCStatement statement,
+                                                        Function<JCTree.JCStatement, Boolean> checkPossibleHeader) {
+        var statements = statement instanceof JCTree.JCBlock block ? block.stats : List.of(statement);
+        return processStatementsWithHeader(statements, checkPossibleHeader);
+    }
+    
+    private List<Statement> processStatementsWithHeader(List<JCTree.JCStatement> stats,
```
alex-chew: It looks like if "header" statements appear after the first non-header statement, they'll be silently ignored by `translateStatement` - we should consider emitting a warning to avoid user confusion.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: We're using the internal APIs because the external APIs do not provide the features we need. Lombok does the same. TBH I don't expect the APIs we're using to change in any significant way, so I'm not worried about this. The only thing is that Java's module system makes it a bit more complicated to use these APIs. We have to pass arguments to the JVM to make it possible.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: If you look at this in an IDE then it can show the names of all the passed arguments. Is that not more readable than a builder would be?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: Removed

verifier/src/test/java/com/aws/jverify/TestVerifier.java
keyboardDrummer: Thanks. Let's look into it when we run into it.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -417,16 +394,23 @@ private Statement translateStatement(JCTree.JCStatement statement) {
...
-            Function<JCTree.JCStatement, Boolean> function =
-                    (JCTree.JCStatement s) -> findWhileHeaderStatement(s, invariants, decreases, modifies);
-            var statements = processStatementWithHeader(whileLoop.body, function);
+            List<Name> returnNames = new ArrayList<>();
+            
+            var statements = processStatementWithHeader(whileLoop.body, preconditions, postconditions,
+                    returnNames, invariants, decreases, modifies);
+            
+            checkEmptyExpressions(preconditions, "preconditions", "loop");
+            checkEmptyExpressions(preconditions, "postconditions", "loop");
```
alex-chew: I think you meant this?

```suggestion
            checkEmptyExpressions(preconditions, "preconditions", "loop");
            checkEmptyExpressions(postconditions, "postconditions", "loop");
```

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -435,51 +419,81 @@ private Statement translateStatement(JCTree.JCStatement statement) {
...
-        var expr = expressionStatement.getExpression();
-        if (!(expr instanceof JCTree.JCMethodInvocation invocation)) {
-            return false;
-        }
+    private void checkEmptyExpressions(List<AttributedExpression> expressions, 
+                                       String typeName, 
+                                       String containerName) {
+      for(var expr : expressions) {
+          throw new RuntimeException(typeName + " are not allowed in a " + containerName);
+      }
```
alex-chew, line 427: tiny nitpick - I feel the intent is clearer this way, what do you think? (I initially thought you might've intended to check something about `expr` until I looked at the call sites.)

```suggestion
      if (!expressions.isEmpty()) {
          throw new RuntimeException(typeName + " are not allowed in a " + containerName);
      }
```

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
alex-chew: Fair enough.

I noticed that IntelliJ IDEA showed import errors for the internal API imports, until I clicked through the IDE's code actions to add the necessary compiler parameters. It looks like it's possible to specify those IDE configuration details in the Gradle file: https://github.com/JetBrains/gradle-idea-ext-plugin/wiki#compiler

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
alex-chew: When viewing the code in an IDE the inlay hints do work well, but I like being able to see such information in GitHub diffs/comments too. It's not a big deal, though.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: In the future I expect we won't throw an exception here but report an error, so then it can report multiple errors, that's why I left the loop here.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: Good tip. Would be super nice to figure this out. Mind if we do that in a follow-up though?

PR #23, opened 2025-03-08
fix: handle source and Dafny paths correctly on Windows

alex-chew:
### What was changed?

* Ensures that paths to the Java files-to-be-verified are transformed without throwing invalid character exceptions on Windows (in particular due to drive letters like `C:`)
* Changes paths to the default `dafny` installation, and to the Gradle wrapper, when running on Windows.

### How has this been tested?


keyboardDrummer (approved):
Amazing, thanks. Should we let a extra CI run test using a Windows machine as well?

alex-chew:
> Amazing, thanks. Should we let a extra CI run test using a Windows machine as well?

Yes, that's a good idea. I'll add this in a follow-up PR.

PR #24, opened 2025-03-13
Operators

olivier-aws:
### What was changed?
Translates more binary operators from Java to Dafny

### How has this been tested?
Added a unit test. Verified on a set of JML tests that no more unimplemented operators exception are raised.

### DIFF
--- a/verifier/src/test/java/com/aws/jverify/Operators.java
+++ b/verifier/src/test/java/com/aws/jverify/Operators.java
@@ -10,11 +10,25 @@ class Operators {
         check(z==7);
     }
 
+    static void PlusKO() {
+        int x = 3;
+        int y = 4;
+        int z = x + y;
+        check(z==6);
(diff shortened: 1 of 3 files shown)


verifier/src/test/java/com/aws/jverify/Operators.java
```diff
@@ -0,0 +1,65 @@
+package com.aws.jverify;
+
+import static com.aws.jverify.JVerify.*;
+
+class Operators {
+    static void Plus() {
```
keyboardDrummer, line 6: Would be better to have a positive and a negative test for each. Having verification pass is easier than having it fail.

olivier-aws: Done, I added tests with failing verification objectives

keyboardDrummer (approved):

PR #25, opened 2025-03-17
feat: add support for binary search example

keyboardDrummer:
### How has this been tested?


library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -17,6 +18,58 @@ public static <T> void postcondition(Function<T, Boolean> predicate) {
 
     public static void postcondition(boolean predicate) {
     }
+
+    /**
```
keyboardDrummer, line 22: Nice that you've added comments here. Seems good to do that from the start.

library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -17,6 +18,58 @@ public static <T> void postcondition(Function<T, Boolean> predicate) {
...
+    public static Sequence<Integer> sequence(int[] array) {
+        return new ArraySequence<>(array);
+    }
+
+    public interface Sequence<T> {
+        /**
+         * Returns {@code true} if this sequence contains the specified element.
+         */
+        default boolean contains(T element) {
+            throw new UnsupportedOperationException("Verification-only method called at runtime");
```
keyboardDrummer: We could use a separate exception for this, then the message does not need to be repeated.

library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -17,6 +18,58 @@ public static <T> void postcondition(Function<T, Boolean> predicate) {
...
+    public static <T> boolean forall(Function<T, Boolean> predicate) {
+        throw new UnsupportedOperationException("Verification-only method called at runtime");
+    }
+
+    public static <T1, T2> boolean forall(BiFunction<T1, T2, Boolean> predicate) {
+        throw new UnsupportedOperationException("Verification-only method called at runtime");
+    }
+
+    public static <T> Sequence<T> sequence(T[] array) {
+        return new ArraySequence<>(array);
```
keyboardDrummer: Consider already throwing an exception here.

library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -17,6 +18,58 @@ public static <T> void postcondition(Function<T, Boolean> predicate) {
...
+         */
+        default boolean contains(T element) {
+            throw new UnsupportedOperationException("Verification-only method called at runtime");
+        }
+
+        /**
+         * Returns {@code true} if the portion of this sequence starting at {@code fromIndex}, inclusive,
+         * contains the specified element.
+         */
+        default boolean contains(T element, int fromIndex) {
```
keyboardDrummer: I'm not used to these two variants of contains in sequence APIs. Why not add take/drop and a single contains without indices?

alex-chew: Mainly, the standard Java APIs commonly include such variants for 0, 1, or 2 indices when relevant - for example, `String#indexOf`, `String#substring`, `List#subList`, etc. - and so it would be natural for users to reach for similar APIs on `Sequence`. By contrast, I'm not aware of index-based `take` or `drop` methods in the standard APIs.

keyboardDrummer: > String#substring, List#subList

How could you define those methods without `from` and `to` parameters?

> String#indexOf

This one indeed does not need the `fromIndex` parameter. Not that it does not have a variant that takes an upper limit.

> and so it would be natural for users to reach for similar APIs on Sequence

`String.contains` and `List.contains` do not take indices, so I still think it's unexpected for a `contains` method to do so. Also, I think from a design perspective it's unnecessary since you can do `.subsequence(from,to).contains(value)`

A separate point, I think it's a bit dangerous not to expose the underlying Dafny API directly, since that might affect verification performance. For example, your current contains implementation always generates Dafny code that creates a subsequence, even if no indices are specified.

alex-chew: I take your point that `subsequence` (with or without `take` and `drop`) would obviate the need for multiple `contains` variants. On the other hand, I don't love the verbosity of writing `sequence(arr).subsequence(lo, hi).contains(key)`, and I suspect users will not love it either. What do you think of this set of APIs?

* Static methods `sequence(arr)`, `sequence(arr, lo)`, and `sequence(arr, lo, hi)`, corresponding to the equivalent Dafny array-to-sequence conversions
* `Sequence` instance methods `drop(lo)`, `take(hi)`, and `subsequence(lo, hi)`, corresponding to the equivalent Dafny sequence expressions
* a `Sequence` instance method `contains(key)`, corresponding to Dafny's `in`

> A separate point, I think it's a bit dangerous not to expose the underlying Dafny API directly, since that might affect verification performance. For example, your current contains implementation always generates Dafny code that creates a subsequence, even if no indices are specified.

I made the simplification of always creating a subsequence even if no indices are specified, under the assumption that when `s` is a sequence, `s` and `s[..]` are effectively identical during verification. But perhaps this isn't true, and it would be safer to avoid the subsequence in this case.

> This one indeed does not need the fromIndex parameter. Not that it does not have a variant that takes an upper limit.

It does have a variant that takes an upper limit, since Java 21: https://docs.oracle.com/en/java/javase/21/docs//api/java.base/java/lang/String.html#indexOf(int,int,int)

keyboardDrummer:
I'm approving this PR, but since I created it I can't approve it on GH.

alex-chew (approved):

PR #29, opened 2025-03-20
Support UserProfile example

keyboardDrummer:
### What was changed?
- Use .java programs from the examples project for tests, when possible, to prevent duplication and guarantee the examples to remain valid.
- Add an option to report source location ranges instead of positions, which is useful for IDEs
- Improve the source locations passed to Dafny
- Add the `@Invariant` annotation. Note that it would be cleaner if this was implementation as as separate compilation step that inserts the pre and postconditions, to reduce the complexity of `JavaToDafnyCompiler`
- Support `enum` declarations (they are translated to Dafny datatypes)
- Support fields without initializers
- Support constructors, `new`, `this` and `super` calls without arguments.
- Support return statements without arguments (in addition to those with arguments)
- Support modifies clauses
- Support `!=`
- Support `@Nullable`
- Simplify debugging by making it easy to pass `--wait-for-debugger` to Dafny
- Prevent getting warnings from Dafny about suspicious  indentation.

### How has this been tested?
- Added the test `UserProfile.java` and updated existing tests with improved source locations


library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -2,8 +2,14 @@
 
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 public class JVerify {
+
+    public static boolean implies(boolean antecedant, boolean consequent) {
+        return !antecedant || consequent;
+    }
```
alex-chew: nit:
```suggestion
    public static boolean implies(boolean antecedent, boolean consequent) {
        return !antecedent || consequent;
    }
```

library/src/main/java/com/aws/jverify/ExternalContract.java
```diff
@@ -0,0 +1,12 @@
+package com.aws.jverify;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.SOURCE)
+@Target({ElementType.TYPE})
+public @interface ExternalContract {
```
alex-chew, line 14: Could you add some documentation to this class (and also to `ContractException`)? It looks like the relevant example will be in a different PR, so if you want to add the documentation there instead, that'd be okay too.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -415,16 +565,131 @@ BinaryExprOpcode toDafny(Symbol.OperatorSymbol operator) {
...
+    
+    private int getStartPos(JCTree tree) {
+        return switch(tree) {
+            case JCTree.JCClassDecl classDecl -> getClassNamePosition(classDecl);
+            case JCTree.JCMethodDecl methodDecl -> getMethodNamePosition(methodDecl);
+            default -> tree.getStartPosition();
+        };
+    }
+
+    private int getMethodNamePosition(JCTree.JCMethodDecl methodDecl) {
```
alex-chew, line 600: It's unfortunate that we have to implement these parsing bits ourselves; I'm surprised that the javac APIs don't seem to have this functionality.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -169,24 +266,39 @@ MemberDecl translateMember(JCTree member) {
...
+                return new Method(origin, name, null, false, null, List.of(),
+                        ins, header.preconditions, header.postconditions, header.getReads(),
+                        header.getDecreases(), header.getModifies(), isStatic, outs,
+                        new BlockStmt(toOrigin(method.body), null, bodyStatements), false);
+            }
+        }
+    }
+
+    private void applyInvariants(JCTree.JCMethodDecl method, HeaderContainer header) {
+        boolean isPublic = (method.getModifiers().flags & Flags.PUBLIC) != 0;
```
alex-chew, line 293: Just a question - do we have a sense whether enforcing `Invariant`s only on public methods, will be sufficient for the case study codebases we want to support?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -169,24 +266,39 @@ MemberDecl translateMember(JCTree member) {
...
-                            new Specification<FrameExpression>(List.of(), null),
-                            new BlockStmt(origin, null, bodyStatements), false);
+            var bodyStatements = translateStatements(postHeader);
+            if (method.name.contentEquals("<init>")) {
+                return new Constructor(origin, new Name(origin, "_ctor"), null, false, null, List.of(), ins,
+                        header.preconditions, header.postconditions, header.getReads(), 
+                        header.getDecreases(), header.getModifies(),
+                        new DividedBlockStmt(toOrigin(method.body), null, bodyStatements, null, List.of()));
+            } else if (annotationsByName.containsKey(Proof.class.getSimpleName())) {
+                return new Method(origin, name, null, false, null, List.of(),
```
alex-chew, line 278: We should consider sanitizing names that may conflict with Dafny keywords, but it's not a blocker for this PR.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: Yes I'm surprised too. It would have helped a lot of they would at least store a starting and end position for each `Name`, but they don't. I've looked at the parser code of the source and the information is just not stored anywhere. I'm guessing this indicates we'll run into more issues with IDE support down the line, like lack of support for incremental compilation. I've been told that the javac codebase is used to implement NetBeans, so that's good, but I have not verified this. Also have not looked into incremental compilation support.

The alternative codebases we could base JVerify on are Eclipse and IntelliJ.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: Users can manually add them to private methods. We could support invoking `Jverify.invariant` at the method level as well, not just for loops, to enable quickly adding both a pre and postcondition at the same time.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: Added a GH issue to the list, although at the bottom: https://github.com/orgs/aws/projects/276/views/1

PR #43, opened 2025-03-24
Allow checking preconditions at runtime

keyboardDrummer:
### What was changed?
- Enable using the javac compiler plugin to convert preconditions into runtime checks

### How has this been tested?
- Added a test `checkPreconditionCheckedAtRuntime`


verifier/src/main/java/com/aws/jverify/verifier/Main.java
```diff
@@ -15,6 +15,10 @@
 import picocli.CommandLine.Option;
 import picocli.CommandLine.Parameters;
 
+import javax.sql.CommonDataSource;
+
+import static com.aws.jverify.common.Common.*;
+
```
alex-chew: nit: are these intentional additions?

javac-plugin/src/test/java/com/aws/jverify/plugin/CompilationRunner.java
```diff
@@ -0,0 +1,123 @@
...
+    private static void deleteRecursively(Path path) throws IOException {
+        if (Files.isDirectory(path)) {
+            try (var files = Files.list(path)) {
+                for (Path file : files.toList()) {
+                    deleteRecursively(file);
+                }
+            }
+        }
+        Files.delete(path);
+    }
```
alex-chew: nit: There's a near-duplicate implementation in TestPlugin.

javac-plugin/src/main/java/com/aws/jverify/plugin/RemoveJVerifyVisitor.java
```diff
@@ -0,0 +1,106 @@
...
+
+            var check = maker.If(maker.Unary(JCTree.Tag.NOT, head), maker.Block(0, List.of(
+                    maker.Throw(maker.NewClass(null,null,
+                            maker.QualIdent(getClassSymbol(PreconditionFailure.class, context)), 
+                            List.nil(), null))
+            )), emptyBlock);
+            return statements.tail.prepend(check);
+        } else {
+            return statements.tail;
+        }
```
alex-chew: It looks like this only supports a single dynamic precondition per relevant method body - is that correct and intentional?

javac-plugin/src/main/java/com/aws/jverify/plugin/RemoveJVerifyVisitor.java
```diff
@@ -0,0 +1,106 @@
...
+
+        var annotations = currentMethod.getModifiers().getAnnotations();
+        var annotationsByName = annotations.stream().collect(Collectors.toMap(
+                (JCTree.JCAnnotation a) -> a.getAnnotationType().toString(),
+                a -> a));
+        this.dynamicPreconditions = annotationsByName.containsKey(CheckPreconditionsAtRuntime.class.getSimpleName());
+        
+        super.visitMethodDef(tree);
+        Attr attr = Attr.instance(context);
+        attr.attribStat(tree, currentEnv);
```
alex-chew, line 56: Just out of curiosity - will this be necessary whenever we add or modify AST nodes during a traversal?

keyboardDrummer: Yes, this invokes the "resolver" on the newly added nodes. I invoked it at the statement level earlier and that caused resolution to happen incompletely.

javac-plugin/src/main/java/com/aws/jverify/plugin/RemoveJVerifyVisitor.java
keyboardDrummer: You're right, thanks for spotting that! Added a test and fix

verifier/src/main/java/com/aws/jverify/verifier/Main.java
keyboardDrummer: They seem to be gone now. Not sure where they went :-D

javac-plugin/src/test/java/com/aws/jverify/plugin/CheckPreconditionAtRuntimeTest.java
```diff
@@ -0,0 +1,35 @@
...
+        noPreconditions(1);
+        noStatements(1);
+    }
+
+    @CheckPreconditionsAtRuntime
+    static int multiplePreconditions(int input) {
+        precondition(true);
+        precondition(input != 0);
+        return input;
+    }
```
alex-chew, line 38: It would be better for both preconditions to be fallible - something like `precondition(input > 0); precondition(input < 10);`, along with another call site demonstrating that the latter will indeed throw.

keyboardDrummer: Why would that be better? We can come up with an arbitrary number of tests, for example a test for each particular number of preconditions. This test identifies the bug you found. Is there a particular bug you're looking for with what you're suggesting?

javac-plugin/src/test/java/com/aws/jverify/plugin/CompilationRunner.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        try {
+            writeCompiledClassFiles(compilation, tempDir);
+
+            // Prepare command to run the class in a separate JVM
+            String javaHome = System.getProperty("java.home");
+            String javaBin = javaHome + File.separator + "bin" + File.separator + "java";
+
+            var process = getProcess(classPath, qualifiedClassName, args, javaBin, tempDir);
```
alex-chew: Consider querying the current process for the Java executable, which has the added benefit of working on Windows:

```suggestion
            // Prepare command to run the class in a separate JVM
            var javaBin = ProcessHandle.current().info().command();
            if (javaBin.isEmpty()) {
                System.err.println("Failed to find Java executable");
                return -1;
            }

            var process = getProcess(classPath, qualifiedClassName, args, javaBin.get(), tempDir);
```

javac-plugin/src/test/java/com/aws/jverify/plugin/CompilationRunner.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @NotNull
+    private static Process getProcess(Iterable<File> classPath, String qualifiedClassName, String[] args, String javaBin, Path tempDir) throws IOException {
+        List<String> command = new ArrayList<>();
+        command.add(javaBin);
+        command.add("-cp");
+        StringBuilder arg = new StringBuilder();
+        for(var cpFile : classPath) {
+            arg.append(cpFile.toString()).append(":");
```
alex-chew: Windows uses `;` instead of `:` to separate classpath entries - consider using:

```suggestion
            arg.append(cpFile.toString()).append(File.pathSeparator);
```

javac-plugin/src/test/java/com/aws/jverify/plugin/CheckPreconditionAtRuntimeTest.java
alex-chew: My bad - I thought that by changing the test this way, we were no longer testing that the first precondition in a `CheckPreconditionsAtRuntime` would be dynamically checked, but in fact the RuntimePreconditionExample still does so.

alex-chew (commented):
I left a couple suggestions so the tests pass on Windows. Everything else looks good to me.

javac-plugin/src/test/java/com/aws/jverify/plugin/CheckPreconditionAtRuntimeTest.java
keyboardDrummer: Added a single precondition case to this file.

keyboardDrummer:
> I left a couple suggestions so the tests pass on Windows. Everything else looks good to me.

Thanks. Created an issue for adding Windows tests: https://github.com/aws/jverify/issues/51

alex-chew (approved):

PR #76, opened 2025-03-31
Tweaks

### LINKED ISSUE #60: Make it easy to demo the runtime checking feature

keyboardDrummer:
Fixes https://github.com/aws/jverify/issues/60

### What was changed?
- Compile the library project using Java 17, so it can be used by projects using that Java version
- Make it easier to run `javac` with the JVerify plugin, using a gradle task `compileWithJVerify`
- Add new annotations, `InheritContract` and `Immutable`, even though they're not interpreted yet.

### How has this been tested?
- Manually tested the task `compileWithJVerify`


build.gradle.kts
```diff
@@ -115,11 +143,11 @@ project(":javac-plugin") {
...
-    tasks.withType<JavaExec> {
-        jvmArgs = createJavacExports(listOf("ALL-UNNAMED", "com.aws.jverify.plugin")). 
-          // Using preview mode, so we can use the package java.lang.classfile
-          plus("--enable-preview")
-    }
+//    tasks.withType<JavaExec> {
+//        jvmArgs = createJavacExports(listOf("ALL-UNNAMED", "com.aws.jverify.plugin")). 
+//          // Using preview mode, so we can use the package java.lang.classfile
+//          plus("--enable-preview")
+//    }
```
alex-chew: nit: Remove if unused

build.gradle.kts
```diff
@@ -26,9 +26,13 @@ java {
 }
 
 project(":library") {
+    java {
+        toolchain {
+            languageVersion = JavaLanguageVersion.of(17) // Use Java 17 for this subproject
```
alex-chew: nit:

```suggestion
            // Use Java 17 for this subproject, so Java 17 projects can depend on it at compile-/run-time
            languageVersion = JavaLanguageVersion.of(17)
```

README.md
```diff
@@ -8,6 +9,12 @@
 - Java extensions for VSCode should work well on these examples
 - Run the VSCode task "verify" on examples to get verification diagnostics
 
+### Compile Java code using JVerify
+- Checkout this repository and its submodules
+- Run `dotnet build dafny/Source/Dafny.sln`
+- Run `gradle compileWithJVerify -PjavacArgs="<pathToYourJavaFile>,-d=<yourDesiredOutputDirectory>`
+- Run the resulting class files using `java`
```
alex-chew: I tried these instructions and ran into a few issues:

- Since `compileWithJVerify` is defined as a task on `javac-plugin`, the paths of both the input Java file and the output directory are interpreted relative to the `javac-plugin` directory, which is a bit confusing.
- The `java` command will require specifying a classpath that includes both the output directory (not relative to `javac-plugin` this time) and the library JAR.

It would be good write out concrete example commands, e.g. for the runtime precondition example.

library/src/main/java/com/aws/jverify/InheritContract.java
alex-chew, line 1: I would strongly prefer to move these additions to the PR(s) that will use and test them.

alex-chew (changes requested):

library/src/main/java/com/aws/jverify/InheritContract.java
keyboardDrummer: I'm currently using them in a customer prototype. They're not functional but I use them to indicate what things will look like. I've updated the PR so JavaToDafnyCompiler will report a not supported error if any if those are used. Currently we're not testing all the not supported messages for not supported features, so I did not add tests for that here either. I think it would be better if we had those tests as well.

README.md
keyboardDrummer: Good callouts, thanks ❤️  updated the text.

README.md
```diff
@@ -8,6 +9,12 @@
 - Java extensions for VSCode should work well on these examples
 - Run the VSCode task "verify" on examples to get verification diagnostics
 
+### Compile Java code using JVerify
+- Checkout this repository and its submodules
+- Run `dotnet build dafny/Source/Dafny.sln`
```
olivier-aws: Why do we need to build Dafny only to compile Java code?

keyboardDrummer: I don't think we do, updated the text, thanks.

alex-chew (approved):

PR #93, opened 2025-04-04
Support external contracts

### LINKED ISSUE #31: Support int.MaxValue and update tests/examples

### LINKED ISSUE #39: Enable adding contracts to existing code
Contracts are added by creating a new type references the existing type

```java
class ExistingType { 
  void foo(int x) { }
  void bar(int x) { }
}

@ExternalContract(ExistingType.class)
class ExistingTypeContract1 extends ExistingType {
  @Override
  void foo(int x) {
    precondition(x > 0);
  }
}

@ExternalContract(ExistingType.class)
class ExistingTypeContract2 extends ExistingType {
  @Override
  void bar(int x) {
    precondition(x > 0);
  }
}
```

keyboardDrummer:
Fixes #39
Fixes #31 

### What was changed?
- Add support for external contracts
- Add support for @Verify, to configure whether symbols and their children should be verified or not.
- Add built-in contract for Integer.MaxValue

### How has this been tested?
- Update `FibonacciValid.java` example test
- Add `ExternalContract` example test
- Add `ContractErrors.java` test
- Add `ShouldVerify.java` test to test `@Verify`. It does not test it on packages yet since we only have single file support.


verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -750,9 +961,12 @@ private int getClassNamePosition(JCTree.JCClassDecl classDecl) {
         }
 
         int pos = classDecl.pos;
+
         if (classDecl.mods != null && classDecl.mods.pos != Position.NOPOS) {
-            pos = getEndPos(classDecl.mods);
+            // Not sure if this statement is useful
+            pos = Math.max(pos, getEndPos(classDecl.mods));
         }
+        compilationUnit.getLineMap().getColumnNumber(325);
```
alex-chew: nit: leftover from debugging?
```suggestion
```

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -605,9 +802,22 @@ private SeqSelectExpr toSubsequence(JCTree.JCExpression seqOrArray, JCTree.@Null
...
+            
+            // Remove the name qualification because we do not support that yet
+            Expression nameSegment;
+            if (expression instanceof ExprDotName exprDotName) {
+                nameSegment = new NameSegment(exprDotName.getOrigin(), exprDotName.getSuffixNameNode().getValue(), null);
+            } else if (expression instanceof NameSegment nameSegment1) {
+                nameSegment = nameSegment1;
+            } else {
+                nameSegment = expression;
+            }
```
alex-chew: Are these branches supposed to behave the same?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -214,15 +373,33 @@ MemberDecl translateMember(JCTree member, Stack<Tree> nestedTypes) {
         }
     }
 
-    private Field translateField(JCTree.JCVariableDecl variableDecl) {
+    private @Nullable Field translateField(JCTree.JCVariableDecl variableDecl) {
```
alex-chew, line 384: nit: If we're not yet supporting final fields without initializers, it would be nice to report an error.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -121,19 +161,126 @@ private void reportError(JCDiagnostic.DiagnosticPosition position, String key, O
...
+    private boolean shouldVerify() {
+        for (int i = shouldVerifies.size() - 1; i >= 0; i--) {
+            var mode = shouldVerifies.get(i);
+            if (mode == ShouldVerifyMode.AlwaysYes || mode == ShouldVerifyMode.DefaultYes) {
+                return true;
+            } else if (mode == ShouldVerifyMode.AlwaysNo || mode == ShouldVerifyMode.DefaultNo) {
+                return false;
+            }
+        }
+        return false;
```
alex-chew: nit: If `shouldVerifies` has the invariant that it should never be empty (i.e. it should always have the initial `DefaultYes` at the bottom of the stack), consider throwing an exception here to help detect bugs.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -121,19 +161,126 @@ private void reportError(JCDiagnostic.DiagnosticPosition position, String key, O
 
     List<AttributedExpression> invariants = new ArrayList<>();
     
+    enum ShouldVerifyMode { AlwaysYes, DefaultYes, AlwaysNo, DefaultNo, Inherit }
+    private final Stack<ShouldVerifyMode> shouldVerifies = new Stack<>();
```
alex-chew, line 165: I'm surprised that I don't see any calls to `shouldVerifies.pop()` - am I missing something?

library/src/main/java/com/aws/jverify/Verify.java
```diff
@@ -10,7 +10,8 @@
  */
 @Retention(RetentionPolicy.SOURCE)
 @Target({ElementType.METHOD, ElementType.TYPE, ElementType.PACKAGE})
-public @interface ShouldVerify {
-    boolean shouldVerify() default true;
+public @interface Verify {
+    boolean should() default true;
+    boolean pushDown() default false;
```
alex-chew: A few comments on this:

- I feel that it would be more user-friendly for `@Verify` to be propagated by default, rather than by opt-in. Intuitively, I'd expect members of a `@Verify(true)`-annotated type to themselves be verified, and likewise for `@Verify`-annotated packages and their member types. (And I don't think the choice matters for methods since they don't have members in the first place.)
- I think naming the first element `value` would make more sense both intuitively (`should` is an awkward name IMO) and because it would allow writing just `@Verify(true)` or `@Verify(false)`.
- I think naming the second element `members` would be clearer, assuming that the intent is to propagate from type-to-methods, type-to-fields, and package-to-types. If instead the intent is to propagate from type-to-subtype, perhaps `inherited` (or `inheritable` or `heritable`) would be clearer. If both meanings are intended, or if we're not sure and want to retain flexibility, maybe even `propagate` would make the most sense.

Whichever decisions we make, we should add documentation to the annotation elements accordingly.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -121,19 +161,126 @@ private void reportError(JCDiagnostic.DiagnosticPosition position, String key, O
...
+                var arguments = getArguments(verifyAnnotation);
+                var shouldArgument = arguments.get("should");
+                var should = true;
+                if (shouldArgument != null) {
+                    if (shouldArgument instanceof JCTree.JCLiteral literal) {
+                        should = literal.value.equals(true);
+                    } else {
+                        throw new JavaViolationException();
+                    }
+                }
```
alex-chew: nit: Extracting a helper method to retrieve literal values from annotation arguments would improve legibility in many places in this class.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: Yes but I'll remove the redundant one.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: Is the code not doing that?

```java
        if (variableDecl.getInitializer() != null) {
            var isFinal = (variableDecl.mods.flags & Flags.FINAL) != 0;
            if (isFinal) {
                var rhs = toExpr(variableDecl.getInitializer());
                var isStatic = (variableDecl.mods.flags & Flags.STATIC) != 0;
                return new ConstantField(origin, fieldName, getAttributes(origin), false, type, rhs, isStatic, false);
            } else {
                reportError(variableDecl, "notSupported", "Field initializers");
                return null;
            }
        }
        ```

File: verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java (hunk 5 above)
Comment (keyboardDrummer) [reply verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java:165]: No you're not missing anything, and similarly no tests were added for this feature either. I didn't intend to add this feature in this PR, but felt it was necessary for the contracts feature, and then somehow added it half-baked. I'll add tests and fix bugs.

File: library/src/main/java/com/aws/jverify/Verify.java (hunk 6 above)
Comment (keyboardDrummer) [reply library/src/main/java/com/aws/jverify/Verify.java]: > I feel that it would be more user-friendly for @Verify to be propagated by default, rather than by opt-in.

Agree, but it already always propagates down. The second argument determines whether it overrides any @Verify annotations placed on children. This is useful if you want to make sure an entire package gets verified, or - during development - want to turn off verification of an entire package.

> because it would allow writing just @Verify(true) or @Verify(false).

That is nice. Would you then prefer the name `@Verify` or `@ShouldVerify` ?

> I think naming the second element members would be clearer

I think you misunderstood the meaning of the second argument. Renamed it to `overrideChildren` now.

> Whichever decisions we make, we should add documentation to the annotation elements accordingly.

Definitely, thanks

File: verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java (hunk 7 above)
Comment (keyboardDrummer) [reply verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java]: Changed it although I'm not sure it's what you had in mind

Comment (alex-chew) [reply verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java]: Sorry, I should've been more specific - I was thinking of a helper method that would do most of the parsing work (get the argument tree by name, check that it's a literal value, cast to the right type, use a default value if absent), so that you could write something like:

```
if (verifyAnnotation != null) {
    var arguments = getArguments(verifyAnnotation);
    var value = getAnnotationArgument<Boolean>("value", true);
    var overrideChildren = getAnnotationArgument<Boolean>("overrideChildren", true);
    addShouldVerify(getVerifyMode(value, overrideChildren));
}
```

But I think the change you made is an improvement and this isn't a blocker.

library/src/main/java/com/aws/jverify/Verify.java
alex-chew: Ah, thanks for the clarification! Having "override" in the name makes the intended use case a lot clearer and it does sound quite useful.

> Would you then prefer the name `@Verify` or `@ShouldVerify`?

I don't have a strong opinion either way, but since `@Verify` is shorter and is seems unlikely to be confusing (at least, I can't recall Dafny users being confused by the meaning of `{:verify ...}`), I would slightly prefer that.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
alex-chew: I don't think so, unless I'm misreading it. If the field in question is final and has no initializer, then this `if` block won't be entered, and the following `return new Field(...)` will be incorrect (i.e. it should instead be a `new ConstantField`), right?

alex-chew (approved):
LGTM. (There are some non-blocking threads still open but we can address them as follow-up)

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: > and the following return new Field(...) will be incorrect

Why would it be incorrect? Looks good to me

PR #94, opened 2025-04-07
Support all primitive types

### LINKED ISSUE #80: Support primitives types, float/double can be opaque
fully supported: byte, short, long, char
supported but opaquely, so nothing can be proven about them: float, double

for char: `type char16 = c: char | '\U{0000}' <= c <= '\U{ffff}'`

keyboardDrummer:
Fixes #80 

### What was changed?
- Support all primitive types

### How has this been tested?
- Add test Types.java. Tests for combinations with operators will be added in follow-up PR when I add all the operators


tmp.dfy
alex-chew, line 1: I think this file is leftover from debugging?

verifier/src/main/java/com/aws/jverify/verifier/Encoder.java
```diff
@@ -4,11 +4,14 @@ public interface Encoder {
...
 
     void writeInt(int value);
+    void writeLong(long value);
 
     void writeQualifiedName(String name);
 
     void writeString(String value);
     void writeNullable(boolean isNull);
+
+    void writeDouble(Double d);
```
alex-chew: nit: Is there a particular reason to use the wrapped type here instead of `double`?

verifier/src/main/resources/additional.dfy
```diff
@@ -1,5 +1,15 @@
...
+type int16 = x: int | -0x8000 <= x <= 0x7fff
 
-type int32 = x: int
-  | -0x7fff_ffff <= x && x <= 0x7fff_ffff // lower bound is incorrect
\ No newline at end of file
+type nat31 = x: int32 | x >= 0
+type int32 = x: int | -0x8000_0000 <= x && x <= 0x7fff_ffff
+  
+type nat63 = x: int64 | x >= 0
+type int64 = x: int | -0x8000_0000_0000_0000 <= x <= 0x7fff_ffff_ffff_ffff
```
alex-chew: nit: I prefer the `nat32`/`nat64`/etc names - they seem a bit easier to work with (no need for the `bits - 1` logic in the type translation) and they're familiar since we also use them in the Dafny standard libraries. But it's not a big deal since these types aren't user-facing (if I understand correctly).

verifier/src/main/resources/builtin-contracts.java
```diff
@@ -0,0 +1,28 @@
...
+
+@Contract(Long.class)
+class LongContract {
+    public static final long MIN_VALUE = 0x8000000000000000L;
+    public static final long MAX_VALUE = 0x7fffffffffffffffL;
+}
+/**
+ * A constant holding the minimum value a {@code long} can
+ * have, -2<sup>63</sup>.
+ */
```
alex-chew: It looks like the doc comments got moved incorrectly?

verifier/src/test/java/com/aws/jverify/ContractErrors.java
```diff
@@ -0,0 +1,28 @@
...
+
+@Contract(Foo.class)
+class FooContract1 {    
+    void bar(int x) {
+        check(x > 0);
+    }
+}
+
+  @Contract(Foo.class)
+//^ error: a class may only be referenced by a @Contract annotation once
```
alex-chew: It looks like this test might not be checked properly - the error markup here says

> `a class may only be referenced by a @Contract annotation once`

but the `messages.properties` file says

> `a class may be referenced by a @Contract annotation at most once`

(On a related note, I'd suggest changing the message to `a class may not be referenced by more than one @Contract annotation` so that it's more consistent with the other messages.)

alex-chew (changes requested):

verifier/src/main/resources/additional.dfy
keyboardDrummer: The nat31 makes it clear to me that the lack of signage does not increase the bound by repurposing the bit. What do you think in regards to that? The AWS IAM Auth library does it like that as well.

verifier/src/test/java/com/aws/jverify/ContractErrors.java
keyboardDrummer: > It looks like this test might not be checked properly - the error markup here says

Updated the framework so we won't miss expected errors any more

> (On a related note, I'd suggest changing the message to

OK!

verifier/src/main/resources/additional.dfy
alex-chew: Yeah, I do understand that angle too and I don't have a strong opinion either way.

PR #96, opened 2025-04-08
Support all operators

### LINKED ISSUE #56: Support all operators (unary/binary/assignment)

keyboardDrummer:
Fixes #56

### What was changed?
- Support assignment operators when used as a statement
- Support increment/decrement operators when used as a statement (like in a for loop)
- Add testing for all operators

### Caveats
Created tickets for not yet support operators:
- https://github.com/aws/jverify/issues/99
- https://github.com/aws/jverify/issues/100

### How has this been tested?
- Add tests for boolean, integer and numeric operators.


verifier/src/test/java/com/aws/jverify/ResolutionErrorsDoubleOperators.java
```diff
@@ -0,0 +1,73 @@
...
+//      ^ error: operator --(double) is not supported
+        //check(l == 3f);
+        
+        var multiplication = l * r;
+//                             ^ error: operator *(double,double) is not supported
+        //check(multiplication == 9f);
+        var division = l / r;
+//                       ^ error: operator /(double,double) is not supported
+        //check(division == 1f);
+        var remainder = l % 2f;
```
alex-chew: nit just for consistency:
```suggestion
        var remainder = l % 2d;
```

verifier/src/test/java/com/aws/jverify/ResolutionErrorsDoubleOperators.java
```diff
@@ -0,0 +1,73 @@
...
+        l -= r;
+//        ^ error: operator -(double,double) is not supported
+        //check(l == 3f);
+        l *= r;
+//        ^ error: operator *(double,double) is not supported
+        //check(l == 9f);
+        l /= r;
+//        ^ error: operator /(double,double) is not supported
+        //check(l == 3f);
+        l %= 2L;
```
alex-chew: nit: why `2L`?

verifier/src/test/java/com/aws/jverify/VerifyBooleanOperators.java
```diff
@@ -0,0 +1,20 @@
+package com.aws.jverify;
+
+import static com.aws.jverify.JVerify.check;
+
+@SuppressWarnings({"ConstantValue", "PointlessBooleanExpression"})
+class VerifyBooleanOperators {
+    public void foo() {
+        check(false);
+//      ^^^^^^^^^^^^ Error: assertion might not hold
```
alex-chew: Wouldn't this mask any verification failures below if there are errors in the implementation? e.g. if `var or = p || q` were incorrectly translated to `var or := p && q`, then the test would still pass?

(And same for `VerifyNumericOperators`)

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -22,6 +22,7 @@
 
 import javax.lang.model.type.ArrayType;
 import javax.lang.model.type.TypeKind;
+import javax.swing.text.html.HTML;
```
alex-chew: nit: accidental addition?

verifier/src/test/java/com/aws/jverify/ResolutionErrorsNumericOperators.java
```diff
@@ -0,0 +1,46 @@
...
+ * Numeric operators are those that apply to the types:
+ * byte, short, int, long, float, double, char
+ */
+@SuppressWarnings("ConstantValue")
+class ResolutionErrorsNumericOperators {
+    public void foo() {
+        var l = 3;
+        var r = 3;
+        var incrementPostfix = l++;
+//                              ^ error: operator ++(int) is not supported
```
alex-chew: This is a bit confusing since the increment/decrement operators are supported in their statement form - would it be feasible to use the "since ... performs mutation, ..." message in this case? Though if we expect to add support for it soon, then it's not a big deal.

verifier/src/test/java/com/aws/jverify/ResolutionErrorsNumericOperators.java
```diff
@@ -0,0 +1,46 @@
...
+        check(incrementPrefix == 4);
+        check(l == 4f);
+        var decrementPrefix = --l;
+//                            ^ error: operator --(int) is not supported
+        check(decrementPrefix == 3);
+        check(l == 3);
+        
+        var negation = -l;
+//                     ^ error: operator -(int) is not supported
+        check(negation == -3);
```
alex-chew: Why not support unary `-`? (And unary `+`, for that matter.)

verifier/src/test/java/com/aws/jverify/ResolutionErrorsDoubleOperators.java
keyboardDrummer: `2` was to get an interesting answer out of the `%`. `L` was because of copy paste. Removed the copied Ls

verifier/src/test/java/com/aws/jverify/VerifyBooleanOperators.java
keyboardDrummer: Good callout. I've moved this to the bottom.

verifier/src/test/java/com/aws/jverify/ResolutionErrorsNumericOperators.java
keyboardDrummer: Do you want to support `-` by translating to `* -1` ? OK I'll do that

verifier/src/test/java/com/aws/jverify/ResolutionErrorsNumericOperators.java
keyboardDrummer: > Though if we expect to add support for it soon, then it's not a big deal.

I'm not sure. I think it might be better not to support it, because it can be confusing to use it in an expression context. I'm thinking of waiting for customer feedback.

verifier/src/test/java/com/aws/jverify/ResolutionErrorsNumericOperators.java
alex-chew: Oops, sorry for not re-reviewing yesterday - This seems fine although we could also translate to Dafny's own unary minus operator, and I'm not sure whether it matters which translation we pick (in particular for verification performance).

keyboardDrummer: Ah, will fix in next PR

PR #108, opened 2025-04-11
Statements

### LINKED ISSUE #37: Support easily translatable Java statements
- continue/break
- do while
- for loop
- native assert statement
- empty statement (`;`)

Out of scope:
- enhanced for loop `for(x : y) { }`
- try/catch/throw/synchronized/nested class

keyboardDrummer:
Fixes #37

### What was changed?
- Moved statement related translation code into `StatementCompiler.java`
- Added support for:
  - continue/break
  - labelled continue and break
  - for loop
  - do while
  - empty statement (;)

### How has this been tested?
- Added the statements.java test file


library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -16,6 +16,15 @@ public static void check(boolean condition) {
     public static void precondition(boolean condition) {
     }
 
+    /**
+     * Tell JVerify which value decreases on each
+     * - iteration of the loop, when used in a loop
+     * - recursive call, when used on a recursive method
+     * Knowing which value decreases is required for termination proofs
+     */
+    public static void decreases(Object value) {
```
robin-aws, line 25: How about a varargs parameter here (`Object... value`) so that we cleanly support multiple values in lexicographical tuples (i.e. `decreases(Repr, 0)`)?

There's always the annoying ambiguity that happens if you pass an array value to a varargs parameter, but I don't think that will bite us here because I've never seen a reason to use an array as a decreases clause value.

I realize we've kept things simpler for other clauses by only supporting multiple calls rather than multiple arguments to a single call (e.g. `reads(this); reads(Repr);`, but given the ordering is much more critical for decreases clause I think it's much better to bind them more tightly. Dafny technically supports multiple `decreases` clauses and maintains their order by concatenating them, but I've literally never seen anyone use that and find it surprising.

verifier/src/test/java/com/aws/jverify/TestVerifier.java
```diff
@@ -159,11 +163,14 @@ public void testRunThroughGradle() throws IOException, InterruptedException {
...
-        var exitCode = process.waitFor();
-        reader.close();
+        int exitCode;
+        try (var reader = new BufferedReader(new InputStreamReader(process.getInputStream()))) {
+            reader.transferTo(writer);
+            exitCode = process.waitFor();
+            reader.close();
+        }
         var output = canonicalizeNewlines(writer.toString());
+        // amazonq-ignore-next-line
```
robin-aws: OoC what was it complaining about?

verifier/src/test/java/com/aws/jverify/VerifyStatements.java
```diff
@@ -0,0 +1,137 @@
+package com.aws.jverify;
+
+import static com.aws.jverify.JVerify.*;
+
+@SuppressWarnings({"ConditionalBreakInInfiniteLoop", "StatementWithEmptyBody", "ConstantValue"})
+class VerifyStatements {
+    void whileWithBreak() {
+        var x = 0;
+        while(true) {
```
robin-aws, line 9: Nit, but I'm used to seeing a space after `while`, `for`, etc.

Will be more important to apply something like `checkstyle` if we promote any of these to actual examples in documentation.

verifier/src/test/java/com/aws/jverify/VerifyStatements.java
```diff
@@ -0,0 +1,137 @@
...
+        check(i == 5);
+    }
+
+    void forLoopBreak() {
+        int i = 0;
+        int x = 0;
+        for(i = 0; true; i = i + 1) {
+            decreases(5 - i);
+            invariant(i <= 5);
+            invariant(x == i);
```
robin-aws, line 42: (For later) I also wonder if it would be cheap to provide a linting rule to separate spec calls from implementation by a empty line (I'm already finding that helpful for reading this code)

verifier/src/test/java/com/aws/jverify/TestVerifier.java
```diff
@@ -242,7 +249,9 @@ private static VerifierOptions getVerifierOptions() {
         var libraryJar = Path.of("../library/build/libs/library.jar");
         var prelude = Path.of("./src/main/resources/additional.dfy");
         return new VerifierOptions(dafnyPath, libraryJar, prelude, 
-                null, null, true,
+                //Path.of("../temp.dfy"),
```
robin-aws, line 250: Intentional?

verifier/src/main/java/com/aws/jverify/verifier/Serializer.java
```diff
@@ -45,6 +47,9 @@ private void serializeValue(Object value, AnnotatedType annotatedType) {
         } else if (type instanceof TypeVariable<?> && value instanceof FrameExpression) {
             // cheat a little bit for `MethodOrFunction.reads`
             expectedClass = FrameExpression.class;
+        } else if (type instanceof TypeVariable<?> && value instanceof Expression) {
+            // cheat a little bit for `MethodOrFunction.decreases`
```
robin-aws, line 50: How is it cheating exactly?

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -4,6 +4,7 @@
 import picocli.CommandLine;
 
 import javax.tools.Diagnostic;
+import javax.tools.JavaFileManager;
```
robin-aws: Unused?

verifier/src/main/java/module-info.java
```diff
@@ -8,6 +8,7 @@
     requires info.picocli;
     requires com.aws.jverify;
     requires java.sql;
+    requires jdk.xml.dom;
```
robin-aws: Unused? (Might have missed it)

verifier/src/main/java/com/aws/jverify/verifier/StatementCompiler.java
```diff
@@ -0,0 +1,462 @@
...
+                ? block.getStatements()
+                : List.of(statement);
+        return translateHeader(statements, header);
+    }
+
+    /**
+     * Translates header statements from the start of {@code statements}
+     * until the first non-header statement or the end of the list,
+     * appending the translations to the given {@link HeaderContainer},
+     * and returning a list view of the remaining statements.
```
robin-aws: (I realize this is just getting moved in this PR, but still need to ask) Is there a separate pass that raises errors for header statements later on in the bodies?

verifier/src/main/java/com/aws/jverify/verifier/StatementCompiler.java
```diff
@@ -0,0 +1,462 @@
...
+    public StatementCompiler(JavaToDafnyCompiler compiler) {
+        this.compiler = compiler;
+        this.maker = TreeMaker.instance(compiler.context);
+    }
+
+    private final Queue<Label> labels = new LinkedList<>();
+    private final Map<String, JCTree.JCStatement> labelToLoop = new HashMap<>();
+    private final Map<JCTree.JCStatement, String> forLoopContinueLabels = new HashMap<>();
+    private final Set<JCTree.JCStatement> loopsWithContinue = new HashSet<>();
+    private JCTree.JCStatement outerLoop;
```
robin-aws: Given the translation of break and continue statements is non-trivial, it's probably worth a short comment paragraph about the strategy (or in a doc somewhere in the repo)

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -31,21 +31,27 @@
 
 public class JavaToDafnyCompiler {
     public static final String JVERIFY_CLASS = JVerify.class.getName();
+    public final Context context;
     JCTree.JCCompilationUnit compilationUnit;
     Stack<IOrigin> contextOrigins = new Stack<>();
     DiagnosticCollector<JavaFileObject> diagnostics = new DiagnosticCollector<>();
     
+    StatementCompiler statementCompiler;
```
robin-aws: Might be more development friendly to create a fresh one on demand for each method, since this has state that affects the synthetic identifiers we generate. That way the translation of a single method is deterministic, which tends to be helpful for debugging or any tests we have to write to assert the exact translated form of things.

library/src/main/java/com/aws/jverify/JVerify.java
keyboardDrummer: Is it OK if I leave finishing the decreases API out of scope for this PR? I had to add this here because it's needed for some of the loop examples that this PR is adding. I've added `decreases` clause to this issue: https://github.com/aws/jverify/issues/18

verifier/src/test/java/com/aws/jverify/TestVerifier.java
keyboardDrummer: O, did not realize it inserted this comment here. Removed it. I don't know what it was on about.

verifier/src/test/java/com/aws/jverify/TestVerifier.java
keyboardDrummer: Yes

verifier/src/main/java/com/aws/jverify/verifier/Serializer.java
keyboardDrummer: It's code that only works for specific cases instead of for the general case. Supporting the general case will be ~100 lines of code that requires knowing the javac API well.

verifier/src/main/java/com/aws/jverify/verifier/StatementCompiler.java
keyboardDrummer: Jep, there's a `contractAfterBody` error

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: > since this has state that affects the synthetic identifiers we generate

Right now that state is in `JavaToDafnyCompiler`, but I'll move it and apply your suggestion.

verifier/src/main/java/com/aws/jverify/verifier/StatementCompiler.java
keyboardDrummer: I've refactored the continue handling code a bit and added a comment there. Besides that, I'm not sure what to explain.

verifier/src/test/java/com/aws/jverify/VerifyStatements.java
keyboardDrummer: Would be nice yes. For now, since we're not checking anything related to styling in CI, I think it will be easier to review without regard for spaces.

library/src/main/java/com/aws/jverify/JVerify.java
robin-aws: Totally!

robin-aws (approved):

PR #110, opened 2025-04-17
feat: translate switch expressions

alex-chew:
### What was changed?

- Translate switch expressions
- Re-generate `QuantifiedVar` from updated Dafny submodule
- Fix detection of out-of-date generated AST (see [silent failure in previous `main` commit's CI](https://github.com/aws/jverify/actions/runs/14478842823/job/40611033819#step:8:29) which allowed `QuantifiedVar` to be out of date)

### How has this been tested?

Added positive and negative tests for both translation and verification.


verifier/src/test/java/com/aws/jverify/TestVerifier.java
```diff
@@ -202,6 +207,13 @@ private void testMarkedSourceVerification(String markedSource, DafnyResults dafn
         assertThat(output, endsWith(ending));
     }
 
+    /**
+     * Verifies the given source code, asserting that each markup annotation appears in the output,
+     * and that the exit code and number of matched error annotations are as expected.
+     * <p>
+     * NOTE: Errors that appear in the verification output, but which don't appear in the marked-up source code,
+     * are currently silently ignored.
```
robin-aws, line 215: Thanks for adding the comment. Is this just because there's no check that the `DafnyResults.errorCount` matches the number of markup annotations in the source? That seems super quick to fix.

I also wonder if we could avoid having to declare the expected number of successes and errors separately. The number of errors should match the number of annotations, and the number of successes is generally just annoying to maintain. IMO just checking that it's greater than 0 (to catch pipeline issues where we don't trigger verification at all) would be enough.

verifier/src/test/java/com/aws/jverify/TestVerifier.java
```diff
@@ -132,6 +132,11 @@ public void operators() throws IOException {
         verifyMarkedSourceFile("Operators.java", new DafnyResults(9, 9));
     }
 
+    @Test
+    public void switches() throws IOException {
```
robin-aws, line 136: Another side note about the testing framework. :) If we can get rid of the extra DafnyResults argument (or in the worst case add it as markup to the source), we can make this a parameterized test that reads all files in a directory. That would make sure we don't accidentally fail to run a test.

Makefile
```diff
@@ -2,4 +2,4 @@ generate-dafny-ast:
 	./gradlew :javaTypesGenerator:run --args="../dafny/Source/Scripts/Syntax.cs-schema ../verifier/src/main/java/com/aws/jverify/generated"
   
 test-generation-nodiff: generate-dafny-ast
-	(git status --porcelain || (echo 'Consider running `make generate-dafny-ast`'; exit 1; ))
\ No newline at end of file
+	test -z "$(git status --porcelain)" || { echo 'Consider running `make generate-dafny-ast`'; exit 1; }
```
robin-aws: Would you mind commenting on how the old version failed to work as expected? @keyboardDrummer and I put our heads together quite a bit to try to get it right and I'm annoyed we still didn't. :)

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -812,6 +813,74 @@ private SeqSelectExpr toSubsequence(JCTree.JCExpression seqOrArray, JCTree.@Null
...
     }
 
+    private Expression toExpr(JCTree.JCSwitchExpression switchExpr) {
+        var origin = toOrigin(switchExpr);
+
+        // A switch block consists of either *switch rules* (label -> body)
+        // or *switch labeled statement groups* (label: {label:} stmts).
+        // Unlike switch rules, switch labeled statement groups automatically "fall through" without break statements,
+        // but Dafny's match statement/expression can't express that easily.
+        // So for now we only support switch blocks using switch rules.
```
robin-aws, line 824: I'm very happy with this limitation and would seriously consider telling JVerify customers to stick to switch rules instead. :)

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -812,6 +813,74 @@ private SeqSelectExpr toSubsequence(JCTree.JCExpression seqOrArray, JCTree.@Null
...
+        if (switchExpr.getCases().getFirst().getCaseKind().equals(JCTree.JCCase.STATEMENT)) {
+            reportError(switchExpr, "notSupported", "switch labeled statement group");
+            return getHole(origin);
+        }
+
+        // Each case has a *switch label*, which is either a *default label* or a *case label*.
+        // A *case label* consists of either:
+        //  - a list of *case constants*
+        //  - a null literal
```
robin-aws: What source code leads to this case? I don't think I see it in the test file.

robin-aws (commented):
Nice work!

verifier/src/test/java/com/aws/jverify/TestVerifier.java
alex-chew: > Thanks for adding the comment. Is this just because there's no check that the `DafnyResults.errorCount` matches the number of markup annotations in the source? That seems super quick to fix.

No, we do check that the `expectedErrorCount` argument matches the number of markup annotations starting with "error"/"Error" (there are other kinds like "Related location"). It's more that we're not handling the output of the `javac` and `dafny` processes in a structured way (we dump everything in a string, just as we would write to stdout), so the number of compilation/verification output errors isn't trivial to obtain. I'll spend some time refactoring this logic so that at least we maintain the structured compilation diagnostics, but for Dafny I'm not sure whether we can do better than searching for output lines that look like errors.

> I also wonder if we could avoid having to declare the expected number of successes and errors separately. The number of errors should match the number of annotations, and the number of successes is generally just annoying to maintain. IMO just checking that it's greater than 0 (to catch pipeline issues where we don't trigger verification at all) would be enough.

Since there are "Related location" annotations (and potentially others), it's not quite that the number of errors should match the number of annotations. I also think it's worthwhile to maintain explicit counts of expected errors/successes, since doing so can help catch mistakes like [malformed error annotations](https://github.com/aws/jverify/pull/94#discussion_r2032108274) and accidentally not verifying symbols (which could easily happen with as we rapidly change the translation, and with `@Verify`).

verifier/src/test/java/com/aws/jverify/TestVerifier.java
alex-chew: That sounds like a good idea, especially with adding the expected success/error count as markup. I'll address the test framework changes in a follow-up PR if that's fine with you.

Makefile
alex-chew: Sure - the main flaw was that `git status` uses its output, not its exit code, to signal whether the working tree is clean or dirty. Also, the parenthesized commands would run in a subshell, so the `exit 1;` would only exit out of the subshell - though in this case it doesn't really matter whether `( ... )` or `{ ... }` is used since there are no following commands.

I also realized while writing this that the command *still* wasn't quite correct, because `$` symbols have to be escaped in Makefiles by doubling them...

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
alex-chew: The javac AST represents the null literal as another case constant (whereas this comment mirrors the JLS's choice to separate them) - the comment a few lines below was supposed to indicate this but I'll make it clearer.

alex-chew: There also wasn't a test for that case, but I'll add one now.

robin-aws: Thanks, I initially read it as "no literal present" as opposed to "the literal `null`" :)

verifier/src/test/java/com/aws/jverify/TestVerifier.java
robin-aws: Yeah I realized shortly after making the suggestion that the number of errors isn't necessarily 1:1 with the annotations.

I'm definitely still concerned about this gap, as it means whenever we have a test case we expect to verify successfully (by not annotating an expected error), if it does in fact produce an error, the test will still pass. I think at a minimum we need to search for any line that might be an error. 

It could be worth it to use a more programmatic interface into Dafny so we get structured diagnostics and can fail if we get more than we expect. That seems to be an argument for using the language server command instead of the `dafny verify` CLI. We could still create a new server for each request so we're not affected by any flakiness related to state.

(This is definitely off-topic for this PR though, probably worth cutting a separate issue)

robin-aws (approved):

PR #111, opened 2025-04-23
feat: use structured diagnostics in verifier tests

alex-chew:
### What was changed?

- Refactor the driver to receive structured output from Dafny by passing `--json-diagnostics`, and to return a new `VerificationResults` record (instead of separately returning an exit code and writing all javac/Dafny output to a `Writer`)
  - (The `Writer`-based overload of `verifyJavaFiles` was adapted for use on the command-line code path.)
- Refactor the verifier tests to check expected results using the structured `VerificationResults`
  - (This closes a [testing gap](https://github.com/aws/jverify/pull/110#discussion_r2049327260) that would silently ignore errors that appeared in the verification output but not in the marked-up source.)
- Reduce verifier test boilerplate by parameterizing test methods, and by storing expected exit codes and Dafny verified/error counts directly in the test files' contents
  - (Previously, the `resolutionErrorBooleanTest` method was missing the `@Test` annotation and so it wasn't actually being tested - test parameterization will help detect that kind of mistake)
- Add `DafnyDiagnostic` to adapt Dafny's diagnostics to the `javax.tools.Diagnostic` interface (which `JCDiagnostic` also implements)
- Change `Position` to use 1-indexed line and column numbers, to better align with both Dafny's `Token` and `javax.tools.Diagnostic`

### How has this been tested?

No new functionality was added; the existing tests continue to pass as expected.

### DIFF
--- a/examples/src/main/java/com/aws/verifier/examples/UserProfile.java
+++ b/examples/src/main/java/com/aws/verifier/examples/UserProfile.java
@@ -1,3 +1,7 @@
+// exitCode: 4
+// dafnyVerified: 5
+// dafnyErrors: 1
+
 package com.aws.verifier.examples;
 
 import static com.aws.jverify.JVerify.*;
@@ -21,8 +25,10 @@ class UserProfile {
     @Pure
     @Invariant // Makes this a pre- and post-condition of all public methods
     private boolean valid() {
+//                  ^^^^^ Related location: this is the postcondition that could not be proved
         reads(this);
         return accountType != AccountType.Premium || premiumFeatures != null;
+//             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Related location: this proposition could not be proved
     }
 
     public void upgradeAccount() {
@@ -30,6 +36,7 @@ class UserProfile {
         this.accountType = AccountType.Premium;
         // this.premiumFeatures = new PremiumFeatures();
         return;
+//      ^^^^^^^ Error: a postcondition could not be proved on this return path
     }
 
     public boolean applyTheme(Theme theme) {
--- a/verifier/src/test/java/com/aws/jverify/AssertFalse.java
+++ b/verifier/src/test/java/com/aws/jverify/AssertFalse.java
@@ -1,3 +1,7 @@
+// exitCode: 4
+// dafnyVerified: 0
+// dafnyErrors: 1
+
 package com.aws.jverify;
 
 import static com.aws.jverify.JVerify.*;
--- a/verifier/src/test/java/com/aws/jverify/TestVerifier.java
+++ b/verifier/src/test/java/com/aws/jverify/TestVerifier.java
@@ -19,10 +19,8 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.StringWriter;
-import java.io.Writer;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.List;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
@@ -149,9 +147,11 @@ public class TestVerifier {
     /**
      * Parses test metadata from the given source content.
      * <p>
-     * Note that if verification is expected to fail before invoking Dafny
-     * (i.e. if there are errors in normal Java type-checking or during compilation to Dafny),
-     * then the {@code dafnyVerified} and {@code dafnyErrors} metadata should not appear in the source code.
+     * The {@code exitCode} value is required.
+     * The {@code dafnyVerified} and {@code dafnyErrors} values must either both be present or both be absent
+     * according to whether or not verification is expected to reach the Dafny verification phase.
+     * (That is, if they are present but Dafny is never invoked because there are javac errors,
+     * or if they are absent but Dafny is invoked and terminates normally, then the test should fail.)
      * <p>
      * Example test metadata:
      * {@snippet :
@@ -180,6 +180,9 @@ public class TestVerifier {
         }
         if (exitCode == null) {
             throw new AssertionError("Expected exit code not found in marked source");
+        }
+        if ((dafnyVerified == null) != (dafnyErrors == null)) {
+            throw new AssertionError("Found only one of dafnyVerified and dafnyErrors");
         }
         return new TestMetadata(exitCode, dafnyVerified, dafnyErrors);
     }
--- a/verifier/src/test/java/com/aws/jverify/TestVerifier.java
+++ b/verifier/src/test/java/com/aws/jverify/TestVerifier.java
@@ -22,7 +22,8 @@ public class TestVerifier {
     public void statements() throws IOException {
         verifyMarkedSourceFile("VerifyStatements.java", new DafnyResults(11, 0));
     }
-    
+
+    @Test
     public void resolutionErrorBooleanTest() throws IOException {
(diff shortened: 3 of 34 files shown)


examples/src/main/java/com/aws/verifier/examples/UserProfile.java
```diff
@@ -21,15 +25,18 @@ public UserProfile(AccountType accountType) {
     @Pure
     @Invariant // Makes this a pre- and post-condition of all public methods
     private boolean valid() {
+//                  ^^^^^ Related location: this is the postcondition that could not be proved
```
robin-aws, line 28: Aha, it wasn't just paranoia, we were already missing errors! :)

verifier/src/test/java/com/aws/jverify/AssertFalse.java
```diff
@@ -1,3 +1,7 @@
+// exitCode: 4
+// dafnyVerified: 0
+// dafnyErrors: 1
```
robin-aws, line 3: Just a thought, probably not worth changing at this point: since all Java source files have to define a single public class, we could also use a `@JVerifyExpected(exitCode=4, dafnyVerified=0, dafnyErrors=1)` annotation instead. More editor/refactoring friendly, more typesafe, and possibly easier to parse if you can reuse javac.

verifier/src/test/java/com/aws/jverify/TestVerifier.java
```diff
@@ -1,165 +1,88 @@
...
+            "VerifyNumericOperators.java",
+            "VerifyStatements.java",
+    })
+    public void verifyTestFile(String fileName) throws IOException {
+        testMarkedSource(TEST_FILES_DIRECTORY.resolve(fileName));
+    }
+
+    // Files are hard-coded for now since not all examples can be compiled and run as-is:
+    //  - RuntimePreconditionExample: the try-catch can't be translated yet
+    //  - BinarySearchProperty: the jqwik dependency needs to be added to the classpath
```
robin-aws, line 70: Can we add an exception mechanisms to deny-list tests/examples that aren't working, or move them to a different directory?

(Fine to do this in a follow-up PR, this is already a big improvement!)

verifier/src/test/java/com/aws/jverify/TestVerifier.java
```diff
@@ -172,70 +95,102 @@ public void testRunThroughGradle() throws IOException, InterruptedException {
...
-        String ending = "Dafny program verifier finished with " +
-                dafnyResults.successCount() + " verified, " +
-                dafnyResults.errorCount() + " error" + pluralization + "\n";
-        assertThat(output, endsWith(ending));
+        assertThat("exit code", verificationResults.getExitCode(), is(metadata.exitCode));
+        assertThat("Dafny verified count", verificationResults.getDafnyVerifiedCount(), is(metadata.dafnyVerified));
+        assertThat("Dafny error count", verificationResults.getDafnyErrorCount(), is(metadata.dafnyErrors));
     }
 
+    private static final Pattern TEST_METADATA_PATTERN = Pattern.compile(
```
robin-aws, line 138: For a minute I thought you were parsing XML with a regexp and was mildly horrified. :)

verifier/src/test/java/com/aws/jverify/TestVerifier.java
```diff
@@ -172,70 +95,102 @@ public void testRunThroughGradle() throws IOException, InterruptedException {
...
+    private record TestMetadata(int exitCode, Integer dafnyVerified, Integer dafnyErrors) {}
+
     /**
-     * Verifies the given source code, asserting that each markup annotation appears in the output,
-     * and that the exit code and number of matched error annotations are as expected.
+     * Parses test metadata from the given source content.
+     * <p>
+     * Note that if verification is expected to fail before invoking Dafny
+     * (i.e. if there are errors in normal Java type-checking or during compilation to Dafny),
+     * then the {@code dafnyVerified} and {@code dafnyErrors} metadata should not appear in the source code.
```
robin-aws: This sounds like it should be explicitly checked in the testing framework.

verifier/src/test/java/com/aws/jverify/TestVerifier.java
```diff
@@ -246,25 +201,17 @@ private static String canonicalizeNewlines(final String text) {
...
         var prelude = Path.of("./src/main/resources/additional.dfy");
-        return new VerifierOptions(dafnyPath, libraryJar, prelude, 
+        return new VerifierOptions(dafnyPath, libraryJar, prelude,
                 //Path.of("../temp.dfy"),
                 null,
-                null, true,
-                new String[] {
+                null,
+                true,
+                new String[]{
```
robin-aws, line 217: Thanks - off topic but we should apply an autoformatter for all this Java code.

robin-aws (changes requested):
Love it! [This is the only blocking request.](https://github.com/aws/jverify/pull/111#discussion_r2056487876)

examples/src/main/java/com/aws/verifier/examples/UserProfile.java
alex-chew: Thankfully not - the errors in `examples` tests were being checked by comparing the textual verification output rather than via source markup. I do think there's a case to be made that the examples would look "cleaner" without the test metadata and diagnostic markup, but right now I think it makes sense to prefer consistency across the test suite.

verifier/src/test/java/com/aws/jverify/AssertFalse.java
alex-chew: That's an interesting idea, though Java source files don't have to define a single public class - most of the verifier tests define a single non-public class and `ContractErrors.java` defines multiple non-public classes.

verifier/src/test/java/com/aws/jverify/TestVerifier.java
alex-chew: Yep, I'm planning to move test files around in a separate PR in order to facilitate automatically populating the test parameters.

verifier/src/test/java/com/aws/jverify/TestVerifier.java
alex-chew: It is being explicitly checked - the values must be present and accurate if verification reaches the Dafny phase and they must be absent otherwise - but I'll clarify this in the comment.

verifier/src/test/java/com/aws/jverify/TestVerifier.java
alex-chew: Good idea, I'll address this in a follow-up.

examples/src/main/java/com/aws/verifier/examples/UserProfile.java
robin-aws: Ah missed the fact that the `userProfile()` test was previously doing that. Agreed on consistency/safety being more important, and I actually think if someone's going to read the example without context, it's better to explicitly indicate it's not valid by including the errors.

verifier/src/test/java/com/aws/jverify/AssertFalse.java
robin-aws: Brain-o, meant to say "at least one class" (at most one public etc) so there's always at least one class to attach the annotation to.

verifier/src/test/java/com/aws/jverify/TestVerifier.java
robin-aws: Ah great, it was the "must be absent otherwise" part I wasn't clear on.

robin-aws (approved):

PR #118, opened 2025-04-28
Interfaces

### LINKED ISSUE #52: Support interfaces and extending/implementing interfaces
They're translated to Dafny traits. I think they should be reference type traits (`extends Object`). I think this could have an impact on how we deal with functional interfaces, since they then need to be implemented using classes instead of datatypes, and that might affect verification performance, but for now let's stay close to Java semantics and figure out those performance issues later.

Annotations on base methods, such as `@Pure` and `@Invariant`, must be inherited. Open question: do they need to be respecified??? We could introduce an annotation `@InheritContract` to make the copying of the inherited contract explicit

### Out of scope
- Support for extending classes

keyboardDrummer:
Fixes https://github.com/aws/jverify/issues/52

### What was changed?
- Basic support for interfaces

Created some related issue: https://github.com/aws/jverify/issues/116, https://github.com/aws/jverify/issues/117

### How has this been tested?
- Added Interfaces.java test

### DIFF
--- a/verifier/src/main/java/com/aws/jverify/verifier/DafnyDiagnostic.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/DafnyDiagnostic.java
@@ -112,4 +112,15 @@ public class DafnyDiagnostic implements Diagnostic<String> {
             return diagnostic;
         }
     }
+
+    @Override
+    public String toString() {
+        return "DafnyDiagnostic {" +
+                "\n  location=" + location +
+                "\n  severity=" + severity +
+                "\n  message='" + message + '\'' +
+                "\n  messageSource='" + messageSource + '\'' +
+                "\n  relatedInformation=" + relatedInformation +
+                "\n}";
+    }
 }
--- a/verifier/src/main/java/com/aws/jverify/verifier/Driver.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/Driver.java
@@ -186,7 +186,6 @@ public class Driver {
                 "Binary",
                 "--stdin",
                 "--json-diagnostics",
-                "--allow-warnings",
                 "--type-system-refresh",
                 "--general-newtypes",
                 "--general-traits=datatype"
--- a/verifier/src/main/java/com/aws/jverify/verifier/Driver.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/Driver.java
@@ -185,7 +185,10 @@ public class Driver {
                 "Binary",
                 "--stdin",
                 "--json-diagnostics",
-                "--allow-warnings"
+                "--allow-warnings",
+                "--type-system-refresh",
+                "--general-newtypes",
+                "--general-traits=datatype"
         );
         if (verifierOptions.printDafny() != null) {
             processBuilder.command().add("--print=" + verifierOptions.printDafny());
--- a/verifier/src/main/java/com/aws/jverify/verifier/Driver.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/Driver.java
@@ -18,6 +18,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collector;
 import java.util.stream.Collectors;
 
 public class Driver {
@@ -236,7 +237,7 @@ public class Driver {
         dafnyOutput.lines().forEach(line -> {
             Matcher matcher;
             if (!exceptionOutput.isEmpty()) {
-                exceptionOutput.append(line + "\n");
+                exceptionOutput.append(line).append("\n");
             } else if (line.isBlank()) {
                 //noinspection UnnecessaryReturnStatement
                 return;  // nothing to do
@@ -251,11 +252,12 @@ public class Driver {
                 outResults.dafnyErrorCount = Integer.parseInt(matcher.group("ErrorCount"));
                 outResults.dafnyFinishedMessage = line;
             } else {
-                exceptionOutput.append(line + "\n");
+                exceptionOutput.append(line).append("\n");
             }
         });
         if (!exceptionOutput.isEmpty()) {
-            throw new RuntimeException("Could not parse Dafny output: " + exceptionOutput);
+            String diagnostics = outResults.getDiagnostics().stream().map(Object::toString).collect(Collectors.joining("\n"));
+            throw new RuntimeException("Could not parse Dafny output: " + exceptionOutput + "\n" + diagnostics);
         }
     }
 
--- a/verifier/src/main/java/com/aws/jverify/verifier/Driver.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/Driver.java
@@ -9,6 +9,7 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 import picocli.CommandLine;
 
 import javax.tools.Diagnostic;
+import javax.tools.JavaFileManager;
 import javax.tools.JavaFileObject;
 import java.io.*;
 import java.nio.file.Files;
--- a/verifier/src/main/java/com/aws/jverify/verifier/Driver.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/Driver.java
@@ -232,9 +232,12 @@ public class Driver {
         var objectMapper = new ObjectMapper();
         objectMapper.addMixIn(Position.class, DafnyJsonPosition.class);
 
+        StringBuilder exceptionOutput = new StringBuilder();
         dafnyOutput.lines().forEach(line -> {
             Matcher matcher;
-            if (line.isBlank()) {
+            if (!exceptionOutput.isEmpty()) {
+                exceptionOutput.append(line + "\n");
+            } else if (line.isBlank()) {
                 //noinspection UnnecessaryReturnStatement
                 return;  // nothing to do
             } else if (line.startsWith("{")) {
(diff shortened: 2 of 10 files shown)


verifier/src/test/java/com/aws/jverify/verifier/tests/Interfaces.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    //@InheritContract
+    @Pure
+    @Override
+    public int f(int x) {
+        precondition(x > 2);
+        reads(this);
+        return c.x;
+//             ^^^ Error: insufficient reads clause to read field; Consider adding 'reads c' or 'reads c`x' in the enclosing function specification for resolution
```
robin-aws, line 95: Off-topic, but do we have a backlog item or items for improving the error messages so they make sense (i.e. this one should at least say "add reads(c) or reads(c.x)" instead of "reads c or reads c`x")? That could be quite a lot of work across all the features.

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -184,7 +186,10 @@ public static void runDafnyProcess(
                 "Binary",
                 "--stdin",
                 "--json-diagnostics",
-                "--allow-warnings"
+                "--allow-warnings",
```
robin-aws: Again off topic, but why do we currently need `--allow-warnings`? Given warnings are our only defense against dangerous features like `{:only}` we should prioritize removing this.

verifier/src/main/java/com/aws/jverify/verifier/DafnyDiagnostic.java
```diff
@@ -112,4 +112,15 @@ public DafnyDiagnostic asDiagnostic() {
             return diagnostic;
         }
     }
+
+    @Override
+    public String toString() {
```
robin-aws, line 117: ❤️

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -435,6 +472,9 @@ private boolean isNullable(JCTree.JCModifiers modifiers) {
         shouldVerifies.pop();
         
         if (annotationsByName.containsKey(InheritContract.class.getName())) {
+//            var types = Types.instance(context);
+//            var container = method.sym.enclClass();
+//            var impl = method.sym.implemented(container, types);
```
robin-aws, line 469: Delete?

verifier/src/main/resources/builtin-contracts.java
```diff
@@ -20,3 +20,7 @@ class LongContract {
     public static final long MIN_VALUE = 0x8000000000000000L;
     public static final long MAX_VALUE = 0x7fffffffffffffffL;
 }
+
+@Contract(Object.class)
+class ObjectContract {
```
robin-aws: I assume this is only here so you can have a hard rule that every supertype of a type with a contract must also have a contract?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -327,14 +343,35 @@ private ClassDecl translateClass(Stack<Tree> nestedTypes, JCTree.JCClassDecl cla
...
+        var isInterface = typeForWhichCurrentClassIsDefiningContract == null ? isInterface(classDecl) :
+                isInterface(typeForWhichCurrentClassIsDefiningContract);
+        
         ArrayList<MemberDecl> members = new ArrayList<>();
         for (var member : classDecl.getMembers()) {
             var dafnyMember = translateMember(member, nestedTypes);
             if (dafnyMember != null) {
+                if (dafnyMember instanceof Constructor && isInterface) {
+                    // When a contract class defines the contract for an interface,
+                    // Ignore any constructors
```
robin-aws: Would be better to explicitly error I think, to avoid silent user error

robin-aws (commented):
LGTM, mostly off-topic general questions.

verifier/src/test/java/com/aws/jverify/verifier/tests/Interfaces.java
keyboardDrummer: There's this issue: https://github.com/aws/jverify/issues/34, but it might be a mistake to do this as a batch task, instead of incrementally. That said, I think many of Dafny's errors are more or less 'unlocked' from the start, so they don't map well to particular Java support tasks in JVerify. IMO we need a way to tell Dafny to provide all its output in a structured format, where every diagnostics has an error key and list of parameters, and then JVerify needs to provide messages for all the error keys.

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
keyboardDrummer: Yeah, good point. Turns out I could remove this without issue so I'll do that in this PR

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: Well, we'll need to implement this at some point and now there are some hints there :))) I'll add a comment to state there are future implementation hints.

verifier/src/main/resources/builtin-contracts.java
keyboardDrummer: If you have a type with a contract, then its supertypes must also have contracts, otherwise you get a resolution error at the Dafny level saying it can't find the supertype. That's why we need a contract for Object.

Should there an error at the JVerify level about this? It would be great if we can rely on Dafny's errors as much as possible. I'll add a task to report good errors in case there are missing contracts.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: Added explicit check and test

verifier/src/main/resources/builtin-contracts.java
robin-aws: > Should there an error at the JVerify level about this?

Probably - I agree if we can rely on Dafny verification errors with custom error messages to catch things that's best, but errors where we can't get past Dafny resolution are probably better to handle at the JVerify level.

verifier/src/test/java/com/aws/jverify/verifier/tests/Interfaces.java
```diff
@@ -0,0 +1,107 @@
...
+
+class User {
+
+}
+
+class Container {
+    int x;
+}
+
+class CInvalidContract implements I {
```
robin-aws, line 38: nit, but the name of this class made me think it was an invalid contract for I, but now I see it's a concrete class that fails to follow the contract for I (`IContract`). Consider calling it `CViolatesContract` or something instead.

verifier/src/test/java/com/aws/jverify/verifier/tests/InterfaceResolutionErrors.java
```diff
@@ -0,0 +1,15 @@
...
+
+import com.aws.jverify.Contract;
+
+interface I2 {
+}
+
+@Contract(I2.class)
+class I2Contract {
+    public I2Contract(int x) {
+//         ^ error: a class that defines the contract for an interface may not have explicit constructors
```
robin-aws, line 13: This is great, but made me realize: should we be using interfaces for the contracts of interfaces instead? AFAICT that would just mean you have to add `default` to the method declarations so they can have bodies.

robin-aws (commented):
LGTM, just thought of an alternative to consider before merging.

verifier/src/test/java/com/aws/jverify/verifier/tests/InterfaceResolutionErrors.java
keyboardDrummer: You mention that the cost of using an interface is small, but is there a benefit?

verifier/src/main/resources/builtin-contracts.java
keyboardDrummer: > but errors where we can't get past Dafny resolution are probably better to handle at the JVerify level.

Why? I was planning to use lots of Dafny's resolution errors

verifier/src/test/java/com/aws/jverify/verifier/tests/InterfaceResolutionErrors.java
robin-aws: Eliminating the potential mistake of defining a constructor (since Java itself blocks that on an interface)

robin-aws: Eliminating the potential mistake of defining a constructor (since Java itself blocks that on an interface)

keyboardDrummer: Why would anyone do that? Seems like a niche case to optimize for

keyboardDrummer: There's a more general 'mistake' of defining things in the contract class that don't map to anything in the type to which you're adding a contract, which is not caught by Java in any case.

keyboardDrummer: Using a class has the benefit that Java will force you to add implementations for all members in the interface that you're contracting.

robin-aws: Okay that's a good point, class it is :)

robin-aws (approved):

PR #119, opened 2025-04-29
feat: add JVerifyTestEngine

alex-chew:
### What was changed?

This PR adds the `JVerifyTestEngine` as a [custom JUnit test engine](https://junit.org/junit5/docs/current/user-guide/#test-engines-custom), and migrates the verifier tests and example tests to use it.

IDEs like IntelliJ IDEA should show appropriate gutter icons for test classes marked with `@JVerifyTest`:

Gutter icon for a test class marked with JVerifyTest

(Note that I've left the `// TEST: ...` metadata format as-is for now, but I plan to integrate that into the annotation in a follow-up.)

Both IDEs and Gradle will automatically discover them just like any other JUnit test - for example, clicking "Run 'All Tests'" on the test module in the `verifier` project will populate the test list with both regular `@Test`s and our `@JVerifyTest`s before running all of them:

Running all tests from the test module in the verifier project

Test list including all tests from the test module

(You can also run individual tests from the file tree and/or from previous test runs, of course.)

### How has this been tested?

The existing verifier and example tests are migrated to use this new test engine.


verifier/src/test/java/com/aws/jverify/verifier/tests/ContractErrors.java
```diff
@@ -3,9 +3,13 @@
 package com.aws.jverify.verifier.tests;
 
 import com.aws.jverify.Contract;
+import com.aws.jverify.testengine.JVerifyTest;
 
 import static com.aws.jverify.JVerify.check;
 
+@JVerifyTest
+public class ContractErrors {}
```
robin-aws, line 11: Why the separate dummy class?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -54,8 +57,15 @@ public JavaToDafnyCompiler(Context context) {
         // don't assume the argument is modifiable
         files = new ArrayList<>(files);
         files.add(new SourceFile("builtin-contracts.java", Common.getResourceFile(getClass(), builtinFile)));
-                
-        List<String> javacOptions = List.of("-classpath", options.libraryJar().toAbsolutePath().toString());
+
+        var classpath = String.join(
+                File.pathSeparator,
+                options.libraryJar().toAbsolutePath().toString(),
+                // TODO clean up
```
robin-aws: Any clean up feasible here before merging?

I imagine the proper solution is to rely on the Gradle-defined dependencies of the verifier. You might be able to leverage the "classpath" Gradle task concept as I did in the dafny gradle plugin: https://github.com/dafny-lang/dafny-gradle-plugin/blob/main/src/main/java/org/dafny/gradle/plugin/DafnyPlugin.java#L64

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -0,0 +1,250 @@
...
+        request.getSelectorsByType(DiscoverySelector.class).stream()
+                .flatMap(selector -> switch (selector) {
+                    case ClassSelector classSelector ->
+                            Stream.of(classSelector.getJavaClass()).filter(this::isJVerifyTest);
+                    case ClasspathRootSelector cpRootSelector ->
+                            ReflectionSupport.findAllClassesInClasspathRoot(
+                                    cpRootSelector.getClasspathRoot(), this::isJVerifyTest, _ -> true)
+                            .stream();
+                    default -> {
+                        LOGGER.warning(() -> "Unexpected selector: " + selector);
```
robin-aws, line 72: This needs to be a hard error instead. Otherwise you could have a CLI command to run a specific test succeed, and without any humans noticing that no tests were ran.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -0,0 +1,250 @@
...
+                .forEach(testClass -> {
+                    LOGGER.fine(() -> "Registering test for class: " + testClass);
+                    var id = uniqueId.append("class", testClass.getName());
+                    descriptor.addChild(new JVerifyTestDescriptor(id, testClass));
+                });
+        return descriptor;
+    }
+
+    private boolean isJVerifyTest(Class<?> clazz) {
+        return clazz.isAnnotationPresent(JVerifyTest.class);
```
robin-aws, line 85: So I actually think we should not require an explicit annotation. Otherwise we have the same risk as before of accidentally not running tests.

I tried just changing this to `true` locally and it mostly just worked, the only difference being some classes showed up as Ignored, either due to explicit `// TEST: skip` annotations (which is even better, because it's good for test results to have visibility on skipped tests) or because they were inner classes and hit an assumption (see also my comment on that)

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        @Override
+        public EngineExecutionContext execute(EngineExecutionContext context, DynamicTestExecutor dynamicTestExecutor)
+                throws IOException
+        {
+            LOGGER.fine(() -> "Executing test for class: " + testClass);
+
+            var sourceRoot = Path.of(System.getProperty("user.dir") , "src", "test", "java");
+            Assumptions.assumeTrue(Files.exists(sourceRoot) && Files.isDirectory(sourceRoot),
```
robin-aws: Also should be a hard error (and the assumption below too)

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -0,0 +1,250 @@
...
+
+    @Override
+    public String getId() {
+        return "jverify-test-engine";
+    }
+
+    @Override
+    public TestDescriptor discover(EngineDiscoveryRequest request, UniqueId uniqueId) {
+        var descriptor = new EngineDescriptor(uniqueId, "JVerify Tests");
+        request.getSelectorsByType(DiscoverySelector.class).stream()
```
robin-aws, line 56: Since you're having to look at the source directory below anyway, why not discover all source files in that directory and then filter them to the files that match the discovery selectors? That will solve the problem of non-public classes not working.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    static class JVerifyTestDescriptor
+            extends AbstractTestDescriptor
+            implements Node<EngineExecutionContext>
+    {
+        private final Class<?> testClass;
+
+        JVerifyTestDescriptor(UniqueId uniqueId, Class<?> testClass) {
+            super(uniqueId, testClass.getSimpleName());
```
robin-aws: I would augment the test name to be clearer about what's being tested, and to distinguish it from regular @JUnit tests. Something like "Expected verification result of Foo" (but hopefully shorter) so it's obviously not the same thing as executing Foo as a test itself.

(For the flavour of this I imagine end users using, just "Verify Foo" will be better)

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -0,0 +1,250 @@
...
+            return Type.TEST;
+        }
+
+        @Override
+        public EngineExecutionContext execute(EngineExecutionContext context, DynamicTestExecutor dynamicTestExecutor)
+                throws IOException
+        {
+            LOGGER.fine(() -> "Executing test for class: " + testClass);
+
+            var sourceRoot = Path.of(System.getProperty("user.dir") , "src", "test", "java");
```
robin-aws, line 112: FYI this may be a bit brittle and not work in all environments, but it's good enough for now and worked in my IntelliJ out of the box. When we provide this as an end user feature we may have to expose more configuration. Hopefully we can pick it up from Gradle build scripts somehow.

robin-aws (changes requested):
This is spectacular, very nice work!

We will absolutely want to provide a flavor of this engine for end users as well in the future, just without the metadata about expected errors.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
alex-chew: Without the explicit annotation, IDEs that rely on source-only test discovery (including IntelliJ IDEA) won't show the nice UI features for the test. That is, the gutter won't show the "run test" icon, the linter will flag the class as unused, and the context menu for the file itself won't include test-related actions. During test discovery (such as when running `gradle test` or using the IDE UI to run all tests in a module) the classes are still detected correctly, as you mentioned, but IMO the UI features are the main reason to implement a custom test engine instead of using parametrized/dynamic tests as before.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
alex-chew: I agree that unhandled selectors should raise an error instead - though in practice I've only gotten to this code path with `MethodSelector`, and for that case I'm inclined to automatically select the surrounding class instead (perhaps with a warning to explain that method-level verification isn't supported yet). What do you think?

verifier/src/test/java/com/aws/jverify/verifier/tests/ContractErrors.java
alex-chew: The test engine currently relies on the package and class name to determine the source file path, so I needed to either add a new class or rename one of the existing ones. The existing classes' names seemed fitting for their purposes, so I opted to add the dummy class instead.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
alex-chew: Agreed; this should work in default Gradle configurations (i.e. ones that don't change the working directory for test tasks) and with the conventional source layout, but I'll add a comment to explain that assumption. The JUnit user guide [recommends using system properties](https://junit.org/junit5/docs/current/user-guide/#running-tests-config-params) to configure the test engine, which would be good to implement as follow-up.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
alex-chew: I'm not sure what problem you're referring to - by "non-public classes" do you mean classes named differently that the source file in which they're declared? If so, I don't think this will solve the issue unless we re-parse the source files during test discovery in search of the right `package ...` + `class ...` combination, and I don't love that idea.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
alex-chew: For now I've refactored this logic to read from a new `extraClassPathEntries` field of `VerifierOptions`, pushing the hard-coded path down to `JVerifyTestEngine` and the verifier's `Main` (allowing to run our examples as-is, as one of the `TestVerifier` tests does).

Using the Gradle-defined dependencies sounds promising, though I'm not sure how I'd access those APIs without writing a Gradle plugin too - or did you have something simpler in mind?

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
robin-aws: Ah I didn't try removing the annotations so I didn't realize you lose the editor support, darn.

On reflection I'm much less worried about accidentally skipping tests with this mechanism, since if you forget to add the annotation you can't even run the test locally at all, and the class is rendered as unused.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
robin-aws: Yeah just verifying the whole class with a warning for now seems a bit friendlier.

verifier/src/test/java/com/aws/jverify/verifier/tests/ContractErrors.java
robin-aws: Fair. I notice annotations can also be applied at the package level (https://docs.oracle.com/javase/8/docs/api/java/lang/annotation/ElementType.html#PACKAGE). Does that mean we could support adding `@JVerifyTest` once for the tests package? (If so just fine to add that in a follow up PR)

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
robin-aws: I'm not sure either, but it feels likely there's a way to use that logic outside of a gradle plugin. Can definitely be investigated later.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
robin-aws: Yes sorry, classes with the "wrong" class name. Currently they are silently skipped. Given other discussions I'm happy if they just lead to an error and we don't support them.

alex-chew: They're not silently skipped, in the sense that if you add `@JVerifyTest` then you'll get an error when running the test as a class or when running a test scope that includes the class (e.g. the file itself, the package, or the entire module/project), and if you don't add the annotation then you won't be able to run the test at all and the linter will flag the class as unused.

verifier/src/test/java/com/aws/jverify/verifier/tests/ContractErrors.java
alex-chew: I tried this too but unfortunately IntelliJ doesn't show the classes as testable if the annotation is placed at the package level only.

alex-chew:
Tracking issue for follow-ups: https://github.com/aws/jverify/issues/122

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
robin-aws: It was silently skipped in the version I tried but now that you've made the warning an error it fails, so I'm happy. :)

robin-aws (approved):

PR #121, opened 2025-05-01
Support of (simple forms of) instanceof

olivier-aws:
### What was changed?

Add support of instanceof in JVerify, translates into TypeTestExpr in Dafny.

This depends on a change in Dafny (adding TypeTestExpr to the deserializer). This is the required diff:
```
@@ -6,10 +6,11 @@ public class TypeTestExpr : TypeUnaryExpr, ICloneable<TypeTestExpr> {
   public TypeTestExpr(Cloner cloner, TypeTestExpr original) : base(cloner, original) {
   }
 
-  public TypeTestExpr(IOrigin origin, Expression expr, Type toType)
-    : base(origin, expr, toType) {
+  [SyntaxConstructor]
+  public TypeTestExpr(IOrigin origin, Expression e, Type toType)
+    : base(origin, e, toType) {
     Contract.Requires(origin != null);
-    Contract.Requires(expr != null);
+    Contract.Requires(e != null);
     Contract.Requires(toType != null);
   }
```

### How has this been tested?

Added two tests to Operators.java


verifier/src/test/java/com/aws/jverify/verifier/tests/Operators.java
```diff
@@ -129,4 +131,13 @@ static void EqKO() {
         check(x == y);
 //      ^^^^^^^^^^^^^ Error: assertion might not hold
     }
+
+    void InstanceOfTrivial() {
```
keyboardDrummer, line 146: Can you add a test where `instanceof` returns false as well?

olivier-aws:
Discarding this pull request and opening a new one as I messed up with my local git and easier to start fresh

PR #125, opened 2025-05-02
Upgrade Java toolchain to 23, Gradle to 8.14

robin-aws:
### What was changed?

Java 22 is generally not available to install any more (at least for Corretto).

A newer Gradle is needed to support Java 23, so also upgraded to the latest, and picked up a newer commit of Dafny master that also upgrades Gradle.

Also removed `gradle init` from CI, since this isn't necessary (`gradle init` is for initializing new projects) and newer versions of Gradle started raising errors on it. :)

### How has this been tested?


build.gradle.kts
```diff
@@ -28,7 +28,7 @@ allprojects {
 
 java {
     toolchain {
-        languageVersion.set(JavaLanguageVersion.of(22))
+        languageVersion.set(JavaLanguageVersion.of(23))
```
olivier-aws, line 31: There is also this on line 57:
```
project(":library") {
    java {
        toolchain {
            // Use Java 17 for this subproject, so Java 17 projects can depend on it at compile-/run-time
            languageVersion = JavaLanguageVersion.of(17)
        }
    }
}
```
Shouldn't we change the language version also?

robin-aws: Nope, because that's the minimum language version for projects that use JVerify, and we still want to target 17. 

The value I'm changing here is the version used to build JVerify itself, which doesn't need to be the same.

olivier-aws:
LGTM, better to not depend on a no longer available Java version.

olivier-aws (approved):

PR #138, opened 2025-05-09
Allow new in return statement

olivier-aws:
### What was changed?
Adds support for `return new Foo();` by transforming it into three `var tmp := new Foo(); return tmp` 

### How has this been tested?
Added a test (Allocate.java)

### DIFF
--- a/verifier/src/main/java/com/aws/jverify/verifier/MethodCompiler.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/MethodCompiler.java
@@ -131,8 +131,22 @@ public class MethodCompiler {
         if (expr == null) {
             return List.of(new ReturnStmt(origin, null, null));
         } else {
-            return List.of(new ReturnStmt(origin, null,
-                    List.of(new ExprRhs(compiler.toOrigin(expr), null, compiler.toExpr(expr)))));
+            // Replace
+            //   return e;
+            // by
+            //   var tmp := e;
+            //   return tmp;
+            // so that we can have allocation in e.
+            var exprOrigin = compiler.toOrigin(expr);
+            var returnExpr = compiler.toAssignmentRhs(expr);
+            var newLocalVarName = getTmpVariableName();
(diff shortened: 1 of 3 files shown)


verifier/src/main/java/com/aws/jverify/verifier/MethodCompiler.java
```diff
@@ -131,8 +131,22 @@ private List<Statement> translateReturn(JCTree.JCReturn returnStatement) {
         if (expr == null) {
             return List.of(new ReturnStmt(origin, null, null));
         } else {
-            return List.of(new ReturnStmt(origin, null,
-                    List.of(new ExprRhs(compiler.toOrigin(expr), null, compiler.toExpr(expr)))));
+            // Replace
```
keyboardDrummer, line 134: I think that this one we could also change at the Dafny level: to allow returning things the same expression as you can use in a RRS. That would give slightly better verification performance.

olivier-aws: I don't see the interest in doing it in Dafny. I would say this is the kind of development that we should keep in Java:
1. dev in Java is simple and not risky
2. dev in Dafny will take a few days (changing the AST) with very few benefits for Dafny

keyboardDrummer: You've already done the work so I'll approve this, but the current implementation will impact verification debugging in the future and it'll be unexpected extra work to change it on the Dafny side then. 

> with very few benefits for Dafny

Why would it not benefit Dafny? And if Dafny is our north star language for deductive verification of mutable languages, then JVerify and future languages that strive to have this position benefit from Dafny being great at its role.

> dev in Dafny will take a few days

We have not explored how much work it would be to change it on the Dafny side. A return expression statement is already desugared into an assignment to the return variable, followed by a return statement. I'm not sure why instantiating a class in a return expression is not allowed.

keyboardDrummer (approved):

PR #139, opened 2025-05-10
Support for lambdas (partial, no capturing yet)

robin-aws:
### What was changed?
Adds support for lambda expressions.

* Added `@Modifiable` and now emitting `extends object` when translating a trait only if it has that annotation.
* Removed `@Immutable` as we now don't have immediate plans for it.
* Generalized `translateMethodDecl` to `translateMethod`, which takes the individual pieces of a method declaration so it can be applied to a lambda expression as well.
  * The body can now be either a `JCBlock` or a `JCExpression`, the latter used for lambda expressions with expression bodies such as `(x, y) -> x`. In practice these often can't be used because there's no way to add header method calls like `postcondition`, and we may want to find a way to inherit contracts on lambdas.
* Changed the main `toType` to take a `Type` instead of `JCTree` (i.e. a resolved type rather than a syntactic type reference), since lambda expressions (and probably other features to be supported) don't always have the latter at hand.
* Implemented support for multiple lines of hat annotations beneath a single source line (since supporting lambdas exposed two errors in a single line in an existing test case).

### How has this been tested?


library/src/main/java/com/aws/jverify/Modifiable.java
```diff
@@ -6,12 +6,9 @@
 import java.lang.annotation.Target;
 
 /**
- * Enables the instantiation of this type in a pure context.
- * Can only be used on types that are either:
- * - a record
- * - a sealed interface whose subclasses are all records.
+ * Enables using this interface in reads and modifies clauses.
```
keyboardDrummer, line 9: It's specifically the methods defined in this interface that can then use `this` in reads/modifies clauses, right?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -477,13 +484,34 @@ private boolean isNullable(JCTree.JCModifiers modifiers) {
                 a -> a.getAnnotationType() instanceof JCTree.JCIdent ident && ident.name.contentEquals("Nullable"));
     }
 
+    private boolean isNullable(com.sun.tools.javac.code.Type type) {
+        if (type.getAnnotation(com.aws.jverify.Nullable.class) != null) {
+            return true;
+        }
+
+        if (type instanceof com.sun.tools.javac.code.Type.ArrayType arrayType && isNullable(arrayType.elemtype)) {
```
keyboardDrummer, line 492: Maybe this could use a comment, to explain why the nullability of the element type is used for the array

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -414,7 +421,7 @@ private IndDatatypeDecl translateEnum(JCTree.JCClassDecl classDecl, IOrigin orig
 
     private static boolean isEnum(com.sun.tools.javac.code.Type type) {
         if (type instanceof com.sun.tools.javac.code.Type.ClassType classType) {
-            return ((Symbol.ClassSymbol) classType.supertype_field.tsym).fullname.contentEquals("java.lang.Enum");
+            return classType.supertype_field != null && ((Symbol.ClassSymbol) classType.supertype_field.tsym).fullname.contentEquals("java.lang.Enum");
```
keyboardDrummer, line 424: Can an enum have a `supertype_field`? How does that happen?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -477,13 +484,34 @@ private boolean isNullable(JCTree.JCModifiers modifiers) {
...
+    }
+
     private @Nullable MethodOrFunction translateMethodDecl(JCTree.JCMethodDecl method) {
+        return translateMethod(method, method.getModifiers(), method.sym, method.body);
+    }
+
+    /**
+     * @param sourceBody Either a JCBlock or a JCExpression. The latter is for the benefit of lambda translation.
+     */
+    private @Nullable MethodOrFunction translateMethod(JCTree source, JCTree.JCModifiers modifiers, Symbol.MethodSymbol methodSymbol, JCTree sourceBody) {
```
keyboardDrummer, line 506: Consider renaming to `translateMethodOrLambda`

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -834,6 +876,25 @@ public Expression toExpr(JCTree.JCExpression expr, IOrigin originOverride) {
                 var type = toType(cast.getType());
                 return new ConversionExpr(origin, castExpr, type, "");
             }
+            case JCTree.JCLambda lambda -> {
+                var types = Types.instance(context);
```
keyboardDrummer, line 880: Could you extract this block to a separate method?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -834,6 +876,25 @@ public Expression toExpr(JCTree.JCExpression expr, IOrigin originOverride) {
...
+
+                var datatypeName = "Lambda" + lambdaDatatypeDecls.size();
+                var datatypeNameNode = new Name(origin, datatypeName);
+                var datatypeCtor = new DatatypeCtor(origin, datatypeNameNode, null, false, List.of());
+                var trait = toType(lambda.target, origin);
+                var datatypeDecl = new IndDatatypeDecl(origin, datatypeNameNode, null, List.of(), List.of(methodDecl),
+                        List.of(trait), List.of(datatypeCtor), false);
+                lambdaDatatypeDecls.add(datatypeDecl);
+
+                // TODO: Using a DatatypeValue directly ends up crashing when printing temp.dfy,
```
keyboardDrummer, line 893: `DatatypeValue` seems to be a resolved construct, so I think this `TODO` can be removed, as well as the `[SyntaxConstructor]` attribute on `DatatypeValue` in the Dafny code.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -1160,32 +1220,26 @@ public Type toType(JCTree tree, boolean isNullable, @Nullable IOrigin originOver
...
 
 
-            reportError(tree, "notSupported", "Primitive type kind %s".formatted(primitiveTypeKind));
+            reportError(origin, "notSupported", "Primitive type kind %s".formatted(primitiveTypeKind));
             return null;
-        } else if (tree instanceof JCTree.JCArrayTypeTree arrayTypeTree) {
-            var elemType = toType(arrayTypeTree.getType(), true, originOverride);
+        } else if (type instanceof com.sun.tools.javac.code.Type.ArrayType arrayTypeTree) {
+            // TODO: Assuming nullable here means it's not possible to have non-nullable array elements?
+            var elemType = toType(arrayTypeTree.elemtype, true, origin);
```
keyboardDrummer, line 1227: I think the `true` here is temporary, although syntactically it's not clear how to mark the array elements as nullable. It would be nice if it was possible. IMO the added TODO does not clarify the current code, so I'd leave it out.

keyboardDrummer (approved):
Awesome!!! Love it. Left some comments but also approved. Feel free to address in a follow-up if you prefer.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
robin-aws: I found `supertype_field` was sometimes null rather than a pointer to `java.lang.Object`, so this would crash. I believe that only happened when I experimented with running more of the javac transformations, but it seemed safe enough to make this change anyway.

library/src/main/java/com/aws/jverify/Modifiable.java
robin-aws: Or methods defined elsewhere:

```java
interface Foo {
  void mutate();
}

@Contract(Foo.class)
@Modifiable
class FooContract implements Foo {
  void mutate() {
    modifies(this);
  }
}

void someOtherMethod(Foo f) {
  modifies(f);

  f.mutate();
}

```

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
robin-aws: Aha, that explains why the printer choked on it earlier - you shouldn't ever have an unresolved `DatatypeValue` in the first place. Thanks for clearing that up, removing the comment.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
robin-aws: Actually as it turns out `@Nullable int[]` can be interpreted as a nullable array of ints, and `int @Nullable []` can be an array of nullable ints! I'll make that change in the next PR so this is fully resolved.

library/src/main/java/com/aws/jverify/Modifiable.java
keyboardDrummer: Er yes, I meant method contracts defined for this interface.

Issue #134, opened 2025-05-07
Enable mutation and non-determinism in expressions that occur in a statement context


robin-aws:
Java expressions can also do things like `new` objects and call other non-pure methods, but Dafny expressions can't. A rewriting phase that generically flattens expression trees into sequences of intermediate statements would avoid this.

e.g.

```java
var result = method(foo(1), bar(foo(2)));
```

becomes

```java
var tmp0 = foo(1);
var tmp1 = foo(2);
var tmp2 = bar(tmp1);
var result = method(tmp0, tmp2);
```

fabiomadge:
Closing: Dafny backend has been removed. The underlying concept may be revisited for the Strata backend if needed.

Issue #127, opened 2025-05-06
Change the number of reported verified things, so it matches Java concepts


keyboardDrummer:
Currently the number of verified things reported does not match Java methods.

PR #140, opened 2025-05-12
Add a user guide

keyboardDrummer:
To view the guide, checkout this PR and follow the `docs/guide/README` instructions.

I've already configured the GH pages for this repository to be private.

### What was changed?
- Add a user guide for JVerify. I have tried to order the topics in a way that you can read it top down, and have tried to use examples that not only show how features work, but also what benefit one can get out of using them. I have tried to introduce the topics in an order that allows you to unlock the simplest use-cases early on.
- Added a GHA file that publishes the documentation to the GH pages of this repository

#### Besides the content, what I would like feedback on
- There does not seem to be an easy way to run JVerify. The guide tells user to run:
```
./verifier/build/install/verifier/bin/verifier \
  ./examples/src/test/java/com/aws/jverify/examples/BinarySearch.java --dafny ./dafny/Scripts/dafny
```
Should I add a script to the root of the repository to reduce some of the paths? In a separate PR though.
- Can I add the examples in a less nested path? It would be nicer if they were just under `./examples/` or `./examples/src`. Should change in a separate PR.

### How has this been tested?
- Three of the code pieces in the guide are references to example files, and these are tested
- TODO. There are many small samples in the guide that are currently not tested. Most are partial: they are missing package/import statements to make it valid code. I think I would need access to a full programming language to extract from actually tested sources, all the code/JVerify output that I want to show in the docs.


.vscode/tasks.json
robin-aws, line 1: Just noticing this file. I'm assuming it's currently just for jverify developers, but could it be useful for jverify users too?

INSTALL.md
```diff
@@ -0,0 +1,11 @@
+JVerify can currently only be used by building it from source. 
+
+1. Ensure the dotnet runtime version 8 or higher is installed on your machine. More information can be found [here](https://dotnet.microsoft.com/en-us/).
+1. Check out the JVerify repository and its submodules, for example using `git clone https://github.com/aws/jverify --recurse-submodules`.
+1. Navigate to the JVerify repository directory.
+1. Run `gradle installDist`
+1. Run `dotnet build dafny/Source/Dafny.sln` to build Dafny, a tool used by JVerify.
```
robin-aws:
```suggestion
1. Run `make -C dafny exe` to build [Dafny](https://dafny.org/), a tool used by JVerify.
```

docs/guide/src/code_purposes.md
```diff
@@ -0,0 +1,59 @@
+# Specification, proof and implementation
+
+This section does not introduce any features, but it defines terminology that we commonly use to talk about Java programs using JVerify. When writing Java code that uses JVerify, each piece of code serves one or several of the following goals:
+
+- Implementation, the Java code that will actually execute, and may be optimized for performance at the cost of readability.
+- Specification, used to specify the behavior of a piece of implementation. Will not be executed.
+- Proof, used to verify that an implementation adheres to a specification and that it does not throw unchecked exceptions. Should not be executed.
```
robin-aws, line 7: I still feel the line between these concepts is too fuzzy. The loop invariants feel much more like they are specifying the behavior of the loop rather than proving the loop body adheres to that spec.

docs/guide/src/code_types.md
```diff
@@ -0,0 +1,63 @@
...
+## Pure code
+
+Code that occurs in a contract, such as the arguments of calls to `precondition`, `postcondition` and `check`, must follow the rules for pure code. These are:
+- Only methods annotated with `@Pure` can be called.
+- Local variables or fields can not be updated.
+
+Code that occurs in a method annotated with `@Pure` must follow those same rules, and additionally:
+- Only the following statements can be used:
+  - variable declaration statements,
+  - calls to `precondition`, `postcondition`, `reads` and `assert`
```
robin-aws:
```suggestion
  - calls to `precondition`, `postcondition`, `reads`, and `check`
```

docs/guide/src/getting_started.md
```diff
@@ -0,0 +1,44 @@
+# Getting Started
+
+To get started with JVerify, first follow the directions in the [Installation manual](https://github.com/verus-lang/verus/blob/main/INSTALL.md) to build it.
```
robin-aws:
```suggestion
To get started with JVerify, first follow the directions in the [Installation manual](https://github.com/aws/jverify/blob/main/INSTALL.md) to build it.
```

Although I'm tickled by "to get started, first give up on Java and use Rust instead" :D

examples/src/test/java/com/aws/jverify/examples/UserProfile.java
```diff
@@ -32,7 +32,7 @@ private boolean valid() {
     public void upgradeAccount() {
         modifies(this);
         this.accountType = AccountType.Premium;
-        // this.premiumFeatures = new PremiumFeatures();
+        this.premiumFeatures = new PremiumFeatures();
         return;
 //      ^^^^^^^ Error: a postcondition could not be proved on this return path
```
robin-aws:
```suggestion
```

INSTALL.md
```diff
@@ -0,0 +1,11 @@
+JVerify can currently only be used by building it from source. 
+
+1. Ensure the dotnet runtime version 8 or higher is installed on your machine. More information can be found [here](https://dotnet.microsoft.com/en-us/).
+1. Check out the JVerify repository and its submodules, for example using `git clone https://github.com/aws/jverify --recurse-submodules`.
+1. Navigate to the JVerify repository directory.
+1. Run `gradle installDist`
```
robin-aws:
```suggestion
1. Run `./gradlew installDist` (for Mac OS or Linux systems) or `./gradlew.bat installDist` (for Windows)
```

docs/guide/src/supported_java_features.md
robin-aws, line 1: I'd put the Supported features first, just to be slightly more positive :)

docs/guide/src/supported_java_features.md
```diff
@@ -0,0 +1,80 @@
...
+- verification of annotations
+- Any classes that rely on concurrent analysis currently can not be analyzed by JVerify
+
+# Supported Java standard library packages
+
+Nothing yet
+
+# Supported Java Features
+
+- all literals excepts strings
```
robin-aws:
```suggestion
- all literals except strings
```

docs/guide/src/supported_java_features.md
```diff
@@ -0,0 +1,80 @@
...
+- string template: `STR."My name is \{name}"`
+- `throw`
+- `try`/`catch`/`finally`
+- `checked exceptions` (ignored)
+- `synchronized`
+- module declarations
+- `?` wildcards
+- `permits`
+- verification of annotations
+- Any classes that rely on concurrent analysis currently can not be analyzed by JVerify
```
robin-aws: More like "All code is assumed to be not used by multiple threads"

docs/guide/src/supported_java_features.md
```diff
@@ -0,0 +1,80 @@
...
+- Array instantiation with initializers: `new int[] { 1, 2, 3 }`
+- type casts: `(SomeType)someValue`
+- type parameters for types, including the declaration, declaration of bounds, and application
+- `extends`: but only classes inheriting other classes
+- enhanced for loop: `for(var x : xs) { ... }`
+- use of mutating operators such as `=` in a nested expression
+- operators on `float` and `double`
+- `instanceof <pattern>`, only when using a pattern
+- string literal: `"hello"`
+- string template: `STR."My name is \{name}"`
```
robin-aws:
```suggestion
- string templats: `STR."My name is \{name}"`
```

docs/guide/src/supported_java_features.md
```diff
@@ -0,0 +1,80 @@
+# Unsupported Java features
+
+- `switch` statement
+- `switch` expression
```
robin-aws: This is at least partially supported according to Switches.java

docs/guide/src/supported_java_features.md
```diff
@@ -0,0 +1,80 @@
...
+- `yield` in switch expressions
+- record pattern in switch expression
+- lambda expressions
+- method references
+- import statement
+- package statement
```
robin-aws:
```suggestion
- import statements
- package statements
```

docs/guide/src/supported_java_features.md
```diff
@@ -0,0 +1,80 @@
...
+- Multi-dimensional array instantiation: `new [,]`
+- Array instantiation with initializers: `new int[] { 1, 2, 3 }`
+- type casts: `(SomeType)someValue`
```
robin-aws:
```suggestion
- Type casts: `(SomeType)someValue`
```

And similarly starting all of them with capitals consistently.

docs/guide/src/getting_started.md
```diff
@@ -0,0 +1,44 @@
...
+found [here](https://github.com/aws/jverify/blob/main/examples/src/test/java/com/aws/jverify/examples/BinarySearch.java):
+
+```java
+{{#include ../../../examples/src/test/java/com/aws/jverify/examples/BinarySearch.java}}
+```
+
+To run JVerify, run the following from the repository root:
+
+```
+./verifier/build/install/verifier/bin/verifier ./examples/src/test/java/com/aws/jverify/examples/BinarySearch.java --dafny ./dafny/Scripts/dafny
```
robin-aws, line 25: It would be really nice not to have to specify `--dafny` unless you're using a copy other than the submodule, could we default `--dafny` to `./dafny/Scripts/dafny`?

examples/src/test/java/com/aws/jverify/examples/BinarySearch.java
```diff
@@ -17,14 +15,14 @@ static boolean sorted(int[] arr) {
...
-        precondition(arr.length <= 0x7fff_ffff /* Integer.MAX_VALUE */);
+        precondition(arr.length <= Integer.MAX_VALUE);
         precondition(sorted(arr));
         postcondition((Integer res) ->
                 (res == -1 && !sequence(arr).contains(key))
                 || (0 <= res && res < arr.length && arr[res] == key)
         );
 
-        var lo = 0;
+        var lo = 1;
```
robin-aws:
```suggestion
        var lo = 0;
```

docs/guide/src/working_with_null.md
```diff
@@ -0,0 +1,27 @@
+# Working with null
+
+JVerify allows detecting all possible `NullReferenceException`s in your Java code. To make it easier to check this, JVerify requires that any reference variable which can be null, is annotated with `@Nullable`. Example:
```
robin-aws:
```suggestion
JVerify allows detecting all possible `NullReferenceException`s in your Java code. To make it easier to check this, JVerify requires that any reference variable which can be null is annotated with `@Nullable`. Example:
```

docs/guide/src/code_purposes.md
```diff
@@ -0,0 +1,59 @@
+# Specification, proof and implementation
+
+This section does not introduce any features, but it defines terminology that we commonly use to talk about Java programs using JVerify. When writing Java code that uses JVerify, each piece of code serves one or several of the following goals:
+
+- Implementation, the Java code that will actually execute, and may be optimized for performance at the cost of readability.
```
robin-aws:
```suggestion
- Implementation, the Java code that will actually execute.
```

The rest feels a bit too judgmental and I don't think it's relevant.

robin-aws (changes requested):
Fantastic start, great to get this in place and then incrementally improve, especially as more support is added. This is one batch of suggestions mostly at a high-level, I still need to review all the actual content.

My take on the open questions:
* Let's have the installation instructions in the guide instead of an INSTALL.md for now at least. Having INSTALL.md as a separate link that works is kind of a one way door, we can extract it out later if needed.
*  I wouldn't add any extra scripts for running for now, I think it will be more worthwhile to find smoother ways of using jverify in the Java build process instead.
*  For cleaning up the examples and testing the examples/snippets, I suspect a bit of custom mdbook scripting will be worthwhile. I notice the Rust book has a custom `#rustdoc_include` macro for e.g.

docs/guide/src/getting_started.md
robin-aws: Or perhaps to just `dafny`, assuming it's on the path somewhere? That will also make it easier to support `dotnet tool` installs of Dafny.

docs/guide/src/code_purposes.md
robin-aws: We can definitely brainstorm more on this, but I like to think of the fundamental split being regular code va ghost code. Ghost code can define specifications of correct runtime behavior, and the verifier can check if they always hold. Instead of talking about "proof" code, it's more just that adding more details to the specifications, down to individual check statements, can help the verifier.

robin-aws:
Another thought: it might be slightly better to give `examples` it's own dedicated `build.gradle.kts` with the content currently in the root build script. That would implicitly help document the typical build script changes needed to add JVerify to a project.

docs/guide/src/code_purposes.md
keyboardDrummer: If you would generalize the language to allow for jumping to arbitrary labels, then you'd also have to specify an invariant for each label. Would you then call this the behavior of the label? 

IMO the method body, and the jumps that occur in it, is all part of "implementation" and we specify invariants to enable us to analyze it, with the purpose of proving that it matches the method's specification

docs/guide/src/supported_java_features.md
keyboardDrummer: The way I see it is that there are contracts missing on the APIs that create concurrency, and JVerify currently can not specify those contracts.

Within verified code, I would not say JVerify assumes anything about concurrency. When calling verified from unverified code though, users have to respect the verified contract, and the contract will often require arguments not to be modified concurrently.

Updated the text.

.vscode/tasks.json
keyboardDrummer: Yes, it's useful for people using JVerify to develop. Might be nicer to expose it through a VSCode extension, although I'm not sure. Left this out of scope for this PR

docs/guide/src/supported_java_features.md
keyboardDrummer: Nice

docs/guide/src/getting_started.md
keyboardDrummer: > could we default

But that once works of the working directly is the repository root. Only letting it work automatically if that is the case, seems like it might cause more confusion than it helps.

I've mentioned the option to configure the environment variable `JVERIFY_DAFNY`

> Or perhaps to just dafny, assuming it's on the path somewhere?

It does default to that if nothing else is specified.

keyboardDrummer:
> Another thought: it might be slightly better to give examples it's own dedicated build.gradle.kts with the content currently in the root build script.

~Great idea. I'll do that.~
Actually, how do I refer to the library dependency then? Can I only do that once we publish it?

.vscode/tasks.json
robin-aws: I'll see how nicely my annotation processor verifier plays with VSCode as well :)

docs/guide/src/getting_started.md
robin-aws: > It does default to that if nothing else is specified.

Ah great! Agreed it's better to keep the explicit argument here though, since it's simpler to not require users to mess with their PATH as long as we have the submodule.

docs/guide/src/SUMMARY.md
```diff
@@ -0,0 +1,30 @@
+# Summary
+
+[JVerify overview](./overview.md)
```
robin-aws:
```suggestion
[JVerify Overview](./overview.md)
```

(And likewise consistent title-case on all the titles)

docs/guide/src/SUMMARY.md
```diff
@@ -0,0 +1,30 @@
...
+- [Loop Invariants](loop_invariants.md)
+- [Pre- and post-conditions](pre_and_postconditions.md)
+- [Code types](code_types.md)
+- [Quantifiers](quantifiers.md)
+- [Code purposes](code_purposes.md)
+- [Termination](termination.md)
+- [Integers](integers.md)
+
+# References
+- [Modifying fields](modifies.md)
```
robin-aws:
```suggestion
# References

- [Modifying fields](modifies.md)
```

docs/guide/src/overview.md
```diff
@@ -0,0 +1,11 @@
+# JVerify
+
+JVerify is a tool that at compile-time can detect most bugs in a Java program. It uses computer-aided theorem proving to prove that the program never throws any uncaught exceptions. If provided with program specifications, it can prove that a program will satisfy those for any possible execution.
```
robin-aws:
```suggestion
JVerify is a tool that can detect most bugs in a Java program at compile-time. It uses computer-aided theorem proving to prove that the program never throws any uncaught exceptions. If you specify properties about how your code should behave, such as what states a field is allowed to be null in, it can prove that a program will satisfy those for any possible execution.
```

docs/guide/src/overview.md
```diff
@@ -0,0 +1,11 @@
...
+
+Program specifications are provided by making calls to the JVerify library. These calls can be removed during compilation, using a plugin for `javac`, so they will not have an effect at run-time. Java code that contains JVerify specifications is still regular Java code, so it can be analyzed by any Java IDE.
```
robin-aws:
```suggestion
Program specifications are provided by making calls to the JVerify library. These calls will be removed during compilation, using a plugin for `javac`, so they will not have an effect at run-time. Java code that contains JVerify specifications is still regular Java code, so it can be developed using any Java IDE.
```

docs/guide/src/getting_started.md
```diff
@@ -0,0 +1,56 @@
+# Installation
+JVerify can currently only be used by building it from source.
+
+1. Ensure the dotnet runtime version 8 or higher is installed on your machine. More information can be found [here](https://dotnet.microsoft.com/en-us/).
```
robin-aws:
```suggestion
1. Ensure the .NET runtime version 8 or higher is installed on your machine. More information can be found [here](https://dotnet.microsoft.com/en-us/).
```

docs/guide/src/getting_started.md
```diff
@@ -0,0 +1,56 @@
...
+1. Check out the JVerify repository and its submodules, for example using `git clone https://github.com/aws/jverify --recurse-submodules`.
```
robin-aws:
```suggestion
1. Check out the [JVerify repository](https://github.com/aws/jverify) and its submodules, for example using `git clone https://github.com/aws/jverify --recurse-submodules`.
```

docs/guide/src/getting_started.md
```diff
@@ -0,0 +1,56 @@
...
+1. Navigate to the JVerify repository directory.
+1. Run `./gradlew installDist` (for Mac OS or Linux systems) or `./gradlew.bat installDist` (for Windows)
+1. Run `make -C dafny exe` to build [Dafny](https://dafny.org/), a tool used by JVerify.
```
robin-aws:
```suggestion
1. Run `make -C dafny exe` to build [Dafny](https://dafny.org/), a verification tool used by JVerify.
```

docs/guide/src/getting_started.md
```diff
@@ -0,0 +1,56 @@
...
+1. Install Z3 using `make -C dafny z3-<OS>`, where `<OS>` can be `mac`, `mac-arm` or `ubuntu`. Pick the one that matches your OS. For systems that do not fit those categories, such as Windows, you will have to take these steps instead:
+    1. Download the release of [Z3 version 4.12.1](https://github.com/Z3Prover/z3/releases/tag/z3-4.12.1) that matches your system.
+    1. Unzip the content of the Z3 folder into dafny\Binaries\z3 so that dafny\Binaries\z3\bin\z3.exe exists.
```
robin-aws:
```suggestion
    1. Unzip the content of the Z3 folder into `dafny\Binaries\z3` so that `dafny\Binaries\z3\bin\z3.exe` exists.
```

examples/src/test/java/com/aws/jverify/examples/BinarySearch.java
```diff
@@ -4,9 +4,7 @@
 
 import com.aws.jverify.Erased;
 import com.aws.jverify.Pure;
-import com.aws.jverify.testengine.JVerifyTest;
 
-@JVerifyTest
```
robin-aws, line 9: Remove the `// TEST` metadata comment too (it shows up in the user guide snippet)

docs/guide/src/getting_started.md
```diff
@@ -0,0 +1,56 @@
...
+
+You should see the following output:
+
+```
+Dafny program verifier finished with 3 verified, 0 errors
+```
+
+This indicates that JVerify successfully verified 3 parts.
+
+You can remove the need for the `--dafny` option by configuring the environment variable `JVERIFY_DAFNY` instead.
```
robin-aws:
```suggestion
You can remove the need for the `--dafny` option by configuring the environment variable `JVERIFY_DAFNY` instead, or modifying your `PATH` so that `dafny` resolves automatically.
```

?

docs/guide/src/getting_started.md
```diff
@@ -0,0 +1,56 @@
...
+./verifier/build/install/verifier/bin/verifier.bat ./examples/src/test/java/com/aws/jverify/examples/BinarySearch.java --dafny ./dafny/Scripts/dafny
+```
+
+You should see the following output:
+
+```
+Dafny program verifier finished with 3 verified, 0 errors
+```
+
+This indicates that JVerify successfully verified 3 parts.
```
robin-aws:
```suggestion
This indicates that JVerify successfully verified 3 batches of specifications.
```

(Hard to do much better before #127 though)

docs/guide/src/supported_java_features.md
```diff
@@ -0,0 +1,79 @@
...
+- String template: `STR."My name is \{name}"`
+- `throw`
+- `try`/`catch`/`finally`
+- `checked exceptions` (ignored)
+- `synchronized`
+- Module declarations
+- Wildcards: `?`
+- `permits`
+- Verification of annotations
+- For now, JVerify can only specify contracts that require none of the arguments are modified concurrently. Support for concurrent modification will be added. When calling verified code from unverified code, be careful not to pass in objects that are being modified concurrently.
```
robin-aws:
```suggestion
- For now, JVerify can only specify contracts that require none of the arguments or `this` are modified concurrently. Support for concurrent modification will be added. When calling verified code from unverified code, be careful not to pass in objects that are being modified concurrently.
```

docs/guide/src/supported_java_features.md
```diff
@@ -0,0 +1,79 @@
...
+- `yield` in switch expressions
+- Lambda expressions
+- Method references
+- `import` statements
+- `package` statements
+- Multi-dimensional array instantiation: `new [,]`
+- Array instantiation with initializers: `new int[] { 1, 2, 3 }`
+- Type casts: `(SomeType)someValue`
+- Type parameters for types, including the declaration, declaration of bounds, and application
+- `extends`: but only classes inheriting other classes
```
robin-aws, line 43:
```suggestion
- `extends`, but only classes inheriting other classes
```

docs/guide/src/uncaught_exceptions.md
```diff
@@ -0,0 +1,43 @@
...
+
+For example, JVerify emits an error for this program:
+```java
+class Foo {
+    int divZero(int x) {
+        if (x > 0) {
+            return 10 / x; // no error
+        }
+        return 20 / x;
+//              ^^^^^^ error: possible ArithmeticException
```
robin-aws:
```suggestion
//             ^^^^^^ error: possible ArithmeticException
```

docs/guide/src/uncaught_exceptions.md
```diff
@@ -0,0 +1,43 @@
...
+    }
+}
+```
+
+While in the above cases, JVerify is correct in detecting that the program has a bug, there are more complex cases where the program is correct, but JVerify can not prove this without hints, so it emits an error. For example here:
```
robin-aws, line 19:
```suggestion
In the above case, JVerify is correct in detecting that the program has a bug. There are more complex cases where the program is correct, but JVerify can not prove this without hints, so it emits an error. For example here:
```

docs/guide/src/uncaught_exceptions.md
```diff
@@ -0,0 +1,43 @@
...
+          } else {
+              return mid;
+          }
+      }
+      return -1;
+  }
+}
+```
+
+In the following section, [Loop Invariants](loop_invariants.md), we learn how to prove that programs with loops do not throw uncaught exceptions. Later on, in the section [Working with null](working_with_null), we'll look at how to verify programs that work with `null` while proving no `NullReferenceException` can occur.
```
robin-aws:
```suggestion
In the following section, [Loop Invariants](loop_invariants.md), we learn how to prove that programs with loops do not throw uncaught exceptions. Later on, in the section [Working with null](working_with_null.md), we'll look at how to verify programs that work with `null` while proving no `NullReferenceException` can occur.
```

docs/guide/src/working_with_null.md
```diff
@@ -0,0 +1,27 @@
...
+  }
+  
+  static void caller() {
+    nonNull(null); // error: could not prove that argument 'null' is not null
+    nullable(null); // pass
+  }
+}
+```
+
+We'll show a more complicated program where JVerify helps with detecting null dereferences, in the next section, [Class Invariants](class_invariants.md).
```
robin-aws, line 26: Doesn't seem to be the next section

docs/guide/src/working_with_null.md
```diff
@@ -0,0 +1,27 @@
+# Working with null
```
robin-aws, line 1: Worth mentioning how this works with arrays? It seems like `@Nullable int[] a` means a nullable reference to an array, but the array elements are always `@Nullable` and I can't find a way to specify they aren't.

docs/guide/src/time_traveling.md
```diff
@@ -0,0 +1,38 @@
+# Time traveling
```
robin-aws, line 1: Love the title :)

docs/guide/src/time_traveling.md
```diff
@@ -0,0 +1,38 @@
...
+
+JVerify allows evaluating an expression at a point in time different from where the expression occurs. This is most commonly used in postconditions to compare the program state before and after the method executed.
```
robin-aws, line 3:
```suggestion
JVerify allows evaluating an expression at a point in time different from where the expression occurs. This is most commonly used in post-conditions to compare the program state before and after the method executed. 
```

docs/guide/src/pre_and_postconditions.md
```diff
@@ -0,0 +1,94 @@
+# Pre- and post-conditions
```
robin-aws:
```suggestion
# Preconditions and postconditions
```

Only because we don't seem to hyphenate the terms anywhere else.

docs/guide/src/loop_invariants.md
```diff
@@ -0,0 +1,52 @@
+# Loop Invariants
+
+JVerify verifies the behavior of code with loops through the use of a _loop invariant_. Since JVerify does not know how often a loop iterates, it needs to work with information that is true for _any_ iteration. The loop invariant is a condition that must be true every time the loop guard is evaluated, so when the loop is entered, and after each execution of the body.
```
robin-aws, line 3:
```suggestion
JVerify verifies the behavior of code with loops through the use of a _loop invariant_. Since JVerify does not know how often a loop iterates, it needs to work with information that is true for _any_ iteration. The loop invariant is a condition that must be true every time the loop guard is evaluated, so when the loop is first entered, and after each execution of the body.
```

docs/guide/src/loop_invariants.md
```diff
@@ -0,0 +1,52 @@
...
+Then JVerify emits the error:
+```
+error: this invariant could not be proved on entry
+ | invariant(lo <= hi);
+             ^^^^^^^^ 
+```
+
+In this section, we've used invariants to prove that no uncaught exceptions occur inside the loop.
+
+In the next sections, [Pre- and post-conditions](pre_and_postconditions.md), [Code types](code_types.md), [Forall and exists](forall_and_exists.md), we will introduce the tools needed to prove the correctness of the above `binarySearch` example, and we will use invariants also to prove facts that hold after the loop exits.
```
robin-aws, line 52:
```suggestion
In the next sections, [Pre- and post-conditions](pre_and_postconditions.md), [Code types](code_types.md), [Forall and exists](forall_and_exists.md), we will introduce the tools needed to prove the correctness of the above `binarySearch` example, and we will also use invariants to prove facts that hold after the loop exits.
```

docs/guide/src/pre_and_postconditions.md
```diff
@@ -0,0 +1,94 @@
...
+
+Besides detecting exceptions, JVerify allows you to precisely state what a method does, and check that it actually does that. We can prove that the binary search method shown in the previous section behaves as intended by giving it a contract, which is done using calls to `precondition` and `postcondition`.
+
+Here’s the updated example. Note that this example contains calls to several JVerify utility methods that we have not seen before: `sequence`, `contains`, `drop` and `take`. 
+
+`sequence` converts an array to a JVerify type `Sequence` which is similar to an immutable Java `List`, and is useful for verification. On a `Sequence` we can call `contains` which behaves as you'd expect. `drop` and `take` create a `Sequence` from an array, with elements selected as you'd expect.
```
robin-aws: > `drop` and `take` create a `Sequence` from an array

I'm not seeing these versions in the current state of the library, I only see overloads like `sequence(array, fromIndex)`, and `Sequence.take/drop`. I think that's the better API, just the text here and the example needs updating.

docs/guide/src/pre_and_postconditions.md
```diff
@@ -0,0 +1,94 @@
...
+  int binarySearch(int[] arr, int target) {
+    // This postcondition guarantees that the method behaves as desired
+    postcondition((Integer r) -> sequence(arr).contains(target)
+      ? 0 <= r && r < arr.length && arr[r] == target
+      : r == -1
+    );
+
+    // The following precondition states that the input arr must be sorted.
+    // Without this, we won't be able to prove the two new loop invariants, 
+    // which are needed to prove the postcondition.
```
robin-aws, line 30: More like "This is an implicit assumption this method makes: it doesn't necessarily behave correctly otherwise..."

i.e. it's not just that we won't be able to prove something without that precondition, it's that you should never actually call this method otherwise.

docs/guide/src/pre_and_postconditions.md
```diff
@@ -0,0 +1,94 @@
...
+    {
+      invariant(0 <= left);
+      invariant(left <= right);
+      invariant(right <= arr.length);
+      invariant(!drop(arr, left).contains(target)); // added
+      invariant(!take(arr, right).contains(target)); // added
+
+      var mid = (left + right) / 2;
+      if (arr[mid] == target) {
+          // we can prove sequence(arr).contains(target)
```
robin-aws: Maybe an actual `check` statement here even?

docs/guide/src/pre_and_postconditions.md
```diff
@@ -0,0 +1,94 @@
...
+          // and 0 <= mid && mid < arr.length && arr[mid] == target
+          return mid;
+      }
+      if (arr[mid] < target) {
+          left = mid + 1; 
+          // we use the preconditon to prove the 4th invariant
```
robin-aws: "we ... prove ..." here doesn't actually make sense, "we" as the code author aren't doing anything, it's more like JVerify applying this reasoning.

docs/guide/src/pre_and_postconditions.md
```diff
@@ -0,0 +1,94 @@
...
+    }
+    // we use the 4th and 5th invariant to prove that !sequence(arr).contains(target)
+    return -1;
+  }
+}
+```
+
+If we now change anything in the implementation of `binarySearch` that would change its meaning, JVerify detects that the implementation no longer matches the specification and emits an error.
+
+We had to add the call to `precondition` to be able to prove the `postcondition`, but this also has the desired effect of preventing us from incorrectly calling `binarySearch`. The method only behaves as intended if called with a sorted list, and this is now checked by JVerify, as you can see here:
```
robin-aws: Ah this is what I was looking for, maybe pull this statement up earlier.

docs/guide/src/pre_and_postconditions.md
```diff
@@ -0,0 +1,94 @@
...
+
+```java
+  void callBinarySearch() {
+    binarySearch(new int[] {4, 1, 5}, 1);
+//              ^ error: could not prove precondition
+    binarySearch(new int[] {1, 3, 5}, 1); // no error
+  }  
+```
+
+## Improvements
```
robin-aws: I think this section would fit more nicely at the beginning of the next section instead. You could just say "In the next section we'll improve the code by extracting out the sorted concept so it's self-documenting and reusable"

docs/guide/src/code_types.md
```diff
@@ -0,0 +1,63 @@
+# Code types
+All code in a Java program that uses JVerify can be assigned two properties: whether it is pure, and whether it is erased. The following two sections discuss what these properties are for and how they work.
+
+## Pure code
+
+Code that occurs in a contract, such as the arguments of calls to `precondition`, `postcondition` and `check`, must follow the rules for pure code. These are:
```
robin-aws: I'd first say simply what "pure" means in a more intuitive sense, perhaps "deterministic and side-effect free"

docs/guide/src/code_types.md
```diff
@@ -0,0 +1,63 @@
...
+- Only methods annotated with `@Pure` can be called.
+- Local variables or fields can not be updated.
```
robin-aws: Will we ever support mutable local variables by changing the translation?

docs/guide/src/code_types.md
```diff
@@ -0,0 +1,63 @@
...
+
+Code that occurs in a method annotated with `@Pure` must follow those same rules, and additionally:
+- Only the following statements can be used:
+  - variable declaration statements,
+  - calls to `precondition`, `postcondition`, `reads`, and `check`
+  - a single `return <expr>` statement at the end, in which other pure methods may be called
```
robin-aws, line 40: This at least is something we'll improve on by translating to LetExprs I hope? We can link to the supported features list to clarify better.

docs/guide/src/code_types.md
```diff
@@ -0,0 +1,63 @@
...
+- If any fields of a reference are read, that reference must occur in a `reads` call. We'll explain what reads calls are in the next section.
```
robin-aws:
```suggestion
- If any fields of a reference are read, that reference must occur in a `reads` call. We'll explain what reads calls are below. 
```

(Since we're generally using "section" to mean separate pages)

docs/guide/src/code_types.md
```diff
@@ -0,0 +1,63 @@
+# Code types
+All code in a Java program that uses JVerify can be assigned two properties: whether it is pure, and whether it is erased. The following two sections discuss what these properties are for and how they work.
+
+## Pure code
```
robin-aws, line 24: I think we also have to mention below that arguments to method calls have to be pure (at least for now)

docs/guide/src/code_types.md
```diff
@@ -0,0 +1,63 @@
...
+        return this.engine.horsePower;
+    }
+}
+``` 
+
+## Erased code
+
+In an earlier example of a binary search program, we saw that a Java method can be given the annotation `@Erased`. This annotation change the rules that are applied to the code inside that method, and change where the method may be called.
+
+Any code inside a method annotated with `@Erased`, or that is part of a contract, is _erased_, meaning it will not be executed at runtime. Erased code is used purely for writing JVerify specifications and proofs. Because it is not executed, it may perform operations that would be costly to execute, such as evaluating `forall` and `exists` expressions, and working with unbounded numbers.
```
robin-aws:
```suggestion
Any code inside a method annotated with `@Erased`, or that is part of a contract, is _erased_, meaning it will not be executed at runtime. Erased code is used purely for writing JVerify specifications and proofs. Because it is not executed, it may perform operations that would be costly or impossible to execute, such as evaluating `forall` and `exists` expressions.
```

Removing the reference to unbounded numbers since we haven't hit that topic, and it's too easy to ask "but what about BigInteger?" :)

docs/guide/src/code_types.md
```diff
@@ -0,0 +1,63 @@
...
+
+| **Effect \ Context**                       | **@Erased method** | **executed method** |
+|--------------------------------------------|--------------------|---------------------|
+| **Can be called from an executed context** | no                 | yes                 |
+| **Can use forall and exists**              | yes                | no                  |
+| **Can use @Unbounded**                     | yes                | no                  |
+
+Code inside a contract is always erased.
+
+In the next section, [Quantifiers](quantifiers.md), we'll explain how the `forall` method works, as well an introduce `exists`.
```
robin-aws: Maybe provide a link to Integers since you mention @Unbounded above as well.

robin-aws:
```suggestion
In the next section, [Quantifiers](quantifiers.md), we'll explain how the `forall` method works, as well as introduce `exists`.
```

docs/guide/src/code_types.md
```diff
@@ -0,0 +1,63 @@
...
+    Engine engine;
+    
+    public int readHorsePower() {
+        reads(this, engine); // without this line, the next line would emit an error
+        return this.engine.horsePower;
+    }
+}
+``` 
+
+## Erased code
```
robin-aws, line 72: Might be worth mentioning this is similar to how Java already erases generics?

docs/guide/src/quantifiers.md
```diff
@@ -0,0 +1,20 @@
+# Quantifiers
+JVerify introduces a special kind of expression called a quantifier. There are two types of quantifiers, `forall` and `exists`. Both take a lambda that can take any number of arguments of any type. The `forall` states that for all inputs the lambda could possibly take, the lambda body must be true, and the `exists` states that there must exist an input to the lambda for which the body is true.
+
+Given a `forall` that ranges over types with infinitely many elements, such as a linked list, we can only ever determine if the forall holds by using logical reasoning. Similar, for an `exists` expression with an infinite domain, we can only prove that it is false by using logical reasoning. For these reasons, these types of expression do not occur in regular Java. `forall` and `exists` can only be used in an erased context, since these can not be executed at runtime.
```
robin-aws: I wouldn't mention linked lists as an example, because that feels like it implies a given linked list value might have infinite elements. Perhaps "all possible array values", alluding back to the binary search running example?

Perhaps "infinitely many values" instead too.

docs/guide/src/quantifiers.md
```diff
@@ -0,0 +1,20 @@
...
+
+You can use quantifiers to say something about the behavior of methods. For example, given an argument of type `Function`, we can say that this function only returns values greater than 10:
+```
```
robin-aws:
```suggestion
```java
```

docs/guide/src/quantifiers.md
```diff
@@ -0,0 +1,20 @@
...
+void higherOrder(f: Function<Integer, Integer>) {
+  precondition(forall((Integer i) -> f(i) > 10);
+}
+```
+
+Or that a `List` only contains even values 
+```
```
robin-aws:
```suggestion
```java
```

docs/guide/src/code_purposes.md
```diff
@@ -0,0 +1,59 @@
+# Specification, proof and implementation
+
+This section does not introduce any features, but it defines terminology that we commonly use to talk about Java programs using JVerify. When writing Java code that uses JVerify, each piece of code serves one or several of the following goals:
+
+- Implementation, the Java code that will actually execute.
+- Specification, used to specify the behavior of a piece of implementation. Will not be executed.
+- Proof, used to verify that an implementation adheres to a specification and that it does not throw unchecked exceptions. Should not be executed.
```
robin-aws, line 7: Why does proof have "Should not" be executed , but specification has "Will not"?

docs/guide/src/code_purposes.md
```diff
@@ -0,0 +1,59 @@
...
+      } else {
+          right = mid;
+      }
+    }
+    return -1;
+  }
+
+  // specification
+  @Pure
+  @Erased
```
robin-aws, line 51: Worth calling out that this definition of `sorted` should definitely only be used for specifications, since it would be quadratic at runtime. But there are cases where you can have a pure but executed method that is also useful for specifications.

docs/guide/src/termination.md
```diff
@@ -0,0 +1,119 @@
+# Termination
+
+Programs that contain loops or recursive calls may run indefinitely. Usually, such behavior is undesired.
+
+By default, JVerify tries to prove that methods terminate, and emits an error if it can't. Even if a program terminates, JVerify might need you input to help prove that it does. In case you do not want to, you can turn off termination checking.
```
robin-aws, line 5: Accurate for now, but I really think we should weaken this to not be on by default soon.

docs/guide/src/termination.md
```diff
@@ -0,0 +1,119 @@
...
+
+Note that if comparing tuples of different sizes, Dafny appends an imaginary 'top' value to the smaller tuple until they are of equal size. The top value is larger than any other value, and can not be expressed explicitly. In the above example of `methodA` had specified `decreases(x)` instead of `decreases(x,1)`, then instead of the `1` the top value would have been used when comparing to `0`, so the result would have been the same.
+
+## Non-terminating Code
+
+In some cases, you might intentionally write code that doesn't terminate (like a server loop). JVerify allows you to specify this with `decreases()`:
+
+```java
+void serverLoop() {
+    decreases();
```
robin-aws: I think we should have a different library method for this. `decreases()` just naturally feels like you're given an empty lexicographic tuple instead. Perhaps `mayNotTerminate()`?

robin-aws:
> > Another thought: it might be slightly better to give examples it's own dedicated build.gradle.kts with the content currently in the root build script.
> 
> ~Great idea. I'll do that.~ Actually, how do I refer to the library dependency then? Can I only do that once we publish it?

I think you can have the examples build script point to the library project on disk, let me experiment.

docs/guide/src/termination.md
```diff
@@ -0,0 +1,119 @@
...
+
+Programs that contain loops or recursive calls may run indefinitely. Usually, such behavior is undesired.
+
+By default, JVerify tries to prove that methods terminate, and emits an error if it can't. Even if a program terminates, JVerify might need you input to help prove that it does. In case you do not want to, you can turn off termination checking.
+
+This page further explains how to prove termination using JVerify.
+
+## Decreases clauses
+
+JVerify needs a decreases clause for each method and loop. If none is specified explicitly, JVerify will guess one. A decreases clause contains an expression. JVerify proves termination by verifying that this expression satisfied two rules:
```
robin-aws:
```suggestion
JVerify needs a decreases clause for each method and loop. If none is specified explicitly, JVerify will guess one. A decreases clause contains an expression. JVerify proves termination by verifying that this expression satisfies two rules:
```

docs/guide/src/termination.md
```diff
@@ -0,0 +1,119 @@
+# Termination
+
+Programs that contain loops or recursive calls may run indefinitely. Usually, such behavior is undesired.
+
+By default, JVerify tries to prove that methods terminate, and emits an error if it can't. Even if a program terminates, JVerify might need you input to help prove that it does. In case you do not want to, you can turn off termination checking.
+
+This page further explains how to prove termination using JVerify.
+
+## Decreases clauses
```
robin-aws, line 9: I think you need to introduce the term "termination metric" a little earlier, and be more explicit that it's kind of untyped. I worry folks will assume you can't call a method if the termination metrics use different types or widths of tuples.

You can also then clarify what "smaller" means, that it includes "less than" for the same type but also things like shorter tuple prefixes.

docs/guide/src/code_purposes.md
keyboardDrummer: > We can definitely brainstorm more on this, but I like to think of the fundamental split being regular code va ghost code. Ghost code can define specifications of correct runtime behavior, and the verifier can check if they always hold. Instead of talking about "proof" code, it's more just that adding more details to the specifications, down to individual check statements, can help the verifier.

How can you put the contract (pre and postconditions), which fully describes the intended behavior of the code, and non-contract erased code (calc statement/check statement/lambda calls/loop invariants), which is only needed if you are verifying that the implementation matches the contract, into the same category?

docs/guide/src/termination.md
```diff
@@ -0,0 +1,119 @@
...
+
+JVerify needs a decreases clause for each method and loop. If none is specified explicitly, JVerify will guess one. A decreases clause contains an expression. JVerify proves termination by verifying that this expression satisfied two rules:
+
+1. Decreasing: the expression becomes smaller with each recursive call or loop iteration
+2. Lower bound: the expression only recurses/loops if at or above a lower bound. The value of the lower bound depends on the type of the expression. For `int`, the lower bound is 0.
```
robin-aws:
```suggestion
2. Has a lower bound: the expression only recurses/loops if at or above a lower bound. The value of the default lower bound depends on the type of the expression. For `int`, the default lower bound is 0.
```

docs/guide/src/termination.md
```diff
@@ -0,0 +1,119 @@
...
+        methodA(x - 1);
+    }
+}
+```
+
+Here, the lexicographically ordered tuples ensure termination:
+- For the call from `methodA(x)` to `methodB(x)`, the tuple `(x, 0)` is smaller than `(x, 1)`
+- For the call from `methodB(x)` to `methodA(x-1)`, the tuple `(x-1, 1)` is smaller than `(x, 0)`
+
+Note that if comparing tuples of different sizes, Dafny appends an imaginary 'top' value to the smaller tuple until they are of equal size. The top value is larger than any other value, and can not be expressed explicitly. In the above example of `methodA` had specified `decreases(x)` instead of `decreases(x,1)`, then instead of the `1` the top value would have been used when comparing to `0`, so the result would have been the same.
```
robin-aws: I'd not even mention the "top" concept, and just say that for any given tuple, any tuple formed by adding more elements to the right is considered smaller.

docs/guide/src/integers.md
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+```
+
+Without the `@Nat` annotation, JVerify would not have been able to prove termination, and indeed the program does not terminate if `n` is smaller than zero. Instead of `@Nat`, we could also have used `precondition(n >= 0)`, but that is more verbose.
+
+## Bounded numbers
+The types `byte`, `int`, `short` and `long` in Java are bounded. The values of these numbers all have lower and upper bounds, and doing arithmetic operations that cause these numbers to surpass these bounds, called an overflow, generally causes undesired results.
+
+JVerify will check that any arithmetic operations on bounded number will not overflow. Example:
```
robin-aws:
```suggestion
JVerify will check that any arithmetic operations on bounded numbers will not overflow. Example:
```

docs/guide/src/supported_java_features.md
robin-aws, line 1: Why don't we state that this list is actively changing for the better, and invite users to cut GitHub issues if they hit unsupported features not in the list?

docs/guide/src/SUMMARY.md
```diff
@@ -0,0 +1,30 @@
+# Summary
+
+[JVerify overview](./overview.md)
+
+# Getting started
+
+- [Getting started](./getting_started.md)
+- [Supported Java features](./supported_java_features.md)
+
+# Fundamentals
```
robin-aws, line 10: Should call this Tutorial, or change the ""JVerify Tutorial and Reference" book title to match.

I think I like calling it a tutorial, since you mostly stick to the binary search running example.

docs/guide/src/class_invariants.md
robin-aws, line 1: We should mention that invariants don't apply to private methods.

verifier/src/test/java/com/aws/jverify/verifier/TestExamples.java
robin-aws, line 1: Good enough for now!

docs/guide/src/code_purposes.md
robin-aws: I'd go as far as saying a `check` statement is a specification of what states the program is allowed to be in at a particular spot. Saying specification is only about method boundaries feels arbitrary.

The other quibble I have is that you never really write a proof in JVerify (or Dafny for that matter), you just state more assertions until the tool is able to infer a proof successfully, and it never actually shows you a step-by-step proof.

But I'm willing to disagree and commit on this, I don't think this is harmful and could be tweaked later.

robin-aws (changes requested):
Whew! Excellent stuff.

docs/guide/src/code_purposes.md
keyboardDrummer: Would you ever place a check statement in an implementation, for another reason than to make a proof go through? I would argue you should not. The contract should fully specify the behavior that you want, so there is no reason to specify behavior in any other way.

To me, a `check` statement outside of a `by` block is an anti-pattern. I think it would be better to warn on these. 

> The other quibble I have is that you never really write a proof in JVerify (or Dafny for that matter), you just state more assertions until the tool is able to infer a proof successfully, and it never actually shows you a step-by-step proof.

I think for a method all the proof for the postcondition clause should be at the end, after the executed code. The proofs for preconditions of statements should be in by blocks for those statements. If you write it like that, you can point at a section and say this is the proof for that goal.

I would like to re-evaluate for JVerify the language constructs that are available to help proofs go through: `check statement`/`calc statement`/`by blocks`, to see if we can make their usage more intuitive. For example, I also feel a by block with more than one check statement is bad, because then you don't know whether one of those is meant to help prove the other, or whether the conjunction of them is meant to prove the by block.

I wonder whether we can set up the language constructs so the user is guided more towards building a proof tree, with the goal at the root, assumptions as leafs and intermediate goals as other nodes.

Something like:
```
var x := prove(<expressionWithAPrecondition).
  by(<intermediateGoal0>).
  by(<callToALemma>).
  by(prove(<intermediateGoal1>).
         by(<intermediateGoal1,0>).
         by(<intermediateGoal1,1>).
         done()).
  done();
```

Also since it would be annoying to have to use a lot of lambda's to tie proofs to particular statements, we could consider a `proveNext` method that serves as a proof for the next statement, so you could do:

```
method foo() {
  postcondition(...)
  ...
  proveNext(...) // proof for the subset constraint
  @Nat int x = ...

  proveNext(...) // proof for the invariant
  while(..) {
    invariant(...)

    if (...) {
       proveNext(...) // proof for the invariant
       continue;
    }
    ...
    proveNext(...) // proof for the invariant
  }
  ...
  proveNext(...); // proof for the postcondition
  return;
}
```

docs/guide/src/code_purposes.md
keyboardDrummer: The contracts are not executed because JVerify will erase them. I think you could add code that is purely intended to prove something, but that is not erased, so you  could accidentally have proofs that are also executed.

docs/guide/src/code_purposes.md
keyboardDrummer: My intention was for JVerify not to allow using quantifiers in non-erased code, so you have no choice but to use @Erased here. Do you think we should allow it? I think determining a domain based on the predicate is pretty complicated and I do not see the use-case.

docs/guide/src/code_types.md
keyboardDrummer: I think that for transparent pure methods, we may never support non-functional code, because it is messy to translate and such a translation would be bad for verification performance.

My intention is not to allow `@Pure` methods to specify postconditions unless they add a `@Opaque` modifier, which hides the body and makes it unrevealable. If a method is pure and opaque, then my hope is that we can allow arbitrary use of statements as long as the method remains deterministic.

docs/guide/src/code_types.md
keyboardDrummer: Good callout

docs/guide/src/code_types.md
keyboardDrummer: Will do

docs/guide/src/code_types.md
keyboardDrummer: What's the improvement you'd like? I didn't follow.

docs/guide/src/pre_and_postconditions.md
keyboardDrummer: That implies that JVerify needs it, which would be more verbose than it actually is.

docs/guide/src/pre_and_postconditions.md
keyboardDrummer: "We" as in the programmer and JVerify? The proofs generally take input from both, right?

docs/guide/src/termination.md
keyboardDrummer: I do like that it's a free contract, like proving the absence of uncaught exceptions. If we turn it off by default then it's easy to forget, although non-termination does not seem to be a common bug.

docs/guide/src/termination.md
keyboardDrummer: What do you mean by default lower bound? It's not something that can be customized. It's always 0 for `int`

docs/guide/src/termination.md
keyboardDrummer: > I worry folks will assume you can't call a method if the termination metrics use different types or widths of tuples.

If there is (mutual) recursion, then you can not do a call if the types are different, right?

docs/guide/src/time_traveling.md
keyboardDrummer: I thought you decided we should use postcondition everywhere. What do you prefer?

docs/guide/src/SUMMARY.md
keyboardDrummer: I'm not a fan of the header "Fundamentals", but "Tutorial" would incorporate the "References" grouping that occurs below as well. Any other ideas?

docs/guide/src/working_with_null.md
keyboardDrummer: It's currently a little buggy. Indeed the element should not be always nullable but it currently is. It seems the array itself can be made nullable with `int atNullable []`.

docs/guide/src/code_purposes.md
robin-aws: > My intention was for JVerify not to allow using quantifiers in non-erased code

Ah interesting, for Dafny I think it was very useful to support executing quantifiers because there wasn't another good option for that kind of logic, but in JVerify you've got enhanced for loops and the streams API etc which are almost as expressive. And more importantly a lot of the time that code will already exist and you just need to add `@Pure` to the method to use it in a contract. So I think not supporting execution of quantifiers this is the right call, especially since it's a two-way door.

I think the point still stands that there will be pure methods like this that are used for both implementation and specification, but you already make the point that code can be used for multiple goals.

docs/guide/src/code_types.md
robin-aws: Oh I missed "variable declaration statements" at first, that's what I was looking for.

We can still support more but I think #134 covers the biggest gap.

docs/guide/src/termination.md
robin-aws: IMO it's not that free because the default decreases clauses don't tend to be enough for a codebase in practice, especially if you have mutable state and have to use the Repr idiom.

We should probably have a way to explicitly state that a method always terminates, separately from `decreases(...)` (but perhaps implied by it), so that users can opt in to proving it with one line, and sometimes that's all they will have to do.

docs/guide/src/termination.md
robin-aws: You definitely can: https://github.com/dafny-lang/dafny/blob/master/Source/DafnyStandardLibraries/examples/TerminationExamples.dfy#L94-L114.

docs/guide/src/time_traveling.md
robin-aws: Oh I probably suggested this earlier and then changed my mind later on, disregard

docs/guide/src/SUMMARY.md
robin-aws: How about we leave "Fundamentals" for now but make the book title "JVerify User Guide"?

docs/guide/src/code_purposes.md
robin-aws: Okay this I love. :) You're pointing to the right kind of changes we need to make for "proof" to be a real concept, so I'm happy leaving this part of the user guide as aspirational. I cut #145 for follow up.

robin-aws:
> > > Another thought: it might be slightly better to give examples it's own dedicated build.gradle.kts with the content currently in the root build script.
> > 
> > 
> > ~Great idea. I'll do that.~ Actually, how do I refer to the library dependency then? Can I only do that once we publish it?
> 
> I think you can have the examples build script point to the library project on disk, let me experiment.

This is a bit of a can of worms, so I cut #147 for follow-up immediately after alpha.

docs/guide/book.toml
```diff
@@ -0,0 +1,11 @@
+[book]
+language = "en"
+multilingual = false
+src = "src"
+title = "JVerify Tutorial and Reference"
```
robin-aws:
```suggestion
title = "JVerify User Guide"
```

robin-aws (approved):

PR #144, opened 2025-05-14
Support method overloading

olivier-aws:
### What was changed?

Add support of method overloading in JVerify. This change adds a transformation pass that updates each method and field name of Java classes into a unique name, based on their type.

### How has this been tested?


verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
```diff
@@ -0,0 +1,113 @@
...
+ * Update the name of all method and field names of symbols within a compilation unit
+ * to ensure uniqueness. The mangling algorithm is deterministic and works as follows:
+ *   - a prefix ("_F") is added to a field name
+ *   - a prefix ("_Z") is added to method names (except constructors)
+ *   - a suffix is added to all method names. This suffix encodes the parameter types, as given by the typeMangling
+ *     method
+ */
+public class NameMangler {
+    private final JavaToDafnyCompiler javaToDafnyCompiler;
+    private Names nameFactory;
```
robin-aws:
```suggestion
    private final Names nameFactory;
```

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
```diff
@@ -0,0 +1,113 @@
...
+    static private final String fieldPrefix = "F_";
+    static private final String methodPrefix = "Z_";
+
+    // Cache of all symbols whose names were already mangled.
+    private HashSet<Symbol> mangledSymbols;
```
robin-aws:
```suggestion
    private final Set<Symbol> mangledSymbols;
```


(More Java idiomatic to use the general collections interfaces like `Set` rather than specific implementations when that detail doesn't matter, which is almost all the time)

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -652,7 +660,7 @@ private void applyInvariants(JCTree.JCMethodDecl method, HeaderContainer header)
         boolean isPublic = (method.getModifiers().flags & Flags.PUBLIC) != 0;
         if (isPublic) {
             for(var invariant : invariants) {
-                if (!method.name.contentEquals("<init>")) {
+                if (!TreeInfo.isConstructor(method)) {
```
robin-aws, line 664: Nice, I was trying to find this myself earlier :)

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
```diff
@@ -0,0 +1,113 @@
...
+        if (mangledSymbols.contains(s)) {
+            return s.name.toString();
+        }
+        String newName;
+        if (s.type instanceof MethodType m) {
+            newName = getMethodName(s);
+        } else { // TODO: ensure this is a field?
+            newName = getFieldName(s);
+        }
+        s.name = nameFactory.fromString(newName);
```
robin-aws: I worry about this stateful approach hurting maintainability in the future. If translating elements to Dafny mutates those elements, I can easily imagine hard to diagnose bugs due to reading a symbol more than once and getting different values without realizing it.

I think it would be cleaner for this class to maintain a `Map<Symbol, String>` and munge symbols as they are used to generate Dafny symbols, rather than writing to the AST in place.

If we do want to have name mangling mutate the Java AST, this is probably better done as a separate pass (perhaps as a TreeScanner) before starting to translate the AST to Dafny. But it may be quite tricky to make sure we change every place a symbol occurs in the tree.

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
```diff
@@ -0,0 +1,113 @@
...
+            baseName = "_ctor";
+        }
+        else {
+            baseName = methodPrefix + baseName;
+        }
+        MethodType methodType = null;
+        if (s.type instanceof MethodType m) {
+            methodType = m;
+        } else {
+            assert(false);
```
robin-aws: Much better in general to throw an exception. `assert` statements in Java are no-ops if you don't pass the `-ea` JVM option, so you'll just get a null pointer exception below that's not as helpful.

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
```diff
@@ -0,0 +1,113 @@
...
+            case VOID : {return "v";}
+            case BYTE : {return "b";}
+            case CHAR : {return "c";}
+            case SHORT: {return "s";}
+            case INT : {return "i";}
+            case LONG : {return "l";}
+            case FLOAT :{return "f";}
+            case DOUBLE : {return "d";}
+            case BOOLEAN: {return "z";}
+            case CLASS: {return "C" + ((ClassType)type).toString();}
```
robin-aws: This will be the fully-qualified class name with dots, which won't be a valid symbol class name in Dafny, so this won't work in general. I don't think we have any class-typed parameters in our examples yet which is why this doesn't break anything. :)

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -758,15 +771,17 @@ private Expression toExpr(JCTree.JCExpression expr) {
         } else if (expr instanceof JCTree.JCFieldAccess fieldAccess) {
             var selectedExpr = toExpr(fieldAccess.selected);
             // TODO does this work if the selected expression isn't trivially of array type?
-            if (fieldAccess.selected.type instanceof ArrayType && fieldAccess.name.contentEquals("length")) {
+            if (fieldAccess.selected.type instanceof ArrayType && fieldAccess.sym.name.contentEquals("length")) {
```
robin-aws: This is exactly the kind of subtlety I worry about below - I assume this is necessary because we changed `sym.name` but there are other copies of that symbol we don't update?

robin-aws (changes requested):
I think it's inevitable that we will have to mangle names, but I have concerns with the exact approach here. We may want to chat offline together about it.

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
keyboardDrummer: @olivier-aws what does it mean to change the name of a symbol? Would that also change the name of the associated `JCTree` ? We must keep the AST in a consistent state.

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
```diff
@@ -0,0 +1,113 @@
...
+        var argTypes = methodType.getParameterTypes();
+        baseName = (argTypes.isEmpty()) ? baseName : baseName+"_";
+        for (var param : argTypes) {
+            baseName += typeMangling(param);
+        }
+        return baseName;
+    }
+
+    public void mangleNames(Tree tree) {
+        if (tree instanceof JCTree.JCClassDecl classDecl) {
```
keyboardDrummer: I would invert the condition, adding an exit return, to avoid nesting in the remaining code

olivier-aws: Regarding changing the symbol name, the reason I suggested this is that Symbols are shared between the method/field declaration and their usage. For example, if the tree `t1` is the declaration of a method `foo` and `t2` is the call to this method, then t1 is `MehodDecl m1` and t2 is `JCMethodInvocation (JCTree.JCIdent ident)` where `m1.sym = ident.sym` (= meaning that there are the same object instance) and `my.sym.name = "foo"`

So changing `m1.sym.name` also changes `indent.sym.name`. We would then just need to go through all declarations of methods and fields, change the name of the symbols, and during translation we just need to read the names of the symbols rather than the trees to have a correct mangling.

Regarding the `indent.name` itself, from what I understood of the javac code, we have that `indent.name=indent.sym.name` according to this snippet of code from TreeMaker.java:

```
public JCIdent Ident(Symbol sym) {
        return (JCIdent)new JCIdent((sym.name != names.empty)
                                ? sym.name
                                : sym.flatName(), sym)
            .setPos(pos)
            .setType(sym.type);
    }
```

This however falls short for symbols declared in a separate compilation unit, because we treat every compilation unit independently. So to make a transformation pass that would update all symbol names we would need to apply it on the whole tree after linking of all compilation units.

I'll thus move to a different approach that does not modify the Tree

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
olivier-aws: Removed

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
olivier-aws: I changed to a stateless approach that does not modify the `JCTree` and does the mangling on the fly. See latest commit.

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
olivier-aws: Done

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
olivier-aws: I rewrote the whole logic, please see latest commit

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -733,10 +748,11 @@ private Expression toExpr(JCTree.JCExpression expr) {
             Symbol.OperatorSymbol operator = binary.getOperator();
             return translateBinary(binary, binary.type, binary.getLeftOperand().type, operator, left, right);
         } else if (expr instanceof JCTree.JCIdent identifier) {
-            if (identifier.name.contentEquals("this")) {
+            var identName = true ? nameMangler.getName(expr) : identifier.name.toString();
```
olivier-aws:
```suggestion
            var identName = nameMangler.getName(expr);
```

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -150,6 +154,7 @@ private FileStart translateFile(JCTree.JCCompilationUnit compilationUnit) {
         remainingTypes.addAll(compilationUnit.getTypeDecls());
         while(!remainingTypes.isEmpty()) {
             var typeDecl = remainingTypes.pop();
+            //this.nameMangler.mangleNames(typeDecl);
```
```suggestion
```

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
```diff
@@ -0,0 +1,136 @@
...
+            case VOID : {return "v";}
+            case BYTE : {return "b";}
+            case CHAR : {return "c";}
+            case SHORT: {return "s";}
+            case INT : {return "i";}
+            case LONG : {return "l";}
+            case FLOAT :{return "f";}
+            case DOUBLE : {return "d";}
+            case BOOLEAN: {return "z";}
+            case CLASS: {return "C" + ((ClassType)type);}
```
robin-aws: AFAICT this still has the same problem I pointed out earlier where the mangled name will have dots in it, which won't be valid syntax when used to build the result of `mangleMethodName` for e.g. (the toString() is just implicit now)

I think I would just replace dots with underscores here. And add a test case to expose the problem first. :)

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
```diff
@@ -0,0 +1,136 @@
...
+ *   - a prefix ("_Z") is added to method names (except constructors)
+ *   - a suffix is added to all method names. This suffix encodes the parameter types, as given by the typeMangling
+ *     method
+ */
+public class NameMangler {
+    static private final String fieldPrefix = "F_";
+    static private final String methodPrefix = "Z_";
+
+    // Map from symbols to mangled names
+    private Map<Symbol, String> symbolStringMap;
```
robin-aws:
```suggestion
    private final Map<Symbol, String> symbolStringMap;
```

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }
+
+    public NameMangler() {
+        this.symbolStringMap = new HashMap<>();
+    }
+
+
+    private String mangleSymbolName(Symbol s) {
+        var name = symbolStringMap.get(s);
```
robin-aws: Minor nit (more to be aware of this in general, good pattern for using a Map as a cache): you could tidy this up a little by using `Map.computeIfAbsent`:

```java
private String mangleSymbolName(Symbol s) {
    return symbolStringMap.computeIfAbsent(s, this::uncachedMangleSymbolName);
}

private String uncachedMangleSymbolName(Symbol s) {
    if (s instanceof Symbol.MethodSymbol m) {
    ...
}
```

robin-aws (changes requested):
Much better approach! Just a few suggestions.

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
olivier-aws: Yes I still need to handle this case.

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
olivier-aws: I was not aware of this API, love it. Thanks for the tip.

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
olivier-aws: I did change the '.' to '_' in the class name, and added a test case. More testing will be needed with more complex class names e.g. when we support generic type instantiating.

robin-aws (approved):
🚀

Issue #156, opened 2025-05-15
Support @Verify on package declarations (package-info.java)


robin-aws:
More involved because we aren't yet handling those source files, and the data in a single CompilationUnit doesn't include ancestor package declarations.

PR #157, opened 2025-05-15
--verify-by-default CLI option, more tests for @Verify

### LINKED ISSUE #131: Finish gradual verification
- Enable having `@Verify(false)` as a default
- Add tests for the `@Verify` annotation

robin-aws:
### What was changed?
Resolves #131.

Added a `--verify-by-default` option, (defaulting to true), so that you can verify only elements with explicit `@Verify` annotations on them.

I attempted to implement support for `@Verify` on packages, which is another useful way to turn verification on or off for large amounts of code, but this is more involved since then we have to process `package-info.java` files, so instead I disabled the ability to put `@Verify` on pacakges for now and cut #156 for later.

### How has this been tested?
Added unit tests. Also converted the `// TEST: ...` comments to `@JVerifyTest` parameters, and added a new `verifyByDefault` parameter to correspond with the new verifier option.


test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTest.java
```diff
@@ -7,8 +7,45 @@
...
+ *      </li>
+ * </ol>
+ *
+ * Other parameters such as {@code verifyByDefault} can be set independently.
+ */
 @Testable
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ ElementType.TYPE })
 public @interface JVerifyTest {
+    /**
```
keyboardDrummer, line 36: Comment is incorrect and this was sometimes not used when it could have been

keyboardDrummer (commented):
Really nice! Left a non-blocking comment

keyboardDrummer (approved):

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTest.java
robin-aws: Obviously true about the comment, will fix. :)

The only cases I'm aware of where we had `// TEST: skip` were in examples, but since we opted to remove all the markup from examples so they are clean when included in the user guide, I was avoiding adding any markup. I did realize the two skipped examples are at least not yet included in the guide, but I tried adding `@JVerifyTest` to them just now it hit ClassNotFoundExceptions. So for now I'll leave things as is.

PR #162, opened 2025-05-21
Fix NullPointer exception when an interface method is not used

### LINKED ISSUE #161: Exception raised when analyzing a Java file with an unused Interface method
Run verifier on this
```
interface Unused {
    void unused(int x);
}
```
and you will have a null pointer exception:
```
 java.lang.NullPointerException at JavaToDafnyCompiler.java:552
```

olivier-aws:
### What was changed?
Fixes issue #161 

### How has this been tested?
Added test case in file Interfaces.java

Removes this exception when analyzing 14 files from the JML test files

### DIFF
--- a/verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
@@ -13,6 +13,7 @@ import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.TypeTag;
 import com.sun.tools.javac.tree.EndPosTable;
 import com.sun.tools.javac.tree.JCTree;
+import com.sun.tools.javac.tree.JCTree.JCStatement;
 import com.sun.tools.javac.tree.TreeMaker;
 
 import com.sun.tools.javac.tree.TreeInfo;
@@ -515,7 +516,8 @@ public class JavaToDafnyCompiler {
 
         if (annotationsByName.containsKey(Pure.class.getName())) {
             var header = new HeaderContainer();
-            var postHeader = methodCompiler.translateHeader(method.body.stats, header);
+            List<JCStatement> statements = method.getBody() == null ? List.of() : method.getBody().stats;
+            var postHeader = methodCompiler.translateHeader(statements, header);
             applyInvariants(method, header);
             if (postHeader.size() != 1) {
                 reportError(method, "pureMethodMultipleStatements");
@@ -549,7 +551,8 @@ public class JavaToDafnyCompiler {
                     body, null, null);
         } else {
             var header = new HeaderContainer();
-            var postHeader = methodCompiler.translateHeader(method.getBody().stats, header);
+            List<JCStatement> statements = method.getBody() == null ? List.of() : method.getBody().stats;
+            var postHeader = methodCompiler.translateHeader(statements, header);
             applyInvariants(method, header);
             methodCompiler.checkEmptyExpressions(method, header.invariants, "invariants", "method");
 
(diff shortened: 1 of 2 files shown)


verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -515,7 +516,8 @@ private boolean isNullable(JCTree.JCModifiers modifiers) {
 
         if (annotationsByName.containsKey(Pure.class.getName())) {
             var header = new HeaderContainer();
-            var postHeader = methodCompiler.translateHeader(method.body.stats, header);
+            List<JCStatement> statements = method.getBody() == null ? List.of() : method.getBody().stats;
```
keyboardDrummer, line 519: Shouldn't we return `null` for the entire `translateMethodDecl ` if the body is empty? Now it looks like we'll generate a Dafny declaration for the empty bodied method, with a trivial contract, even though no contract was specified.

olivier-aws: Actually I think we should not return null, but ensure to have a method with no body (rather than empty body). This way the method is havoc-ed during verification and can be used in the Java code. See latest commit that does this (with a new test).

keyboardDrummer: > Actually I think we should not return null, but ensure to have a method with no body (rather than empty body).

Then it would be a method with a trivial contract (requires true ensures true) that's trivially proven (no body). I still think it's not right to generate something with a trivial contract when on the Java side, no contract has been specified.

If you call an interface method from verified code, then a contract must be specified for that method. Generally that contract specification would look like this:

```java
interface I {
  void foo();

  @Contract
  class IContract implements I {
    void foo() {
      precondition(..);
      postcondition(..);
    }
  }
}
```

> This way the method is havoc-ed during verification

I'm not sure what you mean by that.

olivier-aws: I agree that this is what users should do ultimately. But they will not always, and I believe that having a trivial contract but not being able to prove anything on the output of the method (what I meant by havoc-ed) is a good enough solution in this case.

This is what you get if you use  `@Verify(false)` on a `@Pure` method with no pre/post-conditions, and I feel that interfaces without a Contract could be handled equivalently.

Maybe it's more efficient to discuss this offline.

keyboardDrummer: > I believe that having a trivial contract but not being able to prove anything on the output of the method (what I meant by havoc-ed) is a good enough solution in this case.

I'm fine with getting a trivial postcondition for free, but not a trivial precondition since then you can write a verified method that calls this interface method, and have it verify successfully even when the call is incorrect.

> This is what you get if you use @Verify(false) on a @Pure method with no pre/post-conditions, and I feel that interfaces without a Contract could be handled equivalently.

Do you mean like this?

```java
@Verify(false)
@Pure
int Callee(int x) {
  return 3 / x;
}

@Verify(true)
void Caller() {
  Callee(0);
}
```

What does the `@Pure` annotation matter?

IMO here the user, when they added `@Verify(false)`, explicitly decided to manually determine what the preconditions of `Callee` are, incorrectly thought there were none, and that's why the code in `Caller` is buggy.

keyboardDrummer: Related issue: https://github.com/aws/jverify/issues/132

fabiomadge:

PR #175, opened 2025-05-28
Use structured output and unmangle names in error messages

keyboardDrummer:
### What was changed?
- Use the flag `--json-output` instead of `--json-diagnostics`, so all output from Dafny is JSON.
- Unmangle names that occur in diagnostic message arguments

### How has this been tested?
- Updated tests that were previously expecting mangled names, so that they now expect unmangled names


verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -287,20 +306,31 @@ private static void parseDafnyJsonOutput(BufferedReader dafnyOutput, Verificatio
...
             } else if (line.startsWith("{")) {
                 try {
-                    outResults.diagnostics.add(objectMapper.readValue(line, DafnyDiagnostic.class));
+                    DafnyOutput output = objectMapper.readValue(line, DafnyOutput.class);
+                    if (output instanceof DafnyDiagnostic dafnyDiagnostic) {
+                        for (var index = 0; index < dafnyDiagnostic.arguments.length; index++) {
+                            dafnyDiagnostic.arguments[index] = mangler.safeUnmangleName(dafnyDiagnostic.arguments[index]);
+                        }
+                    }
+                    if (output instanceof StatusMessage statusMessage) {
```
robin-aws: Probably good to check that you don't get more than one of these just in case.

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
```diff
@@ -59,12 +60,29 @@ private static String typeMangling(com.sun.tools.javac.code.Type type) {
...
 
+    public String safeUnmangleName(String name) {
+        if (reverseSymbolStringMap.containsKey(name)) {
+            return reverseSymbolStringMap.get(name).name.toString();
+        }
+        return name;
+    }
+    
+    public Symbol unmangleName(String name) {
+        return reverseSymbolStringMap.get(name);
```
robin-aws: How about an exception right here if it's not in the map, instead of a NPE later on :)

verifier/src/main/java/com/aws/jverify/verifier/DafnyDiagnostic.java
```diff
@@ -89,9 +110,19 @@ public String getCode() {
         return null;
     }
 
-    @Override
     public String getMessage(Locale locale) {
-        return message;
+        String escapedFormatMsg = formatMessage.replace("'", "''");
+        return getSeverityMessage() + ": "+ MessageFormat.format(escapedFormatMsg, arguments);
```
robin-aws: Adding the severity which mostly duplicates `getKind()` seems like it's going to be noisy in some cases. What do we gain out of having DafnyDiagnostic implement Diagnostic right now?

Asking because I found mapping to a `JCDiagnostic` which supports subdiagnostics for relation locations seemed to work much better in my experiments.

robin-aws (commented):
LGTM, just a couple of safety suggestions

verifier/src/main/java/com/aws/jverify/verifier/DafnyDiagnostic.java
keyboardDrummer: What change are you suggesting? I'd prefer to get the error message from the kind as well. I think currently the severity/kind headers are inconsistent between JVerify and Dafny errors. It would be good to straighten that out, but I prefer to do that in a separate PR.

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
keyboardDrummer: And then throw an exception? Why would you expect the duplicate message?

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
keyboardDrummer: Removed this method

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
robin-aws: I don't, but I could imagine a bug in the way we call Dafny leading to duplicates by accident. And I'd rather get an exception rather than just overwriting to only keep the last one.

In fact, you should throw an exception if you get another class other than the two you're handling here too, right?

keyboardDrummer: > In fact, you should throw an exception if you get another class other than the two you're handling here too, right?

Good excuse to use Java's sealed and permits feature

robin-aws (approved):

PR #178, opened 2025-06-02
Partial support for generics

keyboardDrummer:
### What was changed?
- Support type parameters that do not have bounds

### How has this been tested?
- Added tests `PolymorphismWithoutBounds.java` and `PolymorphismWithBounds.java`

### DIFF
--- a/verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
@@ -956,8 +956,7 @@ public class JavaToDafnyCompiler {
                     }
                     return new NameSegment(origin, nameSegment.getName(), arguments);
                 }
-                reportError(expr, "notSupported", "Type application on " + type.getClass().getSimpleName());
-                return getHole(origin);  
+                throw new RuntimeException("All Dafny type references are NameSegments, since we do not use Dafny modules");
             }
             case null, default -> {
             }
--- a/verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
@@ -401,10 +401,9 @@ public class JavaToDafnyCompiler {
         var trees = JavacTrees.instance(context);
         var superTraits = interfaces.stream().
                 filter(type -> this.classWithExternalContract.contains(type.tsym) || trees.getTree(type.tsym) != null).
-                map((com.sun.tools.javac.code.Type type) ->
-                    new UserDefinedType(origin, new NameSegment(origin, nameMangler.mangleSymbolName(type.tsym), null))).
+                map((com.sun.tools.javac.code.Type type) -> toType(origin, type)).
                 collect(Collectors.<Type>toList());
-
+        
         var typeParameters = toTypeParameters(classDecl.typarams);
         if (createTrait) {
             if (classDecl.getModifiers().getAnnotations().stream().
@@ -416,6 +415,12 @@ public class JavaToDafnyCompiler {
         } else {
             return new ClassDecl(origin, name, null, typeParameters, members, superTraits, false);
         }
+    }
+    
+    
+    Type toType(IOrigin origin,  com.sun.tools.javac.code.Type javaType) {
+        return new UserDefinedType(origin, new NameSegment(origin, nameMangler.mangleSymbolName(javaType.tsym),
+                javaType.getTypeArguments().isEmpty() ? null : javaType.getTypeArguments().map(a -> toType(origin, a))));
     }
 
     private List<TypeParameter> toTypeParameters(List<JCTree.JCTypeParameter> typarams) {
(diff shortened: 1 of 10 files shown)


verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -1253,7 +1338,8 @@ public Type toType(com.sun.tools.javac.code.Type type, boolean isNullable, IOrig
             return TypeKind.VOID;
         } else if (type instanceof com.sun.tools.javac.code.Type.JCPrimitiveType primitiveType) {
             return primitiveType.getKind();
-        } else if (type instanceof com.sun.tools.javac.code.Type.ClassType classType
+        } 
+        else if (type instanceof com.sun.tools.javac.code.Type.ClassType classType
```
```suggestion
        } else if (type instanceof com.sun.tools.javac.code.Type.ClassType classType
```

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -382,24 +399,56 @@ private ClassLikeDecl translateClass(Stack<Tree> nestedTypes, JCTree.JCClassDecl
...
-            return new TraitDecl(origin, name, null, List.of(), members, superTraits, false);
+            return new TraitDecl(origin, name, null, typeParameters, members, superTraits, false);
         } else {
-            return new ClassDecl(origin, name, null, List.of(), members, superTraits, false);
+            return new ClassDecl(origin, name, null, typeParameters, members, superTraits, false);
         }
     }
     
+    
+    Type translateType(IOrigin origin, com.sun.tools.javac.code.Type javaType) {
```
robin-aws: Yikes, I think you really want to call this `translateTypeArgument` or something, rather than have an overload of `translateType` with the parameters reversed that behaves very differently. :)

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -895,6 +950,22 @@ public Expression toExpr(JCTree.JCExpression expr, IOrigin originOverride) {
...
+                    if (typeApply.getTypeArguments().isEmpty()) {
+                        // Occurs when the type arguments were inferred
+                        arguments = typeApply.type.getTypeArguments().stream().map(t -> translateType(origin, t)).toList();
+                    } else {
+                        arguments = typeApply.getTypeArguments().stream().map(this::translateType).toList();
+                    }
+                    var name = new NameSegment(origin, nameSegment.getName(), arguments);
+                    return name;
+                }
+                throw new RuntimeException("");
```
```suggestion
                reportError(expr, "notSupported", "Type application on " + type.getClass().getSimpleName());
                return getHole(origin);  
```

verifier/src/main/resources/builtin-contracts.java
```diff
@@ -13,12 +18,43 @@ class ShortContract {
...
+    
+}
+
+@Contract(Collection.class)
+interface CollectionContract<E> {
+
+}
+
+@Contract(List.class)
+abstract class ListContract<E> implements SequencedCollection<E> {
```
robin-aws, line 37: I assume you added these to support different test cases that then didn't work out? Just checking for my understanding, fine to leave in.

verifier/src/test/java/com/aws/jverify/verifier/tests/PolymorphismWithBounds.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public void doFeed() {
+        reads(this);
+        value.feed();
+    }
+}
+
+@Contract(Pet.class)
+class PetContract implements Pet {
+    public void feed() {
```
robin-aws, line 46: I assume the idea is to only add `throw new ContractException();` when the method doesn't return `void`, and not all the time as a safety measure?

robin-aws (changes requested):
Nice!

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: O yeah, actually the existing `translateType` also works, so I'll remove this one.

verifier/src/test/java/com/aws/jverify/verifier/tests/PolymorphismWithBounds.java
keyboardDrummer: That seems OK to me

verifier/src/main/resources/builtin-contracts.java
keyboardDrummer: Yes that's right

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: Clarified the exception message

robin-aws: Ah so it's an internal error we should never hit, thanks for the clarification (and in that case yes, should be an exception instead)

robin-aws (approved):

PR #181, opened 2025-06-03
Mixed external and internal contracts

keyboardDrummer:
### What was changed?
- Allow type member contracts to be specified both in the type and outside of the type, which is useful for abstract classes and interfaces with default implementations

### How has this been tested?
- Added tests `MixedExternalAndInternalContractVerification` and `MixedContractContractErrors`


docs/guide/src/SUMMARY.md
```diff
@@ -18,12 +18,13 @@
 - [Termination](termination.md)
 - [Integers](integers.md)
 
-# References
+# Working with objects
```
robin-aws, line 21: Good change!

docs/guide/src/interfaces.md
```diff
@@ -0,0 +1,63 @@
...
+    int foo(int x);
+
+    @Contract
+    class IContract implements I {
+
+        @Pure
+        public int foo(int x) {
+            precondition(x != 0);
+            postcondition((Integer r) -> r > 10);
+            throw new ContractException();
```
robin-aws, line 17: Minor style thing, but there's something about throwing an exception in a contract that just feels messy to me, since it should never actually be executed.

I've wondered if it would look nicer to have `return contractOnly();` instead, where we define `JVerify.contractOnly<T>()` ourselves and it throws the exception instead. And then that exception can be more internal like `VerificationMethodExecutedException`.

Might make users feel slightly more comfortable with the whole thing, and perhaps even avoid some false positives in other tooling warning about unhandled exceptions. Makes it a bit clearer that it's only necessary because Java forces you to return a value given the signature.

(Fine to address independently if you agree)

docs/guide/src/interfaces.md
```diff
@@ -0,0 +1,63 @@
...
+    
+    default int bar(int x) {
+        postcondition((Integer r) -> r > 10);
+        precondition(x >= 0);
+        return x + 10;
+    }
+}
+```
+
+Notice that for `bar`, the contract is provided by the interface `I`, while for `foo`, the contract is provided by the contract class `IContract`. Annotations on the type, such as `@Modifiable`, must be placed on the non-contract type, while annotations on the method, such as `@Pure`, must be placed on the method that defines the contract.
```
robin-aws: > Annotations on the type, such as `@Modifiable`, must be placed on the non-contract type

Doesn't seem to be true given the example below (and that's good because I don't see how we'd support library interfaces otherwise)

...figure it out after reading to the end. But might read better to state right up front that there are two cases, and this statement here only applies to the first case.

docs/guide/src/interfaces.md
```diff
@@ -0,0 +1,63 @@
...
+    public static <E> List<E> of(E first, E second) {
+        postcondition((List<E> r) -> r.elements == Sequence.of(first, second));
+        throw new ContractException();
+    }
+}
+```
+
+We only wanted to provide a contract for `size`, so we made ListContract an _abstract_ class. This way, the Java compiler is satisfied even though we did not implement all methods from the interface `List<T>`.
+
+Contrary to the `interface I` example from before, because `List<T>` is not defined in source, annotations such as `@Modifiable` are taken from the contract class.
```
robin-aws, line 65: Ah okay. Why not support either location when the interface is defined in source though?

docs/guide/src/interfaces.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Libraries without contracts
+
+You will often come across already libraries for which no contracts were defined. To call a method from such a library from verified code, you will have to define its contract. You can do this using the `@Contract` annotation, similar to how you would for interfaces or abstract classes. Example:
+
+```java
+import javax.sound.midi.Sequence;
+import java.util.List;
+
+// JVerify provides a contract for `List` out of the box,
```
robin-aws:
```suggestion
// JVerify provides a partial contract for `List` out of the box,
```

We can prioritize filling it out ASAP though.

verifier/src/main/java/com/aws/jverify/verifier/MethodOrLoopContract.java
```diff
@@ -1,11 +1,14 @@
...
 import com.aws.jverify.generated.*;
+import com.sun.tools.javac.tree.JCTree;
 
 import java.util.ArrayList;
 import java.util.List;
 
-public class HeaderContainer {
+public class MethodOrLoopContract {
+    JCTree treeOrigin;
+    boolean isPure;
```
robin-aws, line 14: What does "pure" mean in this context? Worth a comment.

verifier/src/main/java/com/aws/jverify/verifier/OverrideFinder.java
```diff
@@ -0,0 +1,81 @@
...
+import com.sun.tools.javac.code.Symbol;
+import com.sun.tools.javac.code.Type;
+import com.sun.tools.javac.code.Types;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class OverrideFinder
```
robin-aws, line 12: I suspect this logic already exists somewhere in the javac code, but not concerned with duplicating it a bit for now. Might be worth refactoring if this is ever a bottleneck and there is some memoized state in the AST we could rely on though.

verifier/src/main/resources/com/aws/jverify/messages.properties
```diff
@@ -7,8 +7,9 @@ compiler.err.notSupported={0} is not supported
...
 compiler.err.javaError=Java compilation had an error: {0}
-compiler.err.duplicateContract=a class may not be referenced by more than one @Contract annotation
-compiler.err.verifiedTypeWithExternalContract=class ''{0}'' may not be verified because it has an externally defined contract
-compiler.err.nestedContractClass=a class with @Contract may not be nested inside another type
+compiler.err.duplicateContract=class ''{0}'' must not be referenced by more than one @Contract annotation
 compiler.err.mutatingExpression=since {0} performs mutation, it may only be used where a statement is allowed
-compiler.err.constructorInInterfaceContract=a class that defines the contract for an interface may not have explicit constructors
\ No newline at end of file
+compiler.err.internalAndExternalContractForMethod=method ''{0}'' already has an internally defined contract
+compiler.err.concreteTypeWithExternalContract=class ''{0}'' must not have an externally defined contract because all its contracts can be defined internally.
```
robin-aws: Actually, could we support multiple non-overlapping contracts? I could see that being useful in the future if we provide a partial contract for something (e.g. List :) ) and users want to unblock themselves by adding their own partial contract for more declarations.

(Fine to add this support in a different change though)

verifier/src/main/resources/com/aws/jverify/messages.properties
```diff
@@ -7,8 +7,9 @@ compiler.err.notSupported={0} is not supported
...
+compiler.err.unusedContractMethod=method ''{0}'' is part of a @Contract class, but it does not override any methods for which to provide a contract
```
robin-aws: Might be clearer to just say everything in a contract class must have `@Override` and make this error message say "does not have @Override"

verifier/src/test/java/com/aws/jverify/verifier/tests/Interfaces.java
```diff
@@ -5,29 +5,37 @@
...
-        reads(this);
-        throw new ContractException();
-    }
-
-    public int m() {
-        modifies(this);
-        postcondition((Integer r) -> r > 2);
-        throw new ContractException();
+    @Contract
+    class IContract implements I {
```
robin-aws, line 24: Did you nest this just to show it's possible?

docs/guide/src/interfaces.md
robin-aws, line 1: I like the UX improvement of not having to specify the target class. Worth mentioning here that you can still do that though, since that's the only way you can attach a contract to a final or private type (since you then you can't extend/implement the target). Worth making sure we have a test case for that too.

robin-aws (changes requested):
Great change, just mostly wanting to tweak documentation

docs/guide/src/interfaces.md
keyboardDrummer: Interesting idea. I don't have an opinion yet but happy to make that change if you think it's good. I think it would be better in a separate PR though.

verifier/src/main/resources/com/aws/jverify/messages.properties
keyboardDrummer: > I could see that being useful in the future if we provide a partial contract for something (e.g. List :) ) and users want to unblock themselves by adding their own partial contract for more declarations.

You're right we should support that. I'll create a GH issue

verifier/src/main/resources/com/aws/jverify/messages.properties
keyboardDrummer: You can provide contracts for static methods too, and you can't @Override those. I've updated the message not to use the word override.

verifier/src/test/java/com/aws/jverify/verifier/tests/Interfaces.java
keyboardDrummer: I think writing it nested like this should be the default for interfaces/abstract classes. I'd even consider making anything else an error, except when you're providing a library contract of course. I think we could also consider two different annotations, one for library and one for source contracts.

docs/guide/src/interfaces.md
keyboardDrummer: > Ah okay. Why not support either location when the interface is defined in source though?

I think that would make the code harder to read, because then you have to look in two places to figure out what the annotations are

verifier/src/main/java/com/aws/jverify/verifier/OverrideFinder.java
keyboardDrummer: > I suspect this logic already exists somewhere in the javac code

Why?

robin-aws: Seems like the kind of logic that's likely used for optimizations in javac somewhere.

robin-aws (approved):

PR #184, opened 2025-06-05
feat: basic string support

### LINKED ISSUE #62: Partial support for String
For String, encode it like Y does:

```
type char16 = c: char | '\U{0000}' <= c <= '\U{ffff}'
type jstring = seq<char16>
```
and support:

string literals
charAt
isEmpty

- [ ] Sketch how other String methods will be implemented for this encoding

alex-chew:
Closes #62.

### What was changed?

- Adds translation for String literals. Dafny string literals are used on a best-effort basis, to make debugging easier, falling back to sequences of raw numeric values.
- Adds translations for a few important operations/methods - concatenation, `charAt`, `equals`, `isEmpty`, `length`, `substring(int)`, and `substring(int, int)`.

### How has this been tested?

Added unit tests for all of the above.


verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -883,8 +887,43 @@ public Expression toExpr(JCTree.JCExpression expr, IOrigin originOverride) {
...
+                var methodSymbol = TreeInfo.symbol(invocation.getMethodSelect());
                 var argBindings = invocation.getArguments().stream().map(a -> new ActualBinding(null, toExpr(a), false)).toList();
+
+                if (methodSymbol.owner instanceof Symbol.ClassSymbol ownerClass
+                        && ownerClass.fullname.contentEquals(String.class.getName())) {
+                    final Expression receiver;
+                    if (invocation.getMethodSelect() instanceof JCTree.JCFieldAccess fieldAccess) {
+                        receiver = toExpr(fieldAccess.selected);
+                    } else {
+                        reportError(invocation, "notSupported", "method reference call");
```
robin-aws, line 1068: What case is "method reference call" describing? Just static methods?

verifier/src/main/resources/additional.dfy
```diff
@@ -7,7 +7,9 @@ type int32 = x: int | -0x8000_0000 <= x && x <= 0x7fff_ffff
 type nat63 = x: int64 | x >= 0
 type int64 = x: int | -0x8000_0000_0000_0000 <= x <= 0x7fff_ffff_ffff_ffff
 
-type char16 = c: char | '\U{0000}' <= c <= '\U{ffff}'
+// Base type is int and not char, because Java's char allows surrogates and Dafny's char does not
```
robin-aws, line 10: Ah this is sensible. Bit of a bummer since it'll be a slight hit to debuggability of the translated Dafny, but not a big deal. We could always switch to `--unicode-char:false` if that ends up mattering enough.

verifier/src/test/java/com/aws/jverify/verifier/tests/Strings.java
```diff
@@ -0,0 +1,63 @@
...
+
+import static com.aws.jverify.JVerify.*;
+
+@SuppressWarnings({
+        "ConstantValue",
+        "SizeReplaceableByIsEmpty",
+        "OnlyOneElementUsed",
+        "StringOperationCanBeSimplified"
+})
+@JVerifyTest(dafnyVerified = 9, dafnyErrors = 0)
```
robin-aws: I would add at least a couple of cases of failing verification, both as a sanity check that verification is happening properly at all, but also to check that source locations of diagnostics are sensible.

verifier/src/test/java/com/aws/jverify/verifier/tests/Strings.java
```diff
@@ -0,0 +1,63 @@
...
+
+    static void stringLiteralControlChars() {
+        check("\u0000".charAt(0) == '\0');
+
+        var s = "\u0001\u0002\u0003";
+        check(s.length() >= 3);
+        check(s.charAt(1) == 2);
+    }
+
+    static void stringLiteralMispairedSurrogates() {
```
robin-aws, line 69: Probably worth a few tests of literals with non-BMP characters too, since Java source supports them.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -1524,6 +1567,68 @@ public static Symbol.MethodSymbol getJVerifyMethod(JCTree.JCMethodInvocation inv
...
+        );
+        var initLambda = new LambdaExpr(
+                origin,
+                List.of(new BoundVar(origin, new Name(origin, "i"), null, false)),
+                indexRange,
+                initBody,
+                null,
+                new Specification<>(List.of(), null)
+        );
+        return new SeqConstructionExpr(origin, null, lengthExpr, initLambda);
```
robin-aws: I'd be tempted to define something like this in `additional.dfy` instead:

```dafny
function JString(s: string): jstring {
  seq(|s|, i requires 0 <= i && i < |s| => s[i] as char16)
}
```

Just to save noise in the translated Dafny.

robin-aws (changes requested):
Nice!

verifier/src/main/resources/additional.dfy
alex-chew: Yeah, I'm hoping that it won't impact the debuggability much, but in case we do decide that `--unicode-char false` is better, it's a pretty easy change to make.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
alex-chew: It's referring to method references as described in https://docs.oracle.com/javase/specs/jls/se21/html/jls-15.html#jls-15.13, e.g. `String::length`.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -1524,6 +1567,44 @@ public static Symbol.MethodSymbol getJVerifyMethod(JCTree.JCMethodInvocation inv
...
+
+        var translatedChars = new StringBuilder();
+        for (var charValue : stringValue.toCharArray()) {
+            if (ESCAPED_CHARS.containsKey(charValue)) {
+                translatedChars.append(ESCAPED_CHARS.get(charValue));
+            } else if (charValue >= 0x20 && charValue <= 0x7e) {
+                translatedChars.append(charValue);
+            } else {
+                // Fallback to sequence of numeric values
+                var charExprs = stringValue.chars().boxed()
```
keyboardDrummer, line 1767: When does this case occur? Can you give an example of a string literal that triggers this?

~Don't you get into trouble if the Dafny types for Strings can be different? I'd expect to get a type error if I compare characters from a `seq<char>` and a `seq<Integer>`.~

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -883,8 +887,43 @@ public Expression toExpr(JCTree.JCExpression expr, IOrigin originOverride) {
...
+                        && ownerClass.fullname.contentEquals(String.class.getName())) {
+                    final Expression receiver;
+                    if (invocation.getMethodSelect() instanceof JCTree.JCFieldAccess fieldAccess) {
+                        receiver = toExpr(fieldAccess.selected);
+                    } else {
+                        reportError(invocation, "notSupported", "method reference call");
+                        receiver = getHole(origin);
+                    }
+
+                    return switch (methodSymbol.name.toString()) {
```
keyboardDrummer, line 1072: Would it also have been possible to define a contract for the String type, like:

```java
@Contract(String.class)
@Immutable
class StringContract {
  Sequence<Integer> data;

  @Pure
  public Integer charAt(@Nat int index) {
    precondition(index < data.size());
    return data.get(index);
  }
}
```

? Any pros/cons to doing that? If you can think of any maybe you can leave a comment here to indicate the choice.

I think for char and string literals, we have no other option than to have special casing in the translation code, as you added in this PR.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
robin-aws: Ah okay. I'm not 100% confident that's the only other legal type of node that could occur here, but I can live with that.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
robin-aws: Agreed this is worth writing down. I think the major upside to the implemented approach is that it's much simpler to use Strings in pure contexts. Otherwise even a string literal becomes an object allocation.

robin-aws (commented):
LGTM but @keyboardDrummer 's questions are worth answering too before merging

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
alex-chew: I'll investigate further and open a follow-up PR if more cases are needed here.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
alex-chew: I hadn't considered the `@Contract` approach, but it's certainly easier to read the "translation", and I'm guessing it's be easier to add new APIs and maintain them since they're less tightly coupled to the AST. 

> Otherwise even a string literal becomes an object allocation.

@robin-aws I'm not sure what you mean by this. If implementing the `@Contract` approach, I'd want to translate a String object to a `datatype` wrapper around a `seq<char16>`, and not to a Dafny class - or does that not work with how we want to treat referential equality?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
alex-chew: This case occurs when the string literal contains anything other than printable ASCII characters and the few Dafny escape sequences - null, newline, tab, etc. (See the doc comment of the surrounding method.)

There are many such cases in the added tests, for example the `"\u0001\u0002\u0003"` in `stringLiteralControlChars`, and the emojis in other test cases.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
robin-aws: Yes I was assuming using `@Contract` meant still treating `java.lang.String` as a reference type in the Dafny translation, and therefore that a string literal would have to translate to `new java_lang_String()` in the Dafny, and therefore you couldn't use that in a pure context at all.

We've discussed being able to map Java classes tagged with something like `@ValueType` to Dafny datatypes, but for the moment we're only planning to do that for records.

Another tricky thing is how to translate `==` on Java reference types. With either approach, mapping a Java String to a Dafny sequence (or a datatype wrapping one) isn't actually sound: `check("a" == "a")` will verify when it shouldn't.

I'm inclined to merge this since it's significant progress and sound so far anyway. The worse case is we have to completely revisit it when we get to `==` on object types, but we'll cross that bridge when we come to it. @alex-chew it might still be worth adding some comments in a follow up PR though.

robin-aws (approved):

PR #187, opened 2025-06-06
Classes extending classes

keyboardDrummer:
### What was changed?
- Fix a bug that caused the source locations of mangled names to become incorrect
- Changed the order in which types are traversed, so they are traversed topologically in the type hierarchy graph. This allows using information collected while compiling base types.
- Support classes extending classes. Does not yet support overriding virtual methods or default methods in interfaces

### How has this been tested?
- Added `ClassesExtendingClassesVerification.java` and `ClassesExtendingClassesErrors.java`


verifier/src/test/java/com/aws/jverify/verifier/tests/Interfaces.java
```diff
@@ -114,6 +114,6 @@ public int m() {
         postcondition((Integer r) -> r > 2);
         c.x = 3;
 //      ^^^ Error: assignment might update an object not in the enclosing context's modifies clause
-        return 0;
+        return 3;
```
robin-aws, line 117: Yikes, was this not caught before the Dafny verification fix?

verifier/src/test/java/com/aws/jverify/verifier/tests/ClassesExtendingClassesVerification.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Pure
+    @Override
+    public int computeX(int input) {
+        return 3;
+    }
+}
+
+@Contract(Extendee.class)
+class ExtendeeContract extends Extendee {
```
robin-aws: You said before you preferred the contracts to be nested inside the contractees? :)

verifier/src/test/java/com/aws/jverify/verifier/tests/Arrays.java
robin-aws, line 1: Are these reorderings actually necessary?

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -216,9 +216,11 @@ private static void checkDafnyVersion(VerifierOptions verifierOptions) {
...
                         throw new RuntimeException("dafny --version failed:\n" + output);
                     }
-                    if (!output.equals(expectedVersion)) {
-                        throw new IllegalStateException("Wrong Dafny version: expected " + expectedVersion + " but found " + output);
-                    }
+                    // Turned off while we're using a Dafny submodule
+                    // Alternatively, we can check whether the submodule version matches the output
+//                    if (!output.equals(expectedVersion)) {
+//                        throw new IllegalStateException("Wrong Dafny version: expected " + expectedVersion + " but found " + output);
+//                    }
```
robin-aws: I'm not okay with removing this unless we also temporarily disable the support for using any Dafny other than the submodule.

examples/src/test/java/com/aws/jverify/examples/UserProfile.java
```diff
@@ -16,6 +16,8 @@ public UserProfile(AccountType accountType) {
         this.accountType = accountType;
         if (AccountType.Premium == accountType) {
             this.premiumFeatures = new PremiumFeatures();
+        } else {
+            this.premiumFeatures = null;
```
robin-aws, line 20: Why did you have to add this?

verifier/src/main/java/com/aws/jverify/verifier/MethodCompiler.java
```diff
@@ -556,16 +560,17 @@ public AssignmentRhs toAssignmentRhs(JCTree.JCExpression expr, IOrigin originOve
...
-                var ty = new UserDefinedType(origin, new ExprDotName(origin, baseType, ctorName, null));
+                var baseType = (NameSegment)compiler.expressionCompiler.toExpr(newClass.clazz);
+                var classBaseType = new NameSegment(baseType.getOrigin(), "_Class_" + baseType.getName(), baseType.getOptTypeArguments());
+                var ty = new UserDefinedType(origin, new ExprDotName(origin, classBaseType, ctorName, null));
 
                 return new AllocateClass(origin, null, ty, new ActualBindings(argBindings));
             }
             case JCTree.JCNewArray newArray -> {
-                var arrayDimensions = newArray.getDimensions().stream().map(expressionCompiler::toExpr).toList();
+                var arrayDimensions = newArray.getDimensions().stream().map(d -> compiler.expressionCompiler.toExpr(d)).toList();
```
robin-aws:
```suggestion
                var arrayDimensions = newArray.getDimensions().stream().map(compiler.expressionCompiler::toExpr).toList();
```

Still works!

verifier/src/main/java/com/aws/jverify/verifier/NameMangler.java
```diff
@@ -130,4 +126,8 @@ private String mangleMethodName(Symbol s) {
         }
         return baseName;
     }
+
+    public static String getConstructorName(Symbol.ClassSymbol enclosingClass) {
+        return "_ctor" + enclosingClass.name.toString();
```
robin-aws:
```suggestion
        return "_ctor_" + enclosingClass.name.toString();
```
?

verifier/src/main/java/com/aws/jverify/verifier/MethodCompiler.java
```diff
@@ -354,11 +347,22 @@ private List<Statement> translateJVerifyMethodInvocation(JCTree.JCMethodInvocati
     private List<Statement> translateVanillaJavaMethodInvocation(JCTree.JCMethodInvocation invocation) {
         var origin = compiler.toOrigin(invocation);
         if (invocation.getMethodSelect() instanceof JCTree.JCIdent ident && ident.name.contentEquals("super")) {
```
robin-aws:
```suggestion
        if (invocation.getMethodSelect() instanceof JCTree.JCIdent ident && ident.name == ident.name.table.names._super) {
```

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -488,27 +529,106 @@ private ClassLikeDecl translateClass(Stack<Tree> nestedTypes, JCTree.JCClassDecl
...
+                    }
+                }
+                case Constructor constructor -> {
+                    classNeeded = true;
+                    Method initMethod = constructorToInitMethod(constructor);
+                    if (initMethod != null) {
+                        traitMembers.add(initMethod);
+                    }
+
+                    var attr = new Attributes(constructor.getOrigin(), "axiom", List.of(), null);
```
robin-aws: Why `{:axiom}` instead of an empty body?

verifier/src/test/java/com/aws/jverify/verifier/tests/ClassesExtendingClassesVerification.java
robin-aws, line 1: Worth some testing of `@Invariant` on these as well, or making that explicitly not supported yet if it's not ready.

verifier/src/main/resources/builtin-contracts.java
```diff
@@ -8,7 +8,8 @@
 import java.util.SequencedCollection;
 
 @Contract(Object.class)
-class ObjectContract {    
+class ObjectContract {
+    public ObjectContract() {}
```
robin-aws, line 12: Why do we need constructors in contracts now?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -488,27 +529,106 @@ private ClassLikeDecl translateClass(Stack<Tree> nestedTypes, JCTree.JCClassDecl
...
+        frameExpressions.add(new FrameExpression(constructor.getOrigin(), new ThisExpr(constructor.getOrigin()), null));
+        return new Method(constructor.getOrigin(), nameNode, constructor.getAttributes(),
+                constructor.getIsGhost(), constructor.getSignatureEllipsis(), constructor.getTypeArgs(), constructor.getIns(),
+                constructor.getReq(), constructor.getEns(), constructor.getReads(), constructor.getDecreases(),
+                modClause,
+                false, List.of(), body, false);
+    }
+
+    public static String getInitMethodName(String constructorName) {
+        return "_init" + constructorName;
```
```suggestion
        return "_init_" + constructorName;
```

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -488,27 +529,106 @@ private ClassLikeDecl translateClass(Stack<Tree> nestedTypes, JCTree.JCClassDecl
         var definingSymbol = getCurrentTypeSymbol(classDecl);
         
         Stream<com.sun.tools.javac.code.Type> baseTypes = definingSymbol.getInterfaces().stream();
-// 'extends' not yet supported when extending a class
-//        if (definingSymbol.getSuperclass() != null)
-//        {
-//            baseTypes = Stream.concat(Stream.of(definingSymbol.getSuperclass()), baseTypes);
-//        }
+        if (definingSymbol.getSuperclass() != null)
+        {
```
```suggestion
        if (definingSymbol.getSuperclass() != null) {
```

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -37,16 +44,24 @@
...
-    Stack<IOrigin> contextOrigins = new Stack<>();
-    DiagnosticCollector<JavaFileObject> diagnostics = new DiagnosticCollector<>();
+
+    public final Set<Symbol.MethodSymbol> symbolsWithAContract = new HashSet<>();
+    private final Stack<IOrigin> contextOrigins = new Stack<>();
+    public final DiagnosticCollector<JavaFileObject> diagnostics = new DiagnosticCollector<>();
     public final NameMangler nameMangler = new NameMangler();
+    private JCDiagnostic.Factory diagnosticFactory;
+    
+    private final Graph<Symbol.ClassSymbol, DefaultEdge> typeHierarchy = new DefaultDirectedGraph<>(DefaultEdge.class);
```
robin-aws, line 57: Can you add a comment about which direction the edges are pointed in?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -102,25 +136,38 @@ public NameMangler getNameMangler() {
...
-            findExternalContracts((JCTree.JCCompilationUnit) compilationUnit);
+        return parsed;
+    }
+
+    private void compileSymbolsTopologically() {
+        var iterator = new TopologicalOrderIterator<>(this.typeHierarchy);
+        var items = new ArrayList<Symbol.ClassSymbol>();
+        while (iterator.hasNext()) {
+            items.add(iterator.next());
         }
```
```suggestion
        iterator.forEachRemaining(items::add);
```

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -102,25 +136,38 @@ public NameMangler getNameMangler() {
...
+                parsed = null;
             }
         }
-        for (var compilationUnit : parsed) {
-            findExternalContracts((JCTree.JCCompilationUnit) compilationUnit);
+        return parsed;
+    }
+
+    private void compileSymbolsTopologically() {
+        var iterator = new TopologicalOrderIterator<>(this.typeHierarchy);
```
robin-aws, line 150: Also worth a comment around here about which direction this ordering ends up being (i.e. up or down the hierarchy)

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -102,25 +136,38 @@ public NameMangler getNameMangler() {
...
-            var fileStart = translateFile((JCTree.JCCompilationUnit) compilationUnit);
-            filesStarts.add(fileStart);
+        for(var currentTypeSymbol : items.reversed()) {
+            var relatedDeclarations = declarationsForSymbolContract.get(currentTypeSymbol);
+            if (relatedDeclarations == null) {
+                continue;
+            }
+            for(var relatedDeclaration : relatedDeclarations) {
+                Enter enter = Enter.instance(context);
+                Env<AttrContext> env = enter.getEnv(relatedDeclaration.sym);
```
robin-aws, line 160: FYI this might be cleaner after my change to reuse the additional `LAMBDA` javac phase, because it will change how we apply that stack such that we end up with a reference to the set of envs that comes out the bottom of the pipeline anyway.

We're going to conflict rather badly unfortunately :) but I'm going to also split off `parseAndResolveJava` and try to keep all the changes inside there.

robin-aws (changes requested):

verifier/src/test/java/com/aws/jverify/verifier/tests/Interfaces.java
keyboardDrummer: I think it relates, but this was caught after I fixed some source location mistakes in this PR as well.

verifier/src/test/java/com/aws/jverify/verifier/tests/Arrays.java
keyboardDrummer: No, but a top-down order is easier to read, right?

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
keyboardDrummer: Brought it back, but only when using a custom Dafny version

examples/src/test/java/com/aws/jverify/examples/UserProfile.java
keyboardDrummer: Otherwise the invariant was not satisfied. I'm not sure why Dafny doesn't use `null` as a default value. I've added a reference to this in https://github.com/aws/jverify/issues/170 to address this

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: The body is already empty. The axiom is to prevent a warning. Actually we are using `--allow-axioms`, so I'm removing this.

verifier/src/main/resources/builtin-contracts.java
keyboardDrummer: This says that `Object` has a public parameterless constructor. It's needed because we have tests that do `new Object()`

verifier/src/test/java/com/aws/jverify/verifier/tests/Arrays.java
robin-aws: Sure! Just sanity checking that the topological traversal you're using means the old version was also valid.

keyboardDrummer: I see :)

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
robin-aws: The body is already not present rather than an empty `{}`. I'm asking why an empty body wouldn't work.

(Realizing this isn't in scope for the PR so not blocking)

verifier/src/main/resources/builtin-contracts.java
robin-aws: Ah it's a contract for the contractees constructor, not a constructor of the contract. :)

verifier/src/test/java/com/aws/jverify/verifier/tests/Operators.java
```diff
@@ -186,7 +186,9 @@ void castRefInvalid() {
         DummyInterface dc2 = new DummyClass2();
 
         //noinspection RedundantCast,NewObjectEquality,DataFlowIssue
-        check(dc1 == (DummyClass) dc2);
+        var casted = (DummyClass) dc2;
 //                   ^^^^^^^^^^^^^^^^ Error: value of expression (of type 'DummyInterface') is not known to be an instance of type 'DummyClass'
+        check(dc1 == casted);
+//      ^^^^^^^^^^^^^^^^^^^^ Error: assertion might not hold
```
robin-aws, line 192: FYI the annotation parser/checker does support multiple annotations from a single source line now, if that's why you had to refactor this (I had to add that in a prior change when I couldn't avoid it)

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
robin-aws: PERFECT, thanks!!

robin-aws (approved):

PR #189, opened 2025-06-09
Adding JML tests translated into JVerify

olivier-aws:
### What was changed?
Adding some integration tests that were translated from JML ones. All should be passing for now.

### How has this been tested?


verifier/src/test/java/com/aws/jverify/verifier/jml_tests/SC.java
```diff
@@ -0,0 +1,35 @@
...
+		    return 1;
+		}
+		return 2;
+
+	}
+
+	//@ ensures a < 0 ==> \result == 2;
+	//@ ensures b < 0 ==> \result == 2;
+	//@ ensures a + b < 0 ==> \result == 2;
+	//@ ensures a >= 0 && b >= 0 && a+b >= 10 ==> \result == 1;
```
robin-aws, line 28: Was it an intentional choice to merge multiple `ensures` into a single `postcondition`? I often find it's easier to read relatively independent specifications if they are declared separately. It's at least probably better testing coverage to not ALWAYS have at most one `postcondition`

robin-aws (commented):
Can you include a README to describe the process for extracting these?

verifier/src/test/java/com/aws/jverify/verifier/jml_tests/SC.java
olivier-aws: I could not find a way to have multiple `postcondition` statements that mention the result of the method.
For example:
```
    public int foo() {
        postcondition((Integer r) -> r > 0);
        postcondition((Integer r2) -> r2 < 100);
        return RandomGenerator.getDefault().nextInt();
    }
```
and
```
    public int foo() {
        postcondition((Integer r) -> r > 0);
        postcondition((Integer r) -> r < 100);
        return RandomGenerator.getDefault().nextInt();
    }
```
both fail with error
```
 error: ensures clauses may introduce only one return variable name
```

olivier-aws: But I agree with you that this should be supported.

keyboardDrummer:
What's the reason for adding these tests? Is it to increase our testing coverage? They also increase our CI build time. If the goal is purely to increase coverage, then it would be better to determine which tests increase coverage and to add only those, or maybe even add them to existing tests.

olivier-aws:
> What's the reason for adding these tests? Is it to increase our testing coverage? They also increase our CI build time. If the goal is purely to increase coverage, then it would be better to determine which tests increase coverage and to add only those, or maybe even add them to existing tests.

I see two reasons:
1. increase coverage - we currently don't have that many tests and I'm afraid that we may be missing important issues if we don't quickly extend our coverage. For example, there are only 16 usage of `precondition` in our tests/ directory, these tests add many usage of it, in particular in cases that were not tested before (e.g. `precondition(aa.tricky != null)`)
2. reveal current limitations - doing the translation from JML to JVerify showed issues in JVerify, e.g. the fact that we cannot use two `postcondition` statements or that we cannot state `modifies(this.x)`.

I'm OK with not adding all these tests if you fear the CI build time will be an issue (e.g. the A[0-9]*.java tests are not super helpful).

keyboardDrummer:
Added build time seems to be less than 30 seconds, which I think is fine

verifier/src/test/java/com/aws/jverify/verifier/jml_tests/SC.java
olivier-aws: Update the test now that JVerify support this feature

olivier-aws:
> Can you include a README to describe the process for extracting these?

Added a README with a tutorial on how to convert from JML to JVerify.

fabiomadge:

PR #193, opened 2025-06-10
Allow multiple calls to postcondition

### LINKED ISSUE #191: Support multiple postcondition statements that constrain the return value in a method
Currently, there is not way to have two `postcondition` statements that both add constraints to the return value:

```
public static int foo() {
    postcondition((Integer r) -> r>0);
    postcondition((Integer r) -> r<0);
    ...
}
```
raises an error:
```error: ensures clauses may introduce only one return variable name```

### PR DESCRIPTION
**Pull request by olivier-aws**
### What was changed?
Update in JVerify compilation to allow multiple calls to postcondition with lambda, if the lambda parameters have the same name

### How has this been tested?
Added test case
Modified the error test case

### DIFF
--- a/verifier/src/main/java/com/aws/jverify/verifier/MethodCompiler.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/MethodCompiler.java
@@ -474,15 +474,27 @@ public class MethodCompiler {
                 }
                 case "postcondition" -> {
                     if (invocation.args.size() != 1) {
-                        throw new JavaViolationException("An postcondition call may have only one argument");
+                        throw new JavaViolationException("A postcondition call may have only one argument");
                     }
                     var first = invocation.getArguments().getFirst();
                     if (first instanceof JCTree.JCLambda lambda) {
                         if (lambda.getParameters().size() != 1) {
-                            throw new JavaViolationException("An ensures call lambda may take only one argument");
+                            throw new JavaViolationException("A postcondition ccall lambda may take only one argument");
                         }
                         var parameter = lambda.getParameters().getFirst();
-                        header.returnNames.add(new Name(compiler.toOrigin(lambda), parameter.getName().toString()));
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: verifier/src/main/java/com/aws/jverify/verifier/MethodCompiler.java [hunk 1]
```diff
@@ -474,15 +474,27 @@ public List<JCTree.JCStatement> translateHeader(List<JCTree.JCStatement> stateme
...
                 case "postcondition" -> {
                     if (invocation.args.size() != 1) {
-                        throw new JavaViolationException("An postcondition call may have only one argument");
+                        throw new JavaViolationException("A postcondition call may have only one argument");
                     }
                     var first = invocation.getArguments().getFirst();
                     if (first instanceof JCTree.JCLambda lambda) {
                         if (lambda.getParameters().size() != 1) {
-                            throw new JavaViolationException("An ensures call lambda may take only one argument");
+                            throw new JavaViolationException("A postcondition ccall lambda may take only one argument");
```
olivier-aws:
```suggestion
                            throw new JavaViolationException("A postcondition call lambda may take only one argument");
```

verifier/src/main/java/com/aws/jverify/verifier/MethodCompiler.java
```diff
@@ -474,15 +474,27 @@ public List<JCTree.JCStatement> translateHeader(List<JCTree.JCStatement> stateme
...
                         }
                         var parameter = lambda.getParameters().getFirst();
-                        header.returnNames.add(new Name(compiler.toOrigin(lambda), parameter.getName().toString()));
+                        var paramName = parameter.getName().toString();
+                        var origin = compiler.toOrigin(lambda);
```
olivier-aws, line 485:
```suggestion
```

keyboardDrummer (approved):

Issue #179, opened 2025-06-02
Support reference semantics of Java reference types translated to Dafny value types (boxed primitives, String, etc.)


keyboardDrummer:
- Also consider how `instanceof T` can be translated for these types `T`

alex-chew:
I'm expanding the scope of this task to include other Java reference types that are translated as Dafny value types, since I think it's likely we'd use the same approach for all of them. That would include String, record types, and (soon) sealed interfaces.

If you feel it'd be better to split them off, I'd be happy to do that instead.

keyboardDrummer:
I'm closing this issue because I think "reference semantics" is too big in scope. I've created one more specific issues but there are probably others as well: https://github.com/aws/jverify/issues/304

PR #200, opened 2025-06-14
feat: support records

### LINKED ISSUE #53: Support records
They are translated to Dafny datatypes with a single constructor.

Notes:
- Include check that hashcode and equals are not overridden
- Try to use the Dafny resolution errors, when the datatype extends a reference trait

alex-chew:
Closes #53.

### What was changed?

This PR adds support for translating Java records to Dafny datatypes. Some Java record features that are difficult to translate soundly are not supported: explicit constructors, explicit accessors, and overridden `equals` and `hashCode` implementations.

It also forbids using Java's equality operators (`==` and `!=`) when both operands' types could be String, a record, or a boxed primitive, *and* neither operand is a `null` literal. Since these types are translated to Dafny value types, the semantic difference between Java's `==` and Dafny's `==` could lead to unsoundness. (See the code comments and user guide changes for details.) It also forbids using the `@Nullable` annotation on these types (String, records, and boxed primitives) since we don't translate those properly anyways.

It also includes some tangential fixes/features:

- `postcondition<T>(Function<T, Boolean>)` can now be used in constructors (of class `T`) to express conditions of the constructed instance.
- The AST type generator no longer silently omits fields whose types have multiple levels of type parameters (e.g. `List<CasePattern<BoundVar>>`).
- The Dafny submodule includes a fix for printing unresolved datatype values.
- The `examples` tests no longer ignore the source markup (and I removed some outdated source markup).

### How has this been tested?

Added unit tests for verification and for rejecting unsupported features.


keyboardDrummer:
Can we add support for instantiating records inside nested expressions? This is a common operation in customer B's codebase, and something Dafny allows. Fine to add it in a follow-up PR

docs/guide/src/supported_java_features.md
```diff
@@ -1,8 +1,13 @@
 JVerify intends to support the full Java language, but is currently being developed and only supports a subset of Java constructs. The supported and not supported features are listed below. If you encounter unsupported features that are not in the unsupported list, please create a GitHub issue.
 
-# Supported Java Features
-- All literals except strings
+# Supported Java features
+- All literals
 - All operators
+  - Equality operators (`==` and `!=`) are not allowed when both operands' types could be String, a record, or a boxed primitive, *and* neither operand is a `null` literal.
```
robin-aws:
```suggestion
  - Equality operators (`==` and `!=`) are not allowed when both operands' types could be String, a record, or a boxed primitive, unless one of the operands is a `null` literal.
```

Reads clearer IMO.

verifier/src/test/java/com/aws/jverify/verifier/tests/PolymorphismWithoutBounds.java
```diff
@@ -5,24 +5,28 @@
...
 
+/*
+ * Note: the "==" comparisons in this file are commented out
+ * because we can't currently soundly translate "==" operands of type Object.
+ */
 @JVerifyTest(dafnyVerified = 5, dafnyErrors = 0)
 public class PolymorphismWithoutBounds {
     public static void root() {
         var obj = new Object();
         GenericContainer<Object> container = new GenericContainer<Object>(obj);
```
robin-aws, line 13: Could you declare an interface with `@Modifiable` and use that in place of `Object`? AFAICT that would allow the use of `==` here.

verifier/src/test/java/com/aws/jverify/verifier/tests/Records.java
```diff
@@ -0,0 +1,136 @@
...
+    public Vec3 allMax() {
+        return new Vec3(this.max(), this.max(), this.max());
+    }
+}
+
+/**
+ * Records are translated to Dafny datatypes,
+ * which don't automatically yield both nullable and non-nullable types (as Dafny classes do).
+ * This wrapper class works around JVerify's current lack of automatic handling
+ * for values that are of reference types in Java but of value types in Dafny.
```
robin-aws, line 135: Do we have this tracked as an issue? Worth linking here.

verifier/src/test/java/com/aws/jverify/verifier/tests/RecordsErrors.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        // Other explicit constructors are also forbidden
+        ExplicitConstructors() {
+//      ^ error: explicit record constructor is not supported
+            this(1, true);
+        }
+    }
+
+    record ExplicitAccessors(int i, boolean b) {
```
robin-aws, line 24: Why do we not support explicit accessors? The naming conflict should be handled by the symbol munging as we translate to Dafny AFAICT.

verifier/src/test/java/com/aws/jverify/verifier/tests/RecordsErrors.java
```diff
@@ -0,0 +1,58 @@
...
+        public int hashCode() {
+//                 ^ error: overridden hashCode method in record is not supported
+            return -1;
+        }
+
+        // A method may share its name with a forbidden-to-override method,
+        // as long as it's not override-equivalent to any forbidden-to-override method.
+        public int hashCode(int i) {
+            return 4;  // chosen by fair dice roll.
+                       // guaranteed to be random.
```
robin-aws, line 56: LOL

The best part is this is sometimes a relatively sane hashCode implementation, if not a sane random implementation :)

verifier/src/test/java/com/aws/jverify/verifier/tests/Records.java
robin-aws, line 1: Worth including at least one record that implements an interface, and one that attempts to extend an interface with `@Modifiable` and hence causes an error.

verifier/src/test/java/com/aws/jverify/verifier/tests/TranslationErrors.java
```diff
@@ -108,6 +111,57 @@ int switchThrowBody(int i) {
             default -> i * i * i;
         };
     }
-    
+
+    static boolean nullablePrimitive(@Nullable int i) {
+//                                   ^ error: nullable primitive type is not supported
```
robin-aws, line 115: I'd clarify that this one is actually a property of Java semantics, whereas the ones below are current implementation limitations.

verifier/src/main/resources/com/aws/jverify/messages.properties
```diff
@@ -13,4 +13,5 @@ compiler.err.concreteTypeWithExternalContract=class ''{0}'' must not have an ext
 compiler.err.unusedContractMethod={0} is part of a @Contract class, but its signature does not match any method from the contractee
 compiler.err.noContractTarget=could not find a target for @Contract class ''{0}''
 compiler.err.annotationOnSourceContractClass=annotation ''{0}'' on @Contract class ''{1}'' is not allowed, because it must be placed on the contractee
-compiler.err.bodylessMethodWithoutContract=method ''{0}'' without a body still needs a contract. Using @Contract to specify a contract class.
\ No newline at end of file
+compiler.err.bodylessMethodWithoutContract=method ''{0}'' without a body still needs a contract. Using @Contract to specify a contract class.
+compiler.err.equalityOperatorRestricted=''{0}'' is not allowed when both operand''s types could be String, a record, or a boxed primitive, and neither operand is a null literal
```
robin-aws:
```suggestion
compiler.err.equalityOperatorRestricted=''{0}'' is not allowed when both operand''s types could be String, a record, or a boxed primitive, unless one of the operands is a null literal
```

verifier/src/main/java/com/aws/jverify/verifier/MethodOrLoopContract.java
```diff
@@ -2,6 +2,7 @@
 
 import com.aws.jverify.generated.*;
 import com.sun.tools.javac.tree.JCTree;
+import org.checkerframework.checker.nullness.qual.Nullable;
```
robin-aws: Unused?

robin-aws (commented):
Partial review

verifier/src/test/java/com/aws/jverify/verifier/tests/RecordsErrors.java
alex-chew: The naming conflict wasn't my concern; rather, it's because I wanted to translate record accessor calls as datatype destructors rather than as datatype member functions/methods, and also because my impression is that explicit accessors aren't used much.

If we translated record accessors (explicit or implicit) as datatype member methods, then they'd be harder to use in pure contexts (specifically, within expressions); and if we translated them as datatype member functions, then we've have to require that they're `@Pure`. Neither option seemed worth the hassle, but AFAICT they're feasible.

verifier/src/test/java/com/aws/jverify/verifier/tests/Records.java
alex-chew: I've expanded the scope of #179 to include these other types - String, record types, and (soon) sealed interfaces - and added a link to the comment here.

verifier/src/test/java/com/aws/jverify/verifier/tests/PolymorphismWithoutBounds.java
alex-chew: That does work, and it turns out there's an even simpler way - making `obj` an instance of a dummy class lets everything else verify as it was before!

verifier/src/test/java/com/aws/jverify/verifier/tests/Records.java
alex-chew: Good catch, I've added those tests (and implemented a check for `@Modifiable` during translation).

verifier/src/test/java/com/aws/jverify/verifier/tests/RecordsErrors.java
robin-aws: Right. I agree they should be pretty rare so this might be good enough for good.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -181,6 +181,7 @@ public static void verifyFile(SourceFile markedSourceFile, JVerifyTest annotatio
                         : Stream.of(diagnostic))
                 .map(JVerifyTestEngine::diagnosticAsAnnotatedRange)
                 .sorted()
+                .distinct()
```
robin-aws: I wouldn't do this: having duplicate diagnostics will be a customer-facing annoyance, and if it happens too often will erode trust, so we shouldn't be blind to them in tests.

In the worst case if we have duplicates that are really expensive to get rid of, you can repeat the hat annotation line in the test file (I made it support multiple consecutive annotations that apply to the same line)

javaTypesGenerator/src/main/java/com/aws/jverify/generator/CSharpToJavaConverter.java
```diff
@@ -311,8 +308,12 @@ private static String convertCSharpTypeToJava(String csharpType) {
         typeMap.put("Node", "Node");
         typeMap.put("NodeWithComputedRange", "NodeWithComputedRange");
         typeMap.put("Attributes", "Attributes");
+        //noinspection Java9CollectionFactory
```
robin-aws: What's this warning about?

verifier/src/main/java/com/aws/jverify/verifier/BlockCompiler.java
```diff
@@ -481,26 +482,27 @@ public List<JCTree.JCStatement> translateHeader(List<JCTree.JCStatement> stateme
...
-                        
                         var type = compiler.translateType(null, parameter.type, compiler.toOrigin(parameter));
-                        var postconditionPredicate = compiler.expressionCompiler.toExpr(lambda.getBody());
-                        var condition = new LetExpr(origin, List.of(new CasePattern(origin, paramName, 
-                                new BoundVar(origin, new Name(origin, paramName), type, false), null)), 
-                                List.of(new NameSegment(origin, compiler.nameCompiler.METHOD_RETURN_VARIABLE_NAME, null)), postconditionPredicate, true, null);
-                        
-                        if (postconditionPredicate != null) {
+                        var origCondition = compiler.expressionCompiler.toExpr(lambda.getBody());
+                        if (origCondition != null) {
```
robin-aws: Necessary or just an optimization?

verifier/src/main/java/com/aws/jverify/verifier/ExpressionCompiler.java
```diff
@@ -289,6 +362,41 @@ public Expression translateBinary(JCTree node,
...
+     * Returns whether a value of the given type is possibly a "Java-Reference-as-Dafny-Value" type ("JRDV type").
+     */
+    private boolean isPossiblyJrdvType(com.sun.tools.javac.code.Type type) {
+        if (type.isPrimitive()) {
+            return false;
+        }
+
+        var types = Types.instance(this.compiler.context);
+        return Stream.concat(primitiveTypes().stream(), jrdvTypes().stream())
+                .anyMatch(t -> types.isAssignable(type, t) || types.isAssignable(t, type));
```
robin-aws, line 375: Why do you check for subtyping in both directions?

verifier/src/main/java/com/aws/jverify/verifier/ExpressionCompiler.java
```diff
@@ -257,22 +277,75 @@ yield new SeqSelectExpr(origin, false, receiver,
...
+
+            // If one operand is definitely not of a JRDV type,
+            // then since we already checked that it's not of primitive type,
+            // we know that it's of reference type and will be translated to a Dafny reference type.
+            // The type of the other operand doesn't matter:
+            // if it's of JRDV type then the comparison will be rejected during Dafny resolution.
+            isSafe |= !isPossiblyJrdvType(leftType) || !isPossiblyJrdvType(rightType);
+
+            if (!isSafe) {
+                compiler.reportError(node, "equalityOperatorRestricted", opName);
```
robin-aws, line 343: First off, excellent comments :)

I understand this is allowing as much as possible to be translated while staying sound. But I wonder if the UX would be better if we rejected more at this point, because we can know it will always cause a Dafny resolution error (for example if you have `<record value> == null`).

Fine for future improvement, but I mention it in case this could be simpler and therefore need less thorough testing.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -834,6 +838,104 @@ private static boolean isEnum(com.sun.tools.javac.code.Type type) {
...
+
+        return new IndDatatypeDecl(origin, name, null, typeParams, members, traits, ctors, false);
+    }
+
+    /**
+     * Returns whether the declaration is a record's synthetic (implicit) canonical constructor.
+     */
+    private static boolean isSyntheticCanonicalConstructor(JCTree.JCMethodDecl methodDecl) {
+        // Ideally we'd check for the SYNTHETIC flag, but it's not set.
+        // So instead we check for its body: just a lone "super()" call.
```
robin-aws, line 930: FYI `methodDecl.getModifiers().flags & Flags.SYNTHETIC` should work AFAICT. It's just that "synthetic" doesn't show up in the result of `methodDecl.getModifiers().getFlags()` because it's not an actual flag in Java source.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -834,6 +838,104 @@ private static boolean isEnum(com.sun.tools.javac.code.Type type) {
...
+
+    private IndDatatypeDecl translateRecord(JCTree.JCClassDecl classDecl, IOrigin origin, Name name) {
+        assert classDecl.getKind() == Tree.Kind.RECORD;
+        if (isAnnotatedRecursive(classDecl.type, Modifiable.class)) {
+            reportError(origin, "modifiableForbidden", "a record class");
+        }
+
+        var typeParams = translateTypeParameters(classDecl.typarams);
+
+        var traits = getCurrentTypeSymbol(classDecl)
```
robin-aws, line 854: Do we support external contracts on record types? Not sure how we'd easily test that. Might be reasonable to track this as a separate issue.

robin-aws (changes requested):
Finished the review. Overall looks great, will be a very important feature to support!

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
alex-chew: Hmm, I've tried `methodDecl.getModifiers().flags & Flags.SYNTHETIC` and it yields `0` for me. I wonder if it could be a difference in JDK distribution? I'm running Corretto 23.0.2 for Windows x64.

verifier/src/main/java/com/aws/jverify/verifier/BlockCompiler.java
alex-chew: Looks like it's obsolete since `toExpr` shouldn't return null in the first place, so I'm removing it.

verifier/src/main/java/com/aws/jverify/verifier/ExpressionCompiler.java
alex-chew: It's mainly for records: the method should return true whether `type` is `java.lang.Object` or a `record MyRecord`, so we check both directions from `java.lang.Record`.

(Since the primitive types and String are both non-extendable, it's enough to just check the one direction for those - I just combined them for simplicity.)

javaTypesGenerator/src/main/java/com/aws/jverify/generator/CSharpToJavaConverter.java
alex-chew: It's just a style warning, though now that I look at it I actually prefer what the inspection was suggesting, so I'll update it.

verifier/src/main/java/com/aws/jverify/verifier/ExpressionCompiler.java
alex-chew: I think the UX could be slightly better if we reject more during translation than during Dafny resolution, but only when it lets us provide better context in error messages (e.g. with more relevant info from the Java AST instead of the post-translation Dafny AST). On the other hand I feel that it'd actually require more complexity and testing if we wanted to specially handle cases like `<record value> == null`, but maybe there's a simpler way.

In any case I'll leave that for future improvement.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
alex-chew: That's a good point, and it looks like it's not needed anyway (I think maybe I was running into duplicates when I hadn't quite figured out how to handle member functions or something like that.)

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
alex-chew: Yeah, I don't see a way to test that without some non-trivial work to add test symbols outside the current `JavacTrees` instance (or an annotation to treat a record class as such, for example), so I've added a separate issue here: https://github.com/aws/jverify/issues/218

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
robin-aws: Oh could be. And I'm only relating what I saw from the output of the UNLAMBDA phase, it's possible javac doesn't put that flag on the synthetic constructors you're trying to detect.

robin-aws (approved):

PR #201, opened 2025-06-16
Only used complex compiled names when necessary

keyboardDrummer:
### What was changed?
- Renamed `MethodCompiler` to `BlockCompiler`, in preparation for creating an actual `MethodCompiler` in another PR
- Only qualify type names when that type name occurs more than once in the compiled output
- Only append parameter types to a method name if that method has overrides
- Only append field or method prefixes to either if there is a name collision
- No longer use the term `mangling` when determining compiled names, since we try to keep the original names.

### How has this been tested?
- Manually inspected names of generated code, to ensure this helps debuggability
- Relies on existing tests to ensure nothing breaks

### DIFF
--- a/verifier/src/main/java/com/aws/jverify/verifier/NameCompiler.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/NameCompiler.java
@@ -70,12 +70,13 @@ public class NameCompiler {
         if (symbolStringMap.containsKey(s)) {
             return symbolStringMap.get(s);
         }
-        var mangled = uncachedMangleSymbolName(s);
-        reverseSymbolStringMap.put(mangled, s);
-        return mangled;
+        var compiledName = uncachedGetSymbolName(s);
+        symbolStringMap.put(s, compiledName);
+        reverseSymbolStringMap.put(compiledName, s);
+        return compiledName;
     }
     
-    private String uncachedMangleSymbolName(Symbol s) {
+    private String uncachedGetSymbolName(Symbol s) {
         switch (s) {
             case Symbol.ClassSymbol classSymbol -> {
                 return classSymbol.getQualifiedName().toString().replace(".", "_");
@@ -85,7 +86,7 @@ public class NameCompiler {
             }
             case Symbol.VarSymbol varSymbol -> {
                 if (varSymbol.getKind().isField()) {
-                    return mangleFieldName(varSymbol);
+                    return getFieldName(varSymbol);
                 } else { // Local variable, do not mangle
                     return s.name.toString();
                 }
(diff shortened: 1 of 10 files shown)


verifier/src/main/java/com/aws/jverify/verifier/NameCompiler.java
```diff
@@ -17,117 +17,159 @@
  *   - a suffix is added to all method names. This suffix encodes the parameter types, as given by the typeMangling
  *     method
  */
-public class NameMangler {
+public class NameCompiler {
     static private final String fieldPrefix = "F_";
     static private final String methodPrefix = "Z_";
     public static final String CTOR_PREFIX = "_ctor_";
 
     // Map from symbols to mangled names
```
olivier-aws: Comment refers to the symbolStringMap. Consider adding a proper comment for classNameOccurrences

verifier/src/main/java/com/aws/jverify/verifier/NameCompiler.java
```diff
@@ -17,117 +17,159 @@
  *   - a suffix is added to all method names. This suffix encodes the parameter types, as given by the typeMangling
```
olivier-aws: This comment needs to be updated to explain the new algorithm that only updates when needed

olivier-aws (approved):
On a high-level, I think it is a little risky to add complex logic to decide on how we translate symbol names. There are less certainty that we cannot have any name clash that would result in a non-understandable error message to the user.
I'm fine with submitting this (with more comments to explain the new algorithm), but let's be ready to have it under a debug flag if we hit new name clashes when we extend our test base.

verifier/src/main/java/com/aws/jverify/verifier/NameCompiler.java
keyboardDrummer: Renamed the variable to make the meaning more clear

keyboardDrummer:
> On a high-level, I think it is a little risky to add complex logic to decide on how we translate symbol names.
> ... let's be ready to have it under a debug flag if we hit new name clashes when we extend our test base.

It sounds like you're saying it'll be too difficult to find all the buggy cases. That seems pessimistic to me. I'm expecting that if there is a bug or two, we can add a test for those and that will be the end of it.

In Dafny there is one feature that has been a source of many bugs: the match constructors. They depend on copying the contents of the match cases, and that copying code is by default not invoked anywhere else and is different for every language construct, so to properly test it, you have to try out every Dafny language construct inside the match case and our tests are not doing that.

So, my feeling is that brittle features arise when you have a interaction between features (match constructs and all other features in this case), but I don't see that pattern in this PR: the rules for scopes for different kind of symbols in Dafny seem clear.

PR #202, opened 2025-06-16
More robust isNullable check

keyboardDrummer:
### What was changed?
- More robust isNullable check

### How has this been tested?
- TODO. Should we use GH actions with OpenJDK as well? Or always use OpenJDK?


robin-aws (changes requested):
Ah I wondered about why we had your version before but it didn't seem to work on my Corretto JDK.

The change looks good, but I think you should also change the CI to run on both JDKs so there's testing coverage. Maybe even verify it fails without this change first?

fabiomadge:

PR #209, opened 2025-06-19
Customer b fixes

keyboardDrummer:
### What was changed?
1. Make several not supported errors more specific
2. Do not crash when encountering a bodyless pure method without a contract
3. Fix support for adding jar files to the class path
4. Fix a crash when not supported types (like a wildcard) were used as type arguments.
5. Support method references as the outer expression of a postcondition.
6. As part of the above change, also support postconditions that use lambdas with different return value names. We're now always naming the return value `#_r` and translating `postcondition((<Type> <Name>) -> <Body>)` to `var <Name>: <Type> := #_r; <Body>`
7. As part of the above changes, also allow local variables to have the same name as the variable used in a postcondition lambda
8. Update the code that generates the Java classes representing the Dafny AST, to also support typed with generic type arguments, like `A<B<C>>`. This was needed for correctly generating `LetExpr`

### How has this been tested?
- Intentionally have no test for change 1
- Add test `InterfacesTranslation.java` to test change 2
- Skipping a test for change 3 for now since it's expensive to write (need to have a program that needs a jar. Which jar? Where will I find that jar?). One way in which we'll cover this is by dogfooding JVerify on its own codebase.
- Add test `Wildcards.java` to test change 4
- Add test `MethodContractsVerification.java` and move tests from `ResolutionErrorsNameClash.java` into this, to test changes 5/6/7
- Change 8 is tested implicitly by the tests that use JVerify features, since they are using the code generated by change 8.


verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -132,7 +134,7 @@ private Iterable<? extends CompilationUnitTree> parseResolveAndDesugarJava(Verif
         var classpath = classpathEntries.stream()
                 .map(Path::toString)
                 .collect(Collectors.joining(File.pathSeparator));
-        var javacOptions = List.of("-classpath", classpath);
+        var javacOptions = List.of("-cp", classpath);
```
olivier-aws, line 137: Why this change? Isn't this equivalent?

verifier/src/main/resources/com/aws/jverify/messages.properties
```diff
@@ -13,4 +12,5 @@ compiler.err.internalAndExternalContractForMethod=method ''{0}'' already has an
 compiler.err.concreteTypeWithExternalContract=class ''{0}'' must not have an externally defined contract because all its contracts can be defined internally
 compiler.err.unusedContractMethod={0} is part of a @Contract class, but its signature does not match any method from the contractee
 compiler.err.noContractTarget=could not find a target for @Contract class ''{0}''
-compiler.err.annotationOnSourceContractClass=annotation ''{0}'' on @Contract class ''{1}'' is not allowed, because it must be placed on the contractee
\ No newline at end of file
+compiler.err.annotationOnSourceContractClass=annotation ''{0}'' on @Contract class ''{1}'' is not allowed, because it must be placed on the contractee
+compiler.err.bodylessMethodWithoutContract=method ''{0}'' without a body still needs a contract. Using @Contract to specify a contract class.
```
olivier-aws, line 16: Consider using the active form for suggested actions: `Use @Contract to specify a contract class.`

verifier/src/test/java/com/aws/jverify/verifier/tests/MethodContractsVerification.java
```diff
@@ -0,0 +1,38 @@
...
+import com.aws.jverify.Pure;
+import com.aws.jverify.testengine.JVerifyTest;
+
+import static com.aws.jverify.JVerify.postcondition;
+
+@JVerifyTest(dafnyVerified = 4, dafnyErrors = 0)
+public class MethodContractsVerification {
+    
+    public int methodReferencePostCondition() {
+        postcondition(MethodContractsVerification::isEven);
```
olivier-aws, line 12: Nice!

verifier/src/test/java/com/aws/jverify/verifier/tests/TranslationErrors.java
```diff
@@ -20,13 +20,6 @@ void pureWithoutReturn() {
         var x = 3;
     }
     
-    int multipleReturnNames() {
-       postcondition((Integer i) -> i > 0);
-       postcondition((Integer j) -> j < 2);
-//                            ^ error: all postcondition lambda parameters must use the same name. Got j instead of i
-       return 1;
-    }
-
```
olivier-aws, line 29: Cool to have removed this limitation

olivier-aws (approved):
LGTM. Just a minor comment on an error message

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: I think it is. I thought it was a bit confusing because long names usually use `--` instead of `-`.

verifier/src/main/resources/com/aws/jverify/messages.properties
keyboardDrummer: I'll do it in a follow-up: https://github.com/aws/jverify/pull/211

PR #211, opened 2025-06-20
In testing, enable producing valid dafny names

### LINKED ISSUE #205: Enable outputting Dafny which contains valid Dafny names

keyboardDrummer:
Fixes #205 

### What was changed?
- In testing, by default let the generated Dafny use names that are valid Dafny as well, at the risk of getting name collisions if the test uses colliding names. It would be nice to have this as a user facing feature as well, but then it is more important to automatically resolve name collisions by creating unique valid names. I'll leave this for future work.
- Remove the test `MultiplePostconditionsSameNames.java` since it is made obsolete by `MethodContractsVerification.java`
- Simplify translation of return statements with expression bodies
Previously:
```dafny
  method methodReferencePostCondition() returns (#_r: int32)
    ensures MethodContractsVerification.isEven(#_r)
    ensures var r: int32 := #_r; r == 2
  {
    var #_tmpVar_0;
    #_tmpVar_0 := 2;
    return #_tmpVar_0;
  }
```

now

```dafny
  method methodReferencePostCondition() returns (result: int32)
    ensures MethodContractsVerification.isEven(result)
    ensures var r: int32 := result; r == 2
  {
    result := 2;
    return;
  }
```

### How has this been tested?
- Added a new test `GeneratedNameCollisions.java` that tries to trigger name collisions from generated names. It turns off the valid Dafny names option to check that this prevents the collisions


verifier/src/test/java/com/aws/jverify/verifier/tests/ShouldVerify.java
```diff
@@ -5,12 +5,18 @@
...
+@JVerifyTest(exitCode = 4, dafnyVerified = 2, dafnyErrors = 5)
 public class ShouldVerify {}
 
 @Verify(value = true, overrideChildren = true)
 class ShouldVerify1 {
     
+    @Verify(false)
+    public ShouldVerify1() {
+        check(false);
+//      ^^^^^^^^^^^^ Error: assertion might not hold
```
olivier-aws, line 17: Why do we need this case on top of the one one line 22?

verifier/src/test/java/com/aws/jverify/verifier/tests/GeneratedNameCollisions.java
```diff
@@ -0,0 +1,44 @@
+package com.aws.jverify.verifier.tests;
+
+import com.aws.jverify.Pure;
+import com.aws.jverify.testengine.JVerifyTest;
+
+import static com.aws.jverify.JVerify.postcondition;
+
+@JVerifyTest(dafnyVerified = 2, dafnyErrors = 0, avoidNameCollisions = true)
```
olivier-aws, line 8: What happens if you remove the `avoidNameCollisions` flag? Does the test fail?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -50,8 +50,6 @@
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
-import static com.aws.jverify.verifier.NameCompiler.CTOR_PREFIX;
-
 public class JavaToDafnyCompiler {
     public static final String JVERIFY_CLASS = JVerify.class.getName();
     public static final String METHOD_RETURN_VARIABLE_NAME = "#_r";
```
olivier-aws: Seems useless now

verifier/src/test/java/com/aws/jverify/verifier/tests/GeneratedNameCollisions.java
```diff
@@ -0,0 +1,44 @@
...
+public class GeneratedNameCollisions {
+
+    public int differentReturnValueNames() {
+        postcondition(this::predicate);
+        postcondition((Integer result) -> result < 3);
+        int result;
```
olivier-aws, line 14: Can you add a local variable with name `_g_result`, which is equal to `METHOD_RETURN_VARIABLE_NAME` in this case? Is there a name clash then?

olivier-aws (commented):
Overall nice change with good simplification of the name handling. Still puzzled by the fact that you don't need to generate a name that starts with '#' or another illegal Java character to completely avoid name clash.

verifier/src/test/java/com/aws/jverify/verifier/tests/ShouldVerify.java
keyboardDrummer: Previously `@Verify` was illegal on constructors

verifier/src/test/java/com/aws/jverify/verifier/tests/GeneratedNameCollisions.java
keyboardDrummer: Yes, it detects all the collisions that we're adding `_` for.

verifier/src/test/java/com/aws/jverify/verifier/tests/GeneratedNameCollisions.java
keyboardDrummer: Added it.

olivier-aws: If the variable name is `_g_result`, there is a name clash and a crash:
`Error: Duplicate local-variable name: _g_result, range=15:9-15:23`

keyboardDrummer: Thanks. I've added more `_` prefixed cases to the test, and fixed the issue.

olivier-aws (dismissed):
LGTM, thankls for the fixes. Some CI tests are still failing though.

olivier-aws (approved):

PR #217, opened 2025-06-23
Add support for Overrides

keyboardDrummer:
### What was changed?
- Add support for Overrides

### How has this been tested?
- Updated tests `ClassesExtendingClassesVerification` and `ClassesExtendingClassesErrors`


verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -738,7 +752,22 @@ private List<ClassLikeDecl> buildTraitAndClassTwin(
...
                     classMembers.add(classConstructor);
                 }
-                case null, default -> traitMembers.add(member);
+                default -> traitMembers.add(member);
+            }
+        }
+        
+        for(var inherited : methodMembers) {
+            var intermediateMember = intermediateMethodsToSymbols.get(inherited);
+            if (intermediateMember instanceof Method method && !method.getHasStaticKeyword()) {
```
robin-aws: Why do you need to redeclare the method on this type if it isn't actually declared in source?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -738,7 +752,22 @@ private List<ClassLikeDecl> buildTraitAndClassTwin(
...
+                if (!classMethodNames.add(method.getNameNode().getValue())) {
+                    continue;
+                }
+                var declaration = new Method(method.getOrigin(), method.getNameNode(), getVerifyFalse(method.getOrigin()), false, null,
```
robin-aws: It would be safer to add `{:axiom}` to the method instead of `{:verify false}`. Is the latter actually necessary?

robin-aws (commented):
Just a high level question for understanding, as I'm reviewing https://github.com/dafny-lang/dafny/pull/6279

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: Turns out we don't need this anymore. I had this before I made the Dafny change.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: Code was removed

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -258,7 +258,7 @@ public static void runDafnyProcess(NameCompiler nameCompiler,
                 "--json-output",
                 "--type-system-refresh",
                 "--general-newtypes",
-                // "--progress", "Batch",
+                "--progress", "Batch",
```
robin-aws: Probably don't want this checked in? Harmless but could slow down CI over time.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -112,7 +114,14 @@ public NameCompiler getNameCompiler() {
         List<FileStart> filesStarts = new ArrayList<>();
         for (var compilationUnit : parsed) {
             List<TopLevelDecl> fileDeclarations = declarationsForFile.get(compilationUnit);
-            // fileDeclarations.sort(t -> ((SourceOrigin)t.getOrigin()).getEntireRange().getStartToken());
+            fileDeclarations.sort(Comparator.comparing(t -> {
```
robin-aws, line 115: Nice!

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -77,6 +77,8 @@ public class JavaToDafnyCompiler {
 
     private final List<Symbol.MethodSymbol> invariants = new ArrayList<>();
     private final List<JCTree.JCVariableDecl> initializers = new ArrayList<>();
+    private final Map<Symbol.MethodSymbol, MemberDecl> intermediateMethodsToSymbols = new HashMap<>();
+    private int implementationIndex = 0;
```
robin-aws: Completely unused?

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -77,6 +77,8 @@ public class JavaToDafnyCompiler {
 
     private final List<Symbol.MethodSymbol> invariants = new ArrayList<>();
     private final List<JCTree.JCVariableDecl> initializers = new ArrayList<>();
+    private final Map<Symbol.MethodSymbol, MemberDecl> intermediateMethodsToSymbols = new HashMap<>();
```
robin-aws: Only populated and not read elsewhere AFAICT, part of an earlier attempt to work around this not being supported in Dafny?

verifier/src/test/java/com/aws/jverify/verifier/tests/ClassesExtendingClassesVerification.java
```diff
@@ -69,3 +75,18 @@ public int computeX(int input) {
         }
     }
 }
+
+interface IExtendee {
+    default int virtualMethod() {
+        postcondition((Integer r) -> r >= 10);
```
robin-aws, line 86: As elsewhere, worth adding a case of successful verification that depends on an override with a stronger spec

robin-aws (changes requested):

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: Part of an attempt to split the pipeline into 

phase 1: Java -> Java, 
phase 2: Java -> Dafny
phase 3: Dafny -> Dafny

and to have available in phase 3, on the Dafny AST, the symbol information from the Java resolver.

But I'll remove this now

PR #222, opened 2025-07-02
Add syntactic but not semantic support for wildcards

keyboardDrummer:
### What was changed?
- Add syntactic but not semantic support for wildcards. This enables working with code that contains wildcards, but only if you don't make use of them.
- Improve diagnostic location for diagnostics on return types

### How has this been tested?
- Added the test `Wildcards.java`


docs/guide/src/supported_java_features.md
```diff
@@ -60,7 +60,7 @@ Nothing yet
 - `checked exceptions` (ignored)
 - `synchronized`
 - Module declarations
-- Wildcards: `?`
+- Wildcards: `C<? extends X>` is interpreted as `C<X>`, and similarly for `C<? super X>`. `C<?>` is translated to a type that can not be instantiated.
```
robin-aws: Translating `C<?>` to `C<Object>` instead might support slightly more cases, and is really just a special case of the `extends` rule. Then you don't have to worry about hiding "Never" in any error messages that occur.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
```diff
@@ -1469,6 +1469,17 @@ public Type translateType(JCTree.JCModifiers modifiers, com.sun.tools.javac.code
...
             }
+            case com.sun.tools.javac.code.Type.WildcardType wildcardType -> {
+                var extendsBound = wildcardType.getExtendsBound();
+                if (extendsBound != null) {
+                    return translateType(extendsBound, origin);
+                }
+                var superBound = wildcardType.getSuperBound();
+                if (superBound != null) {
+                    return translateType(superBound, origin);
+                }
```
robin-aws: Actually, this isn't sound, because it's an under-approximation of possible runtime values:

```java
class General {
  int foo() {
    postcondition((int r) -> r >= 0);
    return 0;
  }
}

class Specific extends General {
  int foo() {
    postcondition((int r) -> r > 0);
    return 1;
  }
}

void <T super Specific> uhOh(T t) {
  var d = t.foo();
  var x = 1 / d;   // exception if t is a General.
}
```

We should just ignore super bounds and only substitute for extends bounds.

robin-aws (changes requested):

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
keyboardDrummer: I don't think the above code is valid Java. Type parameters bounds and variance (wildcards) are two different features. Type parameters bounds can be used where you define type _parameters_, and only accept `extends`. They allow you to write something like `void <T extends General> foo()`. Variance can only be used where you specify type _arguments_, at the usage site. They allow you to write things like `List<? super Specific>`.

I can change your example to:

```java
void uhOh(List<? super Specific> specifics) {
    Specific t = specifics.get(0); // Java compiler gives an error, 
    // because this getter returns a `?` and that can't be assigned to a variable of type `Specific`
    var d = t.foo();
    var x = 1 / d;   // exception if t is a General.
}
```

so then it's syntactically correct, but it still gives an error.

docs/guide/src/supported_java_features.md
keyboardDrummer: I'll do that.

verifier/src/main/java/com/aws/jverify/verifier/JavaToDafnyCompiler.java
robin-aws: Serves me right for rushing the counterexample and not actually . :) We MIGHT be safe because we can rely on both the Java type checker and the Dafny type checker.

I'm still not fully convinced we aren't introducing a soundness issue, if not now, then perhaps too easily as we add more features later on. For example, Dafny allows declaring possibly-empty types. I don't think we declare any newtypes with `witness *` AFAICT, but that could easily change in the future. Here's at least a future soundness issue I'm worried about:

```java
// Java:

void foo(Supplier<? super Empty> p) {
    Object whatever = p.get();
    check(false);
}

void bar() {
    Supplier<Object> s = () -> (Object)"some Object";
    foo(s);
}
```

```dafny
// Dafny:

// What java.lang.Object is translated to
trait Object {}  

// A translation of a subset type, or some other Java type that turns out to be possibly empty
newtype Empty extends Object = x: bool | false witness *

trait Supplier<T> {
  method get() returns (r: T)
}

method foo(p: Supplier<Empty>) {
  var whatever: Object := p.get();
  assert false;
}

// bar omitted
```

In Java foo is clearly possible to call, because it's possible to create a supplier of some super type of Empty. But in Dafny it ends up looking like foo is impossible to call, and hence you can prove false.

This is clearly very contrived and a very weird use of wildcard lower bounds. But it is at least an example of where losing precision in types can lead to unsoundness.

keyboardDrummer: > But it is at least an example of where losing precision in types can lead to unsoundness.

I don't think it is unsound. In your example, I think the method `bar` does not pass Dafny's resolution phase, because the call to `foo` needs to pass in a `Supplier<Empty>`, while you're constructing a `Supplier<Object>`. That the error is in `bar` instead of in `foo` seems fine to me. 

Error in bar, from Dafny's resolver:
```dafny
method foo(p: Supplier<Empty>) {
  var whatever: Object := p.get();
  assert false;
}

method bar() {
  var s: Supplier<Object> := ...;
  foo(s); // error: types don't match.
}
```


If you remove the wildcards from your example, I think you would get an error in the same place, but then from Java's instead of from Dafny's resolution. So even if you don't like this behavior, it does not seem to relate to wildcards.

Error in bar, from Java's resolver:
```java
void foo(Supplier<Empty> p) {
    Object whatever = p.get();
    check(false);
}

void bar() {
    Supplier<Object> s = () -> (Object)"some Object";
    foo(s); // Java resolution error: types don't match
}
```

robin-aws: True. But `bar` could easily be in unverified Java code, so we can't rely on rejecting it to avoid the unsoundness.

I will put some thought into what the safe alternative is here. I'm at least a lot less worried about ignoring lower bounds, even though it means we can't handle a lot of streams definitions for e.g.

keyboardDrummer: > True. But `bar` could easily be in unverified Java code, so we can't rely on rejecting it to avoid the unsoundness.

That only holds if we think the error should be in `foo`, not in `bar`, which I think we didn't discuss so much, but I agree (!!! 🥳 ) it should be in `foo` and it should be on the assertion.

> I will put some thought into what the safe alternative is here

What if `super` in contracts is only allowed in unverified code? I think that would avoid the unsoundness, and still allow us to specify contracts for the Streams API. Example:

```java
void foo(Supplier<? super Empty> p) 
//                  ^ error: JVerify does not support 'super' in parameters of verified methods     
{
    Object whatever = p.get();
    check(false);
}

@Verify(false)
void bar() {
    Supplier<Object> s = () -> (Object)"some Object";
    foo(s);
}

@Contract(Stream.class)
class StreamContract<T> {
  Stream<T> filter(Predicate<? super T> predicate) {
    postcondition((Stream<T> r) -> ...);
  }
}
```

It would still mean that if you use wildcards inside a verified method body, you can get an error at an incorrect location, but at least you still get an error. Example:

```java
void barFoo() {
  Supplier<Object> s = () -> (Object)"some Object";
  Supplier<? super Empty> p = s; // error: Dafny resolution
  Object whatever = p.get();
  check(false); ; // no error, even though this is where it should be
}
```

robin-aws: I think this is headed in the right direction! Especially since I've never actually seen `super` used in the way I did in the example, so I bet we can come up with restrictions that will never block anyone in practice. :)

If we allowed `Predicate<? super T>` in the specification of `filter`, isn't there still a danger of the specification not being well-formed if we accidentally imply `false`?

In `barFoo`, wouldn't we get an error like "JVerify does not support 'super' bounds on wildcards in verified methods ", rather than replacing it with `Empty` and getting the type error since `Empty` is not the same as `Object`? I didn't think you were literally suggesting we only block `super` in types used in method parameters but would still allow them on things like variable types.

keyboardDrummer: > I didn't think you were literally suggesting we only block super in types used in method parameters but would still allow them on things like variable types.

That's what I meant and how I've implemented it now. Why block more than is necessary?

> If we allowed Predicate<? super T> in the specification of filter, isn't there still a danger of the specification not being well-formed if we accidentally imply false?

If you call with something other than `Predicate<T>`, from a verified method, you get a Dafny resolution error at the call-site, and from an unverified method you don't get an error, but then both the caller and callee are not verified so not getting an error is fine.

robin-aws: > That's what I meant and how I've implemented it now. Why block more than is necessary?

The new version is definitely better. I am still uneasy that there is some other case of unsoundness, such as using `super` in a return type. Or even some equally subtle way to break Dafny based on the difference between `? extends T` and `T`.

Let's discuss this further in the sync tomorrow.

keyboardDrummer: I think we're at a stage where if we know something is unsound, we shouldn't do it, but if we don't know then we should continue and if we want to think longer about a potential unsoundness, leave a GH issue to further investigate. We'll have a QA phase later down the road, and then if some of these concerns are still applicable (this one might have already become obsolete by then), then we can determine the risk and whether to block on it.

> such as using super in a return type.

I've disallowed that as well. Added a test for that in a new commit

> based on the difference between `? extends T` and T.

The unsoundness you showed with super relies on the fact that Java has a 'top' type for all types that type parameters can be instantiated with: `object`, which allows you to retrieve the top value from a `C<? super X>`. However, there is no `bottom` type that allows you to insert a value into a `C<? extends T>`, which makes me think there is no symmetrical unsoundness case related to `extends`.

robin-aws: Yup, I wasn't worried about having a symmetrical case for that reason. I AM still worried about the fact that Dafny might be able to reason about the difference between `? extends T` and `T` - in most languages you'd think the latter is strictly more permissive and safer, but given that Dafny can quantify over all values of some types I don't assume it's safe in general. Perhaps something like "all values in a `set<? extends Foo>` have the same concrete type" might have a different answer if you changed it to a `set<Foo>` instead.

Your first point is what convinces me we should merge this in the name of unblocking further development though! I cut https://github.com/aws/jverify/issues/234 and positioned it in the last bucket of tasks before internal release to keep us honest. :)

robin-aws (approved):
WILD

PR #225, opened 2025-07-04
Refactor java to dafny compiler

keyboardDrummer:
### What was changed?
Moved code out of `JavaToDafnyCompiler` into more specific files. Some straightforward moves are to the new files `ClassCompiler` and `JavaFrontEnd` (gets the resolved Java code).

The new package `verifier.compiler` contains classes that compile different parts of the JVerify AST: `JavaToDafnyCompiler` (for a whole project), `ClassCompiler`, `BlockCompiler` and `ExpressionCompiler`. 

This PR moves some of the smart translations out of those general compilation classes and into specific files in the package `verifier.compiler.simplifications`. Each of those simplifications addresses the compilation of a particular JVerify feature.

While ideally the simplifications would be separate compilation phases, or inject themselves into the general compilation classes, right now they are often directly referenced from the general compilation classes.

My intention is for the general compilation classes to contain a dumb mapping from Java to Dafny, while the business logic is in the simplification classes.

####  Review hints
This PR is almost purely moving of code and changing of visibility modifiers. There are a few small logical changes that enable decoupling things better, but I think generally it's most important to review the new structure of the files, more than the individual lines since those did not change in any significant way.

### How has this been tested?
- Pure refactoring, no additional tests needed


verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/temporary/ClassesExtendingClassesCompiler.java
robin-aws, line 1: nit: I don't feel like adding "temporary" to the path has much benefit, when it's always possible this will be the permanent implementation. :) If you really want that I would document why you think it's temporary. Don't want anyone thinking it's temporary because it's actually unsound.

verifier/src/main/java/com/aws/jverify/verifier/compiler/BlockCompiler.java
```diff
@@ -12,17 +16,19 @@
 
 public class BlockCompiler {
 
-    private final JavaToDafnyCompiler compiler;
+    public final JavaToDafnyCompiler compiler;
+    private final List<StatementCompiler> statementSimplifiers = new ArrayList<>();
```
robin-aws: nit: I understand your intention is likely to have these as Java -> Java transformations applied before doing the straightforward Java -> Dafny transformation, but since StatementCompiler is currently Java -> Dafny it doesn't feel right labelling these as simplifiers. In the current state the StatementCompiler abstraction feels like it's just less straightforward than having the cases directly in the `switch (statement) {` block with no actual benefit yet. 

I'd add a comment to explain the future intention.

robin-aws (commented):
Very nice refactoring, thanks! Just some naming nits.

verifier/src/main/java/com/aws/jverify/verifier/compiler/BlockCompiler.java
keyboardDrummer: > nit: I understand your intention is likely to have these as Java -> Java transformations applied before doing the straightforward Java -> Dafny transformation

It would be nice if we could do that, but that might require an AST type which sits between Java and Dafny, so I'm not counting on it.

> In the current state the StatementCompiler abstraction feels like it's just less straightforward than having the cases directly in the switch (statement) { block with no actual benefit yet.

Note that `ImpureStatementExpressionCompiler` matches on `ExpressionStatement` but only _sometimes_, and `ForLoopCompiler` matches on both ForLoop and Continue statements, so it's a little different than having a `case X -> callCompilationClassForX()`

I'm happy to change something but if you want that, could you be super explicit about what you'd like to see?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/temporary/ClassesExtendingClassesCompiler.java
keyboardDrummer: Yeah, I don't like the term either, but I have a desire to mark the transformations that we'd like to get rid of. Is that reasonable? How about I rename the package to `workaround` and add a comment in a `package-info.java`?

I can also fully remove the package, and use a GH issue instead. LMK what you prefer.

verifier/src/main/java/com/aws/jverify/verifier/compiler/BlockCompiler.java
robin-aws: > It would be nice if we could do that, but that might require an AST type which sits between Java and Dafny, so I'm not counting on it.

I'm fairly optimistic we can use the Java AST for many such rewritings, since so much of the javac compiler does that already. But yes some simplifications might not be able to target either the existing Java or Dafny ASTs.

> Note that ImpureStatementExpressionCompiler matches on ExpressionStatement but only sometimes, and ForLoopCompiler matches on both ForLoop and Continue statements, so it's a little different than having a case X -> callCompilationClassForX()

Ah I did indeed miss that, fair enough.

> I'm happy to change something but if you want that, could you be super explicit about what you'd like to see?

To be super pedantic I'd rename `statementSimplifiers` to `statementCompilers`, but that's barely worth the typing energy honestly. :)

PR #230, opened 2025-07-07
Do not verify things from builtin-contracts.java

keyboardDrummer:
### What was changed?
- Do not verify things from builtin-contracts.java

### How has this been tested?
- Updated `builtin-contracts.java` so this change is tested


keyboardDrummer:
Changes seem to trigger an unstable test failure:
```
      message='Error: Internal error occurred during verification: Boogie program had 1 resolution errors:
    Microsoft.Dafny.OverrideCenter: call to undeclared procedure: Call$$_module.Constructable__Object.ctor
```

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -75,14 +77,28 @@ public NameCompiler getNameCompiler() {
...
-        if (parsed == null) {
+        Iterable<JCTree.JCCompilationUnit> parsedIterable = new JavaFrontEnd(this).parseResolveAndDesugarJava(options, files);
+        if (parsedIterable == null) {
             return new FilesContainer(List.of());
         }
+        
+        List<JCTree.JCCompilationUnit> parsed = new ArrayList<>();
+        for (var file : parsedIterable) {
+            parsed.add(file);
+        }
```
robin-aws:
```suggestion
        parsedIterable.forEachRemaining(parsed::add);
```

Or AFAICT we can make parseResolveAndDesugarJava return a Collection instead of just an Iterable and then you can just say `new ArrayList(parsed)`.

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -75,14 +77,28 @@ public NameCompiler getNameCompiler() {
...
+
+        /*
+         * Dafny currently has a bug that will be fixed by this PR: https://github.com/dafny-lang/dafny/pull/6214
+         * To work around this bug, the built-in contracts file must be serialized after 
+         * its users. Because the 's' of 'string://' comes after 'file://', sorting by name achieves this
```
robin-aws, line 97: Modern problems require modern solutions :D

robin-aws (changes requested):
Good change in the right direction (since users shouldn't have to spent cycles verifying the built in contracts), but we also need something in CI that does verify those contracts somehow.

keyboardDrummer:
> Good change in the right direction (since users shouldn't have to spent cycles verifying the built in contracts), but we also need something in CI that does verify those contracts somehow.

Added a test

verifier/src/main/java/com/aws/jverify/verifier/Main.java
```diff
@@ -50,6 +50,9 @@ class AppCommand implements Callable<Integer> {
     @Option(names = "--verify-by-default", description = "Whether to verify code without @Verify(true). Defaults to true.", defaultValue = "true")
     private boolean verifyByDefault;
 
+    @Option(names = "--builtin-contracts", description = "Whether to include built-in contracts for the Java standard library", defaultValue = "true")
```
robin-aws, line 53: Works for now, for the record I think we'll quickly want a more general way to include a Java dependency with contracts that we don't verify, similar to Dafny's `--library` feature. Then the "builtin" contracts can just be a common shared library, one of many.

Don't we want the default value to be "false" though, so users aren't pointlessly re-verifying shared code all the time, and our tests' verified things count is more stable?

verifier/src/main/resources/builtin-contracts.java
```diff
@@ -2,6 +2,7 @@
 
 import com.aws.jverify.Contract;
 import com.aws.jverify.ContractException;
+import static com.aws.jverify.JVerify.check;
```
robin-aws: Unused?

robin-aws (changes requested):

verifier/src/main/java/com/aws/jverify/verifier/Main.java
keyboardDrummer: > Don't we want the default value to be "false" though, so users aren't pointlessly re-verifying shared code all the time, and our tests' verified things count is more stable?

This option means the builtin-contracts are available, and still not verified. Without this, they won't be included, so the default should be true. It's similar to Dafny's `--standard-libraries`.

I'm setting this option to `false` in the test that I've added, which checks that the `builtin-contracts.java` file verifies by passing it as a regular source file.

robin-aws: Ah got it, I misread the option documentation because I'd expected it to control the verification, not whether it was included at all :)

robin-aws (approved):

PR #236, opened 2025-07-09
Integration section user guide

keyboardDrummer:
### What was changed?
- Add a section that explains how to add contracts to third-party code
- Add a section that explains how to partially verify a codebase
- Split the section `Getting started` into smaller parts: `Installation`, `Running JVerify`, `Erase verification code during compilation`
- Add a new section grouping `Adoption`
- Added new draft sections `Example use-cases`, `Testing unverified contracts` and `Runtime checking of contracts`

### How has this been tested?
- Locally inspected the updated documentation
- Added a test for one of the new examples `NullCheck.java`. 
- For the examples `SumCache`, `PositivePrice`, `ValidRanges` I can't add a test yet because we are missing the right sdk contracts. We do still have an issue to test more examples/snippets in the user guide (https://github.com/aws/jverify/issues/146), so we can add these tests later.


docs/guide/src/example_usecases.md
robin-aws, line 1: I'd leave out internal examples, or at least abstract them away from any internal details

docs/guide/src/external_contracts.md
robin-aws, line 1: Nice!

docs/guide/src/SUMMARY.md
```diff
@@ -2,10 +2,19 @@
 
 [JVerify Overview](./overview.md)
 
+# Adoption
```
robin-aws: More like "Integration" - "Adoption" feels like the tool owner's perspective rather than the user's. \

Seems like it should come after Getter started too.

robin-aws (commented):
Something weird about how this change builds - when I do `mdbook serve` locally, "Example use-cases", "Installation" and "Running JVerify" are all blank pages.

docs/guide/src/SUMMARY.md
keyboardDrummer: Okay, I've made those changes. I've struggled with naming this section. My impression is that there's certain high-level questions people have that try to answer the question "should I use this tool?" and I was hoping we could have an early section to try to address those questions. Alternatively or in addition, we could have an FAQ right at the top, under overview, or maybe even in the overview, with those questions and answers that are mostly yes/no with links to more detailed sections.

keyboardDrummer:
> Something weird about how this change builds - when I do `mdbook serve` locally, "Example use-cases", "Installation" and "Running JVerify" are all blank pages.

Ah, sorry I should have done that myself. Fixed the issue now

docs/guide/src/partial_verification.md
```diff
@@ -0,0 +1,75 @@
+# Partial Verification
+
+Given a codebase, you can configure what methods to verify. You can toggle verification on or off at the various levels in your code's hierarchy: `CLI`, `package`, `type`, `method`.
```
robin-aws: I actually wouldn't backquote these terms - feels like it implies they are literally configuration values on options or attributes somewhere.

docs/guide/src/partial_verification.md
```diff
@@ -0,0 +1,75 @@
...
+    void fooCallee() 
+    {
+        modifies(this);
+        i = null;
+        int x = i; // no error, because fooCallee is not verified
+    }
+}
+```
+
+You can enforce an entire hierarchy level to be verified by using the `overrideChildren` argument of `@Verify`. The following code forces any code in the package `my.safe.code` to be verified:
```
robin-aws: Mention this has to be in a package-info.java file too

docs/guide/src/partial_verification.md
```diff
@@ -0,0 +1,75 @@
...
+    @Verify // because this method is verified
+    // we can assume an empty precondition
+    // if none is specified
+    void bar() {
+        // body
+    }
+    
+    // Because this method is not verified,
+    // we can not assume an empty precondition,
+    // so this method missing a precondition
```
robin-aws:
```suggestion
    // so this method missing is a precondition
```

docs/guide/src/partial_verification.md
```diff
@@ -0,0 +1,75 @@
...
+
+When calling a method from verified code, to ensure the caller's code is correct, the callee must always have a precondition, although it may be empty. Here's an example:
+
+```java
+@Verify(false)
+class C {
+    @Verify
+    void foo() {
+        bar(); // no error
+        zaz(); // error, no precondition was specified for `barCallee`
```
robin-aws:
```suggestion
        zaz(); // error, no precondition was specified for `zazCallee`
```

docs/guide/src/external_contracts.md
```diff
@@ -0,0 +1,31 @@
...
+    @Pure
+    BigIntegerContract add(BigIntegerContract v) {
+        postcondition((BigIntegerContract b) -> b.value == this.value + v.value);
+        throw new ContractException();
+    }
+
+}
+```
+
+Note the field `value`, which is not part of the `BigInteger` type. All fields that are part of contract classes are _erased_, meaning they do not occur at runtime, but can be used as part of contracts. The _erased_ concept is further explained in the subsection `Erased` in the section [Code types](code_types.md).
```
robin-aws, line 31: Not blocking, but I think there's a way to link directly to the subsection

docs/guide/src/SUMMARY.md
robin-aws: Yeah I'm not 100% sure either. The FAQ idea could work. I could see having "What is JVerify?" and "Why whould I use JVerify?" towards the top.

robin-aws (changes requested):

docs/guide/src/example_usecases.md
```diff
@@ -0,0 +1,45 @@
...
+
+```java
+{{#include ../../../examples/src/test/java/com/aws/jverify/examples/NullCheck.java}}
+```
+
+#### Keep a cache up to date
+We might store some data that is an aggregate of other data. With JVerify, we can check that the aggregate remains up to date. Here's an example:
+
+```java
+{{#include ../../../examples/src/test/java/com/aws/jverify/examples/SumCache.java}}
```
olivier-aws, line 19: I feel like this test does not pass yet with JVerify, due to lack of support of List. Is this correct? If so, maybe we should add a note about it, so users are not surprised if they test it

examples/src/test/java/com/aws/jverify/examples/ValidRanges.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.stream.IntStream;
+
+import static com.aws.jverify.JVerify.precondition;
+
+@CheckPreconditionsAtRuntime
+record Range(int start, int end, int value) {
+    @Invariant
+    boolean startBeforeEnd() {
+        return start < end;
+    }
```
olivier-aws: Would be nice to have also a static check of this invariant somewhere

docs/guide/src/example_usecases.md
```diff
@@ -0,0 +1,45 @@
...
+
+#### Prove complete program properties
+Sometimes we have programs whose entire desired behavior can be specified with a property that's much simpler than the program itself. Here follows an example of a `findIndex` method that is implemented using binary search. Its desired behavior, to find the index of an element in an array if it exists, is specified in two lines, but the algorithm to compute that result is much more complicated.
+
+```java
+{{#include ../../../examples/src/test/java/com/aws/jverify/examples/BinarySearch.java}}
+```
+
+#### Safely optimize code
+Sometimes we have an algorithm for computing some data, but there's a different algorithm that computes it more quickly, at the cost of being more complex. JVerify allows you to prove that the more compliated algorithm behaves the same as the slower once. Here's an example where that is done for the fibonacci sequence:
```
olivier-aws: nit: more quickly -> faster

docs/guide/src/SUMMARY.md
```diff
@@ -1,10 +1,12 @@
 # Summary
 
 [JVerify Overview](./overview.md)
+[Example use-cases](./example_usecases.md)
```
olivier-aws, line 4: Nice to have this section upfront

docs/guide/src/partial_verification.md
```diff
@@ -0,0 +1,75 @@
+# Partial Verification
+
+Given a codebase, you can configure what methods to verify. You can toggle verification on or off at the various levels in your code's hierarchy: CLI, package, type, method.
```
olivier-aws, line 3: Maybe we should explain why one would want to not verify a method/class. Examples could be:
- the method contains currently unsupported Java features
- the method is too difficult for the verifier and you want to check the rest of the code first

docs/guide/src/partial_verification.md
```diff
@@ -0,0 +1,75 @@
...
+
+When calling a method from verified code, to ensure the caller's code is correct, the callee must always have a precondition, although it may be empty. Here's an example:
+
+```java
+@Verify(false)
+class C {
+    @Verify
+    void foo() {
+        bar(); // no error
+        zaz(); // error, no precondition was specified for `zazCallee`
```
olivier-aws: zazCallee -> zaz

docs/guide/src/testing_unverified_contracts.md
```diff
@@ -0,0 +1,5 @@
+# Testing unverified contracts
+
+If proving that a JVerify method matches its specification is difficult, then you can choose to instead test it using property based testing. This is generally more compute heavy, and is therefor best suited for pieces of code near the bottom of the call-stack. Although it does not require a proof, it does require that the programmer specifies how to generate inputs for the method to test. JVerify supports the generator annotations from the Java property based testing tool [jqwik](https://jqwik.net/).
```
olivier-aws: Do we really already support jqwik annotations?

olivier-aws (commented):
Love the new example_usecases section

docs/guide/src/example_usecases.md
keyboardDrummer: Good point. I'll explicitly state which examples don't work yet.

examples/src/test/java/com/aws/jverify/examples/ValidRanges.java
keyboardDrummer: I'm not sure I follow. I have added another similar example though. Can you check how you feel about the changed examples?

docs/guide/src/testing_unverified_contracts.md
keyboardDrummer: Added a comment at the top warning that we don't have this feature yet.

olivier-aws (dismissed):

examples/src/test/java/com/aws/jverify/examples/MinimumPricePercentage.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class MinimumPricePercentage {
+    // Discounted price should be at least 30% of the original
+    final static BigDecimal minimumRemainingPrice = BigDecimal.valueOf(3, 1);
+    
+    public BigDecimal computePrice(BigDecimal startingPrice, 
+                                   BigDecimal flatDiscount, 
+                                   BigDecimal discountFactor) 
+    {
+        postcondition((BigDecimal result) ->
```
robin-aws: nit: usually better style to have preconditions ordered before postconditions IMO

docs/guide/src/SUMMARY.md
robin-aws: +1!

docs/guide/src/overview.md
```diff
@@ -1,8 +1,36 @@
...
+The specifications and hints used by JVerify are added to the program using regular Java calls to the JVerify library, enabling existing IDEs to be used.
+
+To see examples of bugs that JVerify can detect, have a look at [this section](./example_usecases.md).
+
+# Frequently asked questions
+
+#### Can I use JVerify right now?
+No, JVerify is not yet ready for production use. 
+
+However, you can use the [installation](./installation.md) and [Rrunning JVerify](./first_verification.md) sections in this guide to explore what it can currently do. Note that you will run into poor error messages, unsupported features and bugs - that are not yours ;-).
```
robin-aws:
```suggestion
However, you can use the [installation](./installation.md) and [Running JVerify](./first_verification.md) sections in this guide to explore what it can currently do. Note that you will run into poor error messages, unsupported features and bugs - that are not yours ;-).
```

docs/guide/src/overview.md
```diff
@@ -1,8 +1,36 @@
...
+
+#### Does JVerify support the entire Java language?
+It will. You can find the list of currently supported features here: [supported Java features](./supported_java_features.md).
```
robin-aws: I think we need to qualify this just a little bit. There are likely "features" we will intentionally never support because they aren't worth it. We may also want to point out the distinction between language features and JRE, and that it's impractical to support the entire JRE.

Might also want to point out the version of Java we're targeting here.

docs/guide/src/example_usecases.md
```diff
@@ -0,0 +1,67 @@
+# Example use-cases
+This section contains a number of small code examples that hopefully give you an idea of what types of bugs you can prevent by using JVerify.
+
+The code examples make use of various JVerify features that, unless you've read other parts of this guide, you won't be familiar with. However, most of what you'll see is intuitive so we hope you can follow along. If not, please consult the rest of the guide.
+
+#### Prevent dereferencing null
```
robin-aws, line 6: (not blocking) - I like having the very first example be as simple as possible, but it might be nicer to not immediately have to jump into mutable state specifications (i.e. `modifies`, and understanding why the middle call to `fooer` can't be null but it might be after calling `bar()` because of potential modifications)

docs/guide/src/example_usecases.md
```diff
@@ -0,0 +1,67 @@
...
+Without instructions, JVerify will detect many common types of bugs. Here's some code that shows two `NullPointerException`s being detected by JVerify:
+
+```java
+{{#include ../../../examples/src/test/java/com/aws/jverify/examples/NullCheck.java}}
+```
+
+#### Keep a cache up to date
+We might store some data that is an aggregate of other data. With JVerify, we can check that the aggregate remains up to date. Here's an example:
+
+WARNING: this example does not work yet due to missing contracts for the SDK.
```
robin-aws:
```suggestion
WARNING: this example does not work yet due to missing contracts for the JRE.
```

Or "JCL (Java Class Library)", or perhaps standard library?

docs/guide/src/example_usecases.md
```diff
@@ -0,0 +1,67 @@
...
+```java
+{{#include ../../../examples/src/test/java/com/aws/jverify/examples/ProcessRangesIfAble.java}}
+```
+
+#### Safely interacting with reflective or generated code
+Here's an example that uses the annotation `@CheckPreconditionsAtRuntime` to let contracts from JVerify be checked at runtime, which is useful when code that we don't control calls our verified code. 
+
+In this example, we're using the Jackson library to deserialize JSON into a list of `Range` objects. The library is calling the constructor of `Range`. To make sure it's not constructing invalid `Range` objects, we've marked the record `Range` with `@CheckPreconditionsAtRuntime`, so its constructor checks whether its `@Invariant` holds at runtime.
+
+WARNING: this example does not work yet, since `@CheckPreconditionsAtRuntime` is only available on methods, not on types.
```
robin-aws, line 36: (not blocking) Also reads a bit funny, since it's implicitly the invariant acting as a precondition on the implicit constructor of the record. We might want `@CheckInvariantsAtRuntime` for types instead?

docs/guide/src/example_usecases.md
```diff
@@ -0,0 +1,67 @@
...
+
+```java
+{{#include ../../../examples/src/test/java/com/aws/jverify/examples/ValidRangesJackson.java}}
+```
+
+If we wanted the above error handling to be more granular, like in the "Validating runtime inputs" example, then we could have defined a custom deserializer for Range that calls the default one but handles the `PreconditionFailure`.
+
+When creating a verified library that will be called from unverified code, `@CheckPreconditionsAtRuntime` can be used to ensure calls are made correctly.
+
+#### Prove partial program properties
```
robin-aws: I feel like nit-picking the difference between "partial" and "complete". The example below isn't really fully complete because it's leaving out the performance specification, for e.g. And the "partial" example implies a more complete version is desirable, but often a weaker specification is the right long-term specification. A subset type isn't a partial solution just because it's theoretically possible to use a different exact subset type for every value!

docs/guide/src/example_usecases.md
```diff
@@ -0,0 +1,67 @@
...
+
+#### Prove complete program properties
+Sometimes we have programs whose entire desired behavior can be specified with a property that's much simpler than the program itself. Here follows an example of a `findIndex` method that is implemented using binary search. Its desired behavior, to find the index of an element in an array if it exists, is specified in two lines, but the algorithm to compute that result is much more complicated.
+
+```java
+{{#include ../../../examples/src/test/java/com/aws/jverify/examples/BinarySearch.java}}
+```
+
+#### Safely optimize code
+Sometimes we have an algorithm for computing some data, but there's a different algorithm that computes it faster, at the cost of being more complex. JVerify allows you to prove that the two algorithms return the same result, allowing you to safely use the fast one. Here's an example where that is done for the fibonacci sequence. `spec` is the slow but concise algorithm, while `implementation` is the fast one.
```
robin-aws:
```suggestion
Sometimes we have an algorithm for computing some data, but there's a different algorithm that computes it faster, at the cost of being more complex. JVerify allows you to prove that the two algorithms return the same result, allowing you to safely use the fast one. Here's an example where that is done for the Fibonacci sequence. `spec` is the slow but concise algorithm, while `implementation` is the fast one. 
```

docs/guide/src/installation.md
```diff
@@ -0,0 +1,12 @@
+# Installation
+JVerify can currently only be used by building it from source.
+
+1. Ensure the .NET runtime version 8 or higher is installed on your machine. More information can be found [here](https://dotnet.microsoft.com/en-us/).
+1. Ensure Java23 is installed on your machine. If you have a higher Java version installed add `-Dorg.gradle.java.home=path-to-local-java-to-be-use` to the gradle build command below.
```
robin-aws:
```suggestion
1. Ensure Java 23 is installed on your machine. If you have a higher Java version installed add `-Dorg.gradle.java.home=path-to-local-java-to-be-use` to the gradle build command below.
```

docs/guide/src/testing_unverified_contracts.md
```diff
@@ -0,0 +1,9 @@
+# Testing unverified contracts
+
+WARNING: this feature is not implemented yet
+
+If proving that a JVerify method matches its specification is difficult, then you can choose to instead test it using property based testing. This is generally more compute heavy, and is therefor best suited for pieces of code near the bottom of the call-stack. Although it does not require a proof, it does require that the programmer specifies how to generate inputs for the method to test. JVerify supports the generator annotations from the Java property based testing tool [jqwik](https://jqwik.net/).
```
robin-aws:
```suggestion
If proving that a JVerify method matches its specification is difficult, then you can choose to instead test it using property-based testing. This is generally more compute heavy, and is therefore best suited for pieces of code near the bottom of the call-stack. Although it does not require a proof, it does require that the programmer specifies how to generate inputs for the method to test. JVerify supports the generator annotations from the Java property based testing tool [jqwik](https://jqwik.net/).  
```

docs/guide/src/example_usecases.md
```diff
@@ -0,0 +1,67 @@
...
+WARNING: this example does not work yet due to missing contracts for the SDK.
+
+```java
+{{#include ../../../examples/src/test/java/com/aws/jverify/examples/SumCache.java}}
+```
+
+#### Validating runtime inputs
+Here's an example that parses text into a list of `Range` objects. There is a constraint that needs to hold for each individual range, and a constraint on the list of range objects. Using the method `JVerify.callIfAble`, we can do a runtime check of JVerify contracts, which allows us to concisely enter the safe contractual world without having to repeat preconditions.
+
+WARNING: this example does not work yet, since we don't support `JVerify.callIfAble` yet.
```
robin-aws, line 25: (non blocking) I'm also not convinced this is the exact API we want for this feature. I can imagine strongly wanting information about WHAT precondition failed and where. The example seems overly simplified because I don't think it would ever be good enough to just return false and give no useful error messages.

We can discuss that on https://github.com/aws/jverify/issues/268 of course.

robin-aws (changes requested):
Miscellaneous typos and nit-picks, take the latter with a grain of salt

examples/src/test/java/com/aws/jverify/examples/MinimumPricePercentage.java
keyboardDrummer: That order feels backwards to me. The postcondition tells you what the method provides, so you can figure out if you want to call it, and then once you've decided you want to call it, the preconditions tell you what you need to do. In writing the guide, I more often refer to the postconditions since they are the specification that you want to prove, so it's nice to have them closer to the signature.

Also, when you're defining the method, a good workflow is to start with writing the postcondition, then write the implementation, and then add preconditions to satisfy the proof.

docs/guide/src/overview.md
keyboardDrummer: > There are likely "features" we will intentionally never support because they aren't worth it.

Can you give an example? I think there may be programs where the programmer can reason it's correct, but at the verification level there's not enough information so the program is rejected.

An example could be:

```java
class HasReferenceEquals {
  void user() {
     Object me = this;
     needsReferenceEquals(me, me); // could not prove precondition
     needsReferenceEquals(this, this); // passes
     needsReferenceEquals((HasReferenceEquals)me, me); // passes
  }
 
  static <T extends @Modifiable Object> void needsReferenceEquals(T one, T other) {
    precondition(one != null && implies(one == other, one.equals(other));
  }
}
```

Here the verification depends on type information.

docs/guide/src/example_usecases.md
keyboardDrummer: I did not like seeing `modifies` in the first example either, but I do like showing cases that would not be handled by other null analysis tools, so I went for this. Happy to change if we have a better idea.

docs/guide/src/example_usecases.md
keyboardDrummer: Yeah, good point, I'll leave updating this for when we settle on the design.

docs/guide/src/example_usecases.md
keyboardDrummer: I agree these terms are vague and I don't have a clear mental model of them either. I've removed them and changed the partial example a bit with a different description as well.

docs/guide/src/example_usecases.md
keyboardDrummer: Sounds good.

docs/guide/src/example_usecases.md
keyboardDrummer: changed SDK into "the standard library", which seems to be the most common way of referring to it.

examples/src/test/java/com/aws/jverify/examples/MinimumPricePercentage.java
```diff
@@ -0,0 +1,47 @@
...
+    void discountKeepsThirtyPercentAtLeast(BigDecimal startingPrice,
+                                           BigDecimal flatDiscount,
+                                           BigDecimal discountFactor) {
+        postcondition(() -> {
+//      ^^^^^^^^^^^^^ Related: this is the postcondition that could not be proven
+            var result = computeDiscountedPrice(startingPrice, flatDiscount, discountFactor);
+            return result.compareTo(startingPrice.multiply(minimumRemainingPrice)) >= 0;
+        });
+        
+        preconditionOf(computeDiscountedPrice(startingPrice, flatDiscount, discountFactor));
```
robin-aws: What does `preconditionOf` mean? If it's meant to attach a precondition to another declaration extrinsically, I don't think I want to support that - it means to know the preconditions of something you want to call, you have to look in many more potentially locations in the code.

docs/guide/src/example_usecases.md
robin-aws, line 1: I'm a little concerned that we're advertising examples that use features that are not only not implemented yet, but were we haven't even figured out the right design for them. e.g. `callIfAble`, `preconditionOf`, and even `@Proof`. I don't love adding those library methods just for the sake of the examples, without any javadoc documenting what they mean.

I'd be more comfortable if we had a disclaimer on these use cases and/or the `JVerify` class itself that everything is subject to change.

robin-aws (changes requested):

examples/src/test/java/com/aws/jverify/examples/MinimumPricePercentage.java
keyboardDrummer: It was so that I wouldn't have to copy+paste the preconditions of `computeDiscountedPrice`, which I think would look ugly. You can image a method `getPrecondition(<expr>)` that returns the precondition of `<expr>`, and `preconditionOf(<expr>)` would mean `precondition(getPrecondition(<expr>))`.

I've changed it to `copyPrecondition` now and added some documentation.

docs/guide/src/example_usecases.md
keyboardDrummer: I've removed `@Proof` and renamed `preconditionOf`.

docs/guide/src/example_usecases.md
keyboardDrummer: I'd love to improve the API, that would make this example more compelling. Do you have ideas? Without any improvements I still think this is a useful example though. It would be easy to add error messages even with this API.

examples/src/test/java/com/aws/jverify/examples/MinimumPricePercentage.java
robin-aws: AH, so it's similar to `requires f.requires` in Dafny then. That makes much more sense to me.

docs/guide/src/example_usecases.md
robin-aws: Great - we may indeed want `@Proof` later but I don't think we have a strong notion of how it differs from `@Erased` yet.

robin-aws (approved):
Great stuff, thanks for accommodating all my nit-picking!

PR #241, opened 2025-07-10
Add user profile edit test

keyboardDrummer:
### What was changed?
- Fixed a bug in the `--paths` option
- Fixed a bug in where errors on invariants were shown
- There is a remaining bug that relates to constructors, for which I've created a GH issue here: https://github.com/aws/jverify/issues/206

### How has this been tested?
- Added a test for various states of UserProfile.java
- Added a test for the `--paths` option of the CLI


test-engine/src/main/java/com/aws/jverify/testengine/TestMarkup.java
```diff
@@ -45,7 +45,18 @@ public class TestMarkup {
...
+        var outputParts = parseParts(input, positions, spans);
+
+        var output = String.join("", outputParts);
+        var hatAnnotations = findHatAnnotationSpans(String.join("", output));
+        spans.addAll(hatAnnotations);
+        
+        return output;
+    }
+
+    public static  List<String> parseParts(String input, List<Integer> positions, List<AnnotatedSpan> spans) {
```
robin-aws, line 58: (Offtopic, probably for a future PR - I realize this functionality was already here, but this is the first I'm seeing how this testing feature is used)

Could the annotations used to demarkate these "parts" be in comments as well? i.e. 

```java
private /*[>*/ @Nullable /*<]*/ PremiumFeatures premiumFeatures;  // Should be non-null for Premium accounts
```

That was the source is still valid Java, and you don't have to hide them in `resources`, they'll be easier to edit in IDEs, etc.

robin-aws (approved):

test-engine/src/main/java/com/aws/jverify/testengine/TestMarkup.java
keyboardDrummer: > Could the annotations used to demarkate these "parts" be in comments as well? i.e.

Not as they are implemented now, but we could add that feature. It's a nice idea. Probably the syntax should be slightly different then. We could expose all comments as separate annotated ranges, so you could decide to add certain comments back in.

PR #246, opened 2025-07-11
Add contracts to some String methods

olivier-aws:
### What was changed?
Adding a `StringContract` class in `builtin-contracts.java` that defines contracts for some`String` methods by changing them to static methods taking an extra parameter (`this`). Modification in the `ExpressionCompiler` to transform calls to these methods like this:
`s1.substring(s2);`
into
`String.substring(s1,s2);`

### How has this been tested?
Extended Strings.java test

### DIFF
--- a/verifier/src/main/java/com/aws/jverify/verifier/compiler/ExpressionCompiler.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/compiler/ExpressionCompiler.java
@@ -85,6 +85,20 @@ public class ExpressionCompiler {
         return new NestedMatchExpr(origin, source, translatedCases, true, null);
     }
 
+    // Given a symbol, assumed to be a MethodSymbol, returns a new MethodSymbol with
+    // same name but with a new argument type prepended.
+    static private Symbol addReceiverType(Symbol sym, com.sun.tools.javac.code.Type typ) {
+        // Create a new symbol with same name but different type (one new argument)
+        var methodType = (com.sun.tools.javac.code.Type.MethodType) sym.type;
+        var newArgTypes = methodType.argtypes.prepend(typ);
+        var newMethodType =
+                new com.sun.tools.javac.code.Type.MethodType(
+                        newArgTypes, methodType.restype, methodType.thrown, null);
+        return new Symbol.MethodSymbol((long)0, sym.name,
+                newMethodType,
+                sym.owner);
+    }
+
     public Expression toExpr(JCTree.JCExpression expr) {
         return toExpr(expr, null);
     }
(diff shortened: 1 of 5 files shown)


verifier/src/main/resources/builtin-contracts.java
```diff
@@ -193,3 +192,29 @@ public int signum() {
...
+    @Pure
+    public static boolean startsWith(String s, String prefix) {
+        postcondition((Boolean b) -> implies(prefix.length() > s.length(),b==false));
+        postcondition((Boolean b) -> implies(prefix.length() <= s.length() && forall((Integer i) -> implies(0<=i && i<prefix.length(), prefix.charAt(i) == s.charAt(i))), b==true));
+        postcondition((Boolean b) -> implies(prefix.length() <= s.length() && exists((Integer i) -> 0<=i && i<prefix.length() && prefix.charAt(i) != s.charAt(i)), b==false));
+        throw  new ContractException();
+    }
+
+    @Pure
+    public static String concat(String s1, String s2) {
```
keyboardDrummer, line 214: It should have a fresh clause on the result as well, but really we should make the String class Immutable : )

verifier/src/main/java/com/aws/jverify/verifier/compiler/ExpressionCompiler.java
```diff
@@ -201,6 +219,16 @@ public Expression toExpr(JCTree.JCExpression expr, IOrigin originOverride) {
                             yield new SeqSelectExpr(origin, false, receiver,
                                     argBindings.getFirst().getActual(), endIndex, null);
                         }
+                        case "indexOf", "startsWith","concat" -> {
+                            var strSeg = new NameSegment(origin, ownerClass.name.toString(), null);
```
keyboardDrummer, line 223: How come we need this for `String` instead of being able to use the same path as other third party code takes? Why do the StringContract methods need to be static? Would be good to add a comment in the source.

Also, I think it should be cheap to add `@Immutable` support for external classes, if you add it without adding support for class constructors, then you could use it for `String`

verifier/src/main/resources/builtin-contracts.java
olivier-aws: Given that the String type is currently translated as `seq<char>`, we cannot write `fresh(s)` with `s` a String. What would be the semantics of the Immutable annotation?

verifier/src/main/java/com/aws/jverify/verifier/compiler/ExpressionCompiler.java
olivier-aws: I'll add comments for why we handle these separately and why they are static in the builtin-contracts.java file.
All for implementing the Immutable annotation, but I want to understand its semantics first ;)

verifier/src/main/resources/builtin-contracts.java
keyboardDrummer: > Given that the String type is currently translated as seq<char>, we cannot write fresh(s) with s a String

I see

> What would be the semantics of the Immutable annotation?

`@Immutable` on an external class would mean its translated to a Dafny `datatype`. It means the programmer knows that this type can not be modified and that its equals and hash-code are structurally defined.

I'm a bit concerned that we have all this special casing code for String now that would be better to avoid.

I think `String` should be more like `BigInteger`, with a `seq<char>` as a ghost field and without special casing code for translating its type and invocations on its members. That would also slightly change the contracts you've defined here, so it seems better to do that first, before continuing

fabiomadge:

PR #247, opened 2025-07-11
feat: translate unions to trait-and-subset pairs

alex-chew:
### How has this been tested?

### DIFF
--- a/library/src/main/java/com/aws/jverify/Union.java
+++ b/library/src/main/java/com/aws/jverify/Union.java
@@ -1,0 +1,18 @@
+package com.aws.jverify;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * A sealed interface annotated with {@link Union} is treated by the verifier
+ * as having only the permitted subclasses
+ * (as opposed to having arbitrary, possibly externally-defined subclasses).
+ * <p>
+ * Note that all subclasses of a {@link Union} sealed interface must be records.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.TYPE)
+public @interface Union {
+}
--- a/verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
@@ -8,20 +8,16 @@ import com.aws.jverify.verifier.compiler.simplifications.ExternalContractCompile
 import com.aws.jverify.verifier.compiler.simplifications.LambdaCompiler;
 import com.aws.jverify.verifier.compiler.simplifications.NameCompiler;
 import com.aws.jverify.verifier.compiler.simplifications.VerifyAnnotationCompiler;
-import com.sun.source.tree.*;
-import com.sun.tools.javac.api.JavacTrees;
 import com.sun.tools.javac.code.Flags;
 import com.sun.tools.javac.code.Kinds;
 import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.Symtab;
-import com.sun.tools.javac.code.Kinds;
 import com.sun.tools.javac.code.TypeMetadata;
 import com.sun.tools.javac.code.Types;
 import com.sun.tools.javac.comp.AttrContext;
 import com.sun.tools.javac.comp.Env;
 import com.sun.tools.javac.tree.EndPosTable;
 import com.sun.tools.javac.tree.JCTree;
-import com.sun.tools.javac.comp.Enter;
 
 import com.sun.tools.javac.tree.TreeInfo;
 import com.aws.jverify.generated.*;
--- a/verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
@@ -8,7 +8,10 @@ import com.aws.jverify.verifier.compiler.simplifications.ExternalContractCompile
 import com.aws.jverify.verifier.compiler.simplifications.LambdaCompiler;
 import com.aws.jverify.verifier.compiler.simplifications.NameCompiler;
 import com.aws.jverify.verifier.compiler.simplifications.VerifyAnnotationCompiler;
+import com.sun.source.tree.*;
+import com.sun.tools.javac.api.JavacTrees;
 import com.sun.tools.javac.code.Flags;
+import com.sun.tools.javac.code.Kinds;
 import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.Symtab;
 import com.sun.tools.javac.code.Kinds;
@@ -18,6 +21,7 @@ import com.sun.tools.javac.comp.AttrContext;
 import com.sun.tools.javac.comp.Env;
 import com.sun.tools.javac.tree.EndPosTable;
 import com.sun.tools.javac.tree.JCTree;
+import com.sun.tools.javac.comp.Enter;
 
 import com.sun.tools.javac.tree.TreeInfo;
 import com.aws.jverify.generated.*;
@@ -68,7 +72,7 @@ public class JavaToDafnyCompiler {
     public static final String builtinFile = "/builtin-contracts.java";
     public static final String objectFile = "/object-contract.java";
     private boolean skipDiagnostics;
-    
+
     public JavaToDafnyCompiler(Context context, VerifierOptions verifierOptions) {
         this.context = context;
         this.verifierOptions = verifierOptions;
@@ -89,7 +93,7 @@ public class JavaToDafnyCompiler {
         }
         var contractSource = new SourceFile(JavaToDafnyCompiler.objectFile, Common.getResourceFile(getClass(), JavaToDafnyCompiler.objectFile));
         files.add(contractSource);
-        
+
         Set<JCTree.JCCompilationUnit> parsedSet = new JavaFrontEnd(this).parseResolveAndDesugarJava(options, files);
         if (parsedSet == null) {
             return new FilesContainer(List.of());
@@ -224,7 +228,7 @@ public class JavaToDafnyCompiler {
         this.skipDiagnostics = previous;
         return result;
     }
-    
+
     private void reportError(JCDiagnostic.DiagnosticPosition position, String key, Object... args) {
         if (this.skipDiagnostics) {
             return;
@@ -298,7 +302,7 @@ public class JavaToDafnyCompiler {
         return false;
     }
 
-    boolean isRecord(com.sun.tools.javac.code.Type type) {
+    public boolean isRecord(com.sun.tools.javac.code.Type type) {
         return type instanceof com.sun.tools.javac.code.Type.ClassType classType
                 && (classType.asElement().flags() & Flags.RECORD) != 0;
     }
@@ -394,7 +398,7 @@ public class JavaToDafnyCompiler {
         }
         return false;
     }
-    
+
     public @Nullable Type translateType(JCTree tree) {
         return translateType(tree.type, toOrigin(tree), null);
     }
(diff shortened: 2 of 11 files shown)


library/src/main/java/com/aws/jverify/Union.java
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * A sealed interface annotated with {@link Union} is treated by the verifier
+ * as having only the permitted subclasses
+ * (as opposed to having arbitrary, possibly externally-defined subclasses).
+ * <p>
+ * Note that all subclasses of a {@link Union} sealed interface must be records.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.TYPE)
+public @interface Union {
```
keyboardDrummer, line 17: I wonder, now that we're using the subset type translation, should we not use such a subset type for any sealed types, regardless of whether they have the `@Union` annotation?

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -514,15 +525,23 @@ public Type translateType(com.sun.tools.javac.code.Type type, IOrigin origin, JC
...
-
-
                 if (isRecord(classType) && isNullable) {
                     reportError(origin, "notSupported", "nullable record type");
                     return null;
                 }
 
+                var isUnion = classType.isInterface() && type.tsym.isSealed()
+                        && isAnnotated(classType, Union.class);
+                if (isUnion && isNullable) {
```
keyboardDrummer, line 535: What would it take to support nullable as well? Do we then need a second subset type, that also allows null?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/workaround/ClassesExtendingClassesCompiler.java
```diff
@@ -131,4 +137,51 @@ private Method constructorToInitMethod(String className, Constructor constructor
                 modClause,
                 false, List.of(), body, false);
     }
+
+    private @Nullable SubsetTypeDecl buildUnionSubsetType(
```
keyboardDrummer: Can we put this in a different file, and call it in the same place as we're calling this file? Something like the following:

Before:
```java

            if (intermediateResult != null) {
                result.addAll(extendingClassesCompiler.compile(intermediateResult, classSymbol));
            }
```
After:
```java
            if (intermediateResult != null) {
                for(var postUnionDecl : unionCompiler.compile(intermediateResult, classSymbol)) {
                  result.addAll(classDeclCompiler.compile(postUnionDecl, classSymbol));
                }
            }
```

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/workaround/ClassesExtendingClassesCompiler.java
```diff
@@ -131,4 +137,51 @@ private Method constructorToInitMethod(String className, Constructor constructor
...
+            Symbol.ClassSymbol classSymbol, IOrigin origin, Name name,
+            List<TypeParameter> typeParameters, Type traitType) {
+        if (!isSealedInterface(classSymbol)) {
+            classCompiler.compiler.reportError(origin, "unionMustBeSealedInterface");
+            return null;
+        }
+
+        // Any @Contract class must implement the interface and so must appear in the permits clause,
```
keyboardDrummer: A `@Contract` class for an interface does not have to implement the interface. JVerify has custom code to check that the methods defined in the contract class match methods in the contracted type. Having the contract class implement an interface is only to let the IDE help the programmer figure out what the methods to define are.

I think for sealed types with a contract, it is better not to let the contract class implement the sealed type.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/workaround/ClassesExtendingClassesCompiler.java
```diff
@@ -131,4 +137,51 @@ private Method constructorToInitMethod(String className, Constructor constructor
...
+        // but shouldn't be translated to a datatype constructor.
+        var variantTypes = classSymbol.getPermittedSubclasses()
+                .stream()
+                .filter(type -> !classCompiler.compiler.isAnnotated(type, Contract.class))
+                .toList();
+        if (!variantTypes.stream().allMatch(classCompiler.compiler::isRecord)) {
```
keyboardDrummer: Now that we are not translating to a datatype, is it still a requirement that the permits clause only contains records?

verifier/src/test/java/com/aws/jverify/verifier/tests/Unions.java
```diff
@@ -0,0 +1,90 @@
...
+            return s.length() % 2 == 0;
+        }
+    }
+}
+
+@SuppressWarnings("unused")
+@Union
+sealed interface Either<T, U> {
+    @Pure
+    boolean bifoldMap(Predicate<T> applyLeft, Predicate<U> applyRight);
```
keyboardDrummer: How did you come up with the name `bifoldMap`? I would call this method `test` or `contains`

verifier/src/test/java/com/aws/jverify/verifier/tests/Unions.java
```diff
@@ -0,0 +1,90 @@
...
+
+import com.aws.jverify.Contract;
+import com.aws.jverify.ContractException;
+import com.aws.jverify.Pure;
+import com.aws.jverify.Union;
+import com.aws.jverify.testengine.JVerifyTest;
+
+import static com.aws.jverify.JVerify.*;
+
+@JVerifyTest(exitCode = 0, dafnyVerified = 9, dafnyErrors = 0, useBuiltinContracts = true)
```
keyboardDrummer, line 11: What do you need the builtins for? If you don't really need them then I would avoid using them.

verifier/src/test/java/com/aws/jverify/verifier/tests/Unions.java
```diff
@@ -0,0 +1,90 @@
...
+class Unions {
```
keyboardDrummer, line 12: You're not using the subset class in this test file right?

Shouldn't you add a test like:

```java
<L, R> void useEither(Either<L, R> either) {
  if (either instanceof Left<L> left) {
     ...
  } else if (either instanceof Right<R> right) {
     ...
  } else {
     impossible(); // JVerify method that translates to assert(false), can also be used to produce values of any type
  }
}
```

Or ideally something with patterns:
```java
<L, R> void useEither(Either<L, R> either) {
  switch(either) {
    case Left<L> left -> ...
    case Right<R> right -> ...
  }
}
```

library/src/main/java/com/aws/jverify/Union.java
alex-chew: The `@Union` translation currently has the restriction that all implementing interfaces must be records, whereas in general, sealed interfaces can be implemented by non-record classes and even non-final classes. It might be worthwhile to support the more general cases in the future, but in the meanwhile I think it's better for the specialized translation to be opt-in so that we don't prevent users from using general sealed interfaces.

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
alex-chew: The union's base trait isn't a reference trait, so it doesn't seem feasible to just define a second subset type that also allows null. My current thinking is that we should have a general approach to support all of the Java reference types that are translated to Dafny value types (nullable records/strings/etc.), e.g. a `datatype Nullable<T>` in the prelude. But I think that should be in a separate PR.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/workaround/ClassesExtendingClassesCompiler.java
alex-chew: Ah, thanks for the correction - I thought that our contract-matching mechanism didn't work correctly with type parameters (and incorrectly generalized the limitation here). It turns out that the root cause of the test failure ("method ... without a body still needs a contract") was that I'd added `@Pure` to both an interface method and the method's implementations, and removing `@Pure` from the interface method resolved the issue. I'll add some logic to clarify the user error in this case.

I agree that it's simpler in both implementation and usage if we don't allow a sealed type's contract class to implement the sealed type itself, so I'll go ahead and implement that. On the other hand, I expect that the UX will be a bit worse without the help from the IDE in matching signatures.

alex-chew: Hmm, upon closer inspection I found an unintended effect of making the contract class (`EitherContract` in the Unions tests) not implement the interface: now the base `trait Either<T, U>` has no members (and instead the implementations are only defined in the implementing `datatype`s). It's still possible to call the methods on values of the union's subtypes, but not of the union type itself, so I'll need to look into this more.

verifier/src/test/java/com/aws/jverify/verifier/tests/Unions.java
alex-chew: Yeah, `test` makes more sense now. (Originally the parameters had types `Function<T, V>` and `Function<U, V>` and so I took the name from Haskell's `Data.Bifoldable`, though I now realize that didn't quite match either.)

verifier/src/test/java/com/aws/jverify/verifier/tests/Unions.java
alex-chew: The builtins are included for `java.lang.Integer`'s contracts, as without them I get:

> Error: Type or type parameter is not declared in this scope: java_lang_Integer (did you forget to qualify a name or declare a module import 'opened'?)

Is there a better way to ensure that `Integer` is in scope?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/workaround/ClassesExtendingClassesCompiler.java
alex-chew: It's not a requirement, strictly speaking, but I don't see practical value in relaxing this restriction.

I originally thought that a Dafny trait could not have both reference-typed and value-typed implementers, and so in order to allow ergonomic usage of `@Union`s in expression contexts, I opted to only allow record subtypes. It turns out that a Dafny trait can indeed have both reference- and value-typed implementers, so in theory we could allow non-record subtypes to implement `@Union` interfaces. But since a `@Union` interface's base trait doesn't have `extends object`, it's invalid to add a `reads this` clause on the trait methods, and in turn the implementations can't have `reads this` - which seems to prevent doing anything interesting (unless I'm missing something).

library/src/main/java/com/aws/jverify/Union.java
keyboardDrummer: What prevents enabling `@Union` for non-record types? Also, how about automatically applying union to sealed record types?

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
keyboardDrummer: Can this condition ever trigger? You are already checking and returning on `if (isRecord(classType) && isNullable)` above.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/workaround/ClassesExtendingClassesCompiler.java
keyboardDrummer: So you're saying there's a bug in `@Contract` ? Sounds like `trait Either` should have the member that was defined in the `EitherContract` and the `Either` interface

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/workaround/ClassesExtendingClassesCompiler.java
keyboardDrummer: > But since a @Union interface's base trait doesn't have extends object

Could we not allow both `@Modifiable` and `@Union` on a sealed interface? It's no longer clear to me how `@Union` has an interaction with value and reference types. A type test in Dafny can be done on both value and reference traits. Example of how we could translate things. `SomeC` and `SomeD` are the sealed union interfaces in this example.

image

verifier/src/test/java/com/aws/jverify/verifier/tests/Unions.java
keyboardDrummer: No there's no better way :-) besides note using Integer but it seems useful here.

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
alex-chew: Yes, the condition can trigger (see the `NullableUnion` test in `UnionsErrors.java`). `@Union` is applied to the sealed interface and not to its subclasses, so the `isRecord(classType) && isNullable` condition doesn't catch that case.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/workaround/ClassesExtendingClassesCompiler.java
alex-chew: Yes, I think so. I'll spend some time investigating why it doesn't work correctly, but I'd prefer not to block this PR on it.

keyboardDrummer: Let's file a GH issue for the bug and leave it for another PR. @olivier-aws also encountered it

fabiomadge:

PR #250, opened 2025-07-16
Resolve printed dafny

keyboardDrummer:
### What was changed?
- Various fixes that enable parsing the generated Dafny for more Java programs:
  - Support usage of some reserved Dafny keywords
  - Support wildcards in parameters of overloaded methods
  - Support generated jump labels
  - Support variables named `_`
  - Support variables whose name starts with `_`
  - Support more valid Java names for types and parameters
- Turn off `useBuiltinContracts` for tests by default, to:
  - improve developer speed when using step-through debugging, since there is less unimportant code to compile
  - make tests less dependent on both the standard library and what they actually want to test
  - improve test performance
- Rename test `Constructors` to `OverloadsTest`
- Changed the option `resolvePrintedDafny` so it can be used for tests that have a non-zero exit code. It will then only test that the exit code remains the same
- Changed the exit code, so a failure on the JVerify side returns 2, and a non-zero Dafny exit code is prefixed with "2".

### How has this been tested?
- Turn on `resolvePrintedDafny` for more tests
- Add a test `ResolvePrintedDafny` that test for Dafny specific issues with generated names
- Add cases to `OverloadsTest` to ensure that when there are overloads with wildcards, the printed Dafny can be resolved

### DIFF
--- a/test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
+++ b/test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
@@ -3,12 +3,7 @@ package com.aws.jverify.testengine;
 import com.aws.jverify.common.AnnotatedRange;
 import com.aws.jverify.common.Position;
 import com.aws.jverify.common.Range;
-import com.aws.jverify.verifier.DafnyDiagnostic;
-import com.aws.jverify.verifier.Driver;
-import com.aws.jverify.verifier.SourceFile;
-import com.aws.jverify.verifier.VerifierOptions;
-import com.aws.jverify.verifier.compiler.JavaFrontEnd;
-import com.aws.jverify.verifier.compiler.JavaToDafnyCompiler;
+import com.aws.jverify.verifier.*;
 import com.google.auto.service.AutoService;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.junit.jupiter.api.Assertions;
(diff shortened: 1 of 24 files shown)


test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -208,27 +201,30 @@ public static void verifyFile(SourceFile sourceFile, JVerifyTest annotation, Lis
         );
     }
 
-    private static void resolvePrintedDafny(VerifierOptions verifierOptions) 
+    private static void resolvePrintedDafny(VerificationResults previousResults, VerifierOptions verifierOptions)
```
olivier-aws: The method is called `resolvePrintedDafny` but calls Dafny with `verify` mode. This is a little weird.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/NameCompiler.java
```diff
@@ -135,6 +151,16 @@ private String getMethodName(Symbol.MethodSymbol s) {
         return result.toString();
     }
 
+    private String encodeName(String name) {
+        if (reservedDafnyNames.contains(name)) {
+            name = "r_" + name;
+        }
+        if (name.startsWith("_")) {
+            name = "a" + name;
+        }
```
olivier-aws: Since this is called for local variables as well, can't you have a name clash between a variable named `_f` and a variable named `a_f`?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/NameCompiler.java
```diff
@@ -135,6 +151,16 @@ private String getMethodName(Symbol.MethodSymbol s) {
         return result.toString();
     }
 
+    private String encodeName(String name) {
+        if (reservedDafnyNames.contains(name)) {
+            name = "r_" + name;
+        }
```
olivier-aws: Can't this make a name clash between a variable named `set` and another named `r_set`?

olivier-aws (commented):
Some concerns about potential name clashes

PR #251, opened 2025-07-16
Support lambdas on polymorphic interfaces and in polymorphic types/methods

keyboardDrummer:
Depends on https://github.com/aws/jverify/pull/254

### What was changed?
- Support various combination of lambdas and polymorphism
- Slightly improved error locations for lambdas

### How has this been tested?
- Update expected errors in `Lambdas` test
- Added more cases to the `Lambdas` test
- Add test `PolymorphicLambdas`


verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -564,12 +555,23 @@ public Name getName(JCTree tree, String name) {
     }
 
     public Name getName(JCTree tree, String name, int length) {
+        IOrigin origin;
+        if (name.startsWith("Lambdas$")) {
```
keyboardDrummer, line 587: Is this still needed?

verifier/src/main/java/com/aws/jverify/verifier/compiler/PositionCalculator.java
```diff
@@ -23,6 +24,9 @@ public int getStartPos(JCTree tree) {
     }
 
     private int getMethodNamePosition(JCTree.JCMethodDecl methodDecl) {
+        if ((methodDecl.mods.flags & Flags.SYNTHETIC) != 0) {
```
keyboardDrummer, line 27: There should be a mapping from the synthetic JCTree to the original one, so this is not necessary (and work better)

verifier/src/main/java/com/aws/jverify/verifier/compiler/PositionCalculator.java
```diff
@@ -89,6 +93,10 @@ public int getEndPos(JCTree node) {
     }
 
     private int getClassNamePosition(JCTree.JCClassDecl classDecl) {
+        if ((classDecl.mods.flags & Flags.SYNTHETIC) != 0) {
```
keyboardDrummer, line 96: There should be a mapping from the synthetic JCTree to the original one, so this is not necessary (and work better)

PR #254, opened 2025-07-17
Introduction of RefObject and ValueObject

keyboardDrummer:
#### Not user facing
- Add ability to change working directory through `user.dir` property.
- Fix a bug that could cause (generated) classes not to be verified, by accidentally considering them part of the built-in file.
- Remove obsolete use of the `{:verify false}` attribute

#### User facing
- Support assigning any type parameter to `@Immutable Object`. `@Immutable Object` does not support `==`
- Support assigning a record to `@Immutable Object` and to `Record`.
- Because `Object` is now a type that's referenced during compilation, its contract has been moved to a separate contract file, so it's always included even when `useBuiltinContracts` is false.

### Caveat
- We need to add a section to the user documentation about immutable and mutable types, but I prefer to write that until we finish the related features.
- I'm happy to change the default so that `@Immutable Object` -> `Object` and `Object` -> `@Modifiable Object`, but let's do that in a separate PR so the effect of that change is crystal clear.

### How has this been tested?
- Add test-case to `PolymorphismWithoutBounds`
- Added a test `refObjectIsNotTop` and `valueObjectIsTop`
- Added tests `valueObjectHasNoEquality` and `valueObjectHasNoEqualityBoth`
- Added a test `assignRecordToRefObject` and `assignRecordToValueObject`


verifier/src/main/java/com/aws/jverify/verifier/compiler/ClassCompiler.java
```diff
@@ -192,7 +189,14 @@ public List<TypeParameter> translateTypeParameters(List<JCTree.JCTypeParameter>
         return typarams.stream().map(p -> {
             var name = compiler.getName(p, p.getName());
             var bounds = p.bounds.map(compiler::translateType);
-            return new TypeParameter(compiler.toOrigin(p),
+            
+            IOrigin origin = compiler.toOrigin(p);
+            if (!this.compiler.verifierOptions.includeBuiltinContracts() &&
+                    !this.compiler.compilationUnit.getSourceFile().getName().contains("builtin-contracts.java")) {
```
alex-chew: nit: The path should reference the corresponding constant in JavaToDafnyCompiler

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -86,6 +88,8 @@ public NameCompiler getNameCompiler() {
             builtinSource = new SourceFile("builtin-contracts.java", Common.getResourceFile(getClass(), JavaToDafnyCompiler.builtinFile));
             files.add(builtinSource);
         }
+        var contractSource = new SourceFile("object-contract.java", Common.getResourceFile(getClass(), JavaToDafnyCompiler.objectFile));
```
alex-chew: nit: May as well reference the corresponding constant instead of repeating it here

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -301,11 +300,7 @@ public NotImplementedException(String message) {
     }
 
     public Attributes getAttributes(IOrigin origin) {
```
alex-chew: nit: Looks like this should be removed entirely?

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -492,6 +487,14 @@ public Type translateType(JCTree.JCModifiers modifiers, com.sun.tools.javac.code
                 return new UserDefinedType(origin, new NameSegment(origin, "array" + nullableSuffix, List.of(elemType)));
             }
             case com.sun.tools.javac.code.Type.ClassType classType -> {
+                if (isAnnotated(modifiers, com.aws.jverify.Immutable.class)) {
+                    Symtab symtab = Symtab.instance(context);
+                    if (classType.tsym == symtab.objectType.tsym) {
+                        return new UserDefinedType(origin, new NameSegment(origin, "ValueObject", null));
+                    } else {
+                        reportError(origin, "notSupported", "@Immutable on another type reference than Object");
```
alex-chew: nit:
```suggestion
                        reportError(origin, "notSupported", "@Immutable on a reference type other than Object");
```

verifier/src/main/resources/object-contract.java
```diff
@@ -0,0 +1,25 @@
+package com.aws.jverify.builtin;
+
+import com.aws.jverify.*;
```
alex-chew: nit: Most of these are unused?

verifier/src/test/java/com/aws/jverify/verifier/tests/PolymorphismDafnyResolutionErrors.java
```diff
@@ -0,0 +1,12 @@
+package com.aws.jverify.verifier.tests;
+
+import com.aws.jverify.testengine.JVerifyTest;
+
+@JVerifyTest(exitCode = 22)
```
alex-chew, line 5: What's exit code 22?

verifier/src/test/java/com/aws/jverify/verifier/tests/PolymorphismDafnyResolutionErrors2.java
```diff
@@ -0,0 +1,16 @@
...
+
+@JVerifyTest(exitCode = 22)
+public class PolymorphismDafnyResolutionErrors2 {
+    
+    public static void valueObjectHasNoEquality() {
+        // Annotations on method parameters are broken ATM, so these are locals
+        @Immutable Object a = new Object();
+        Object b = new Object();
+        var c = a == b;
+//              ^ Error: == can only be applied to expressions of types that support equality (got ValueObject)
```
alex-chew, line 14: Can we make this error message say "types that support reference equality" instead of just "types that support equality"? I think that would be easier for typical users to understand.

library/src/main/java/com/aws/jverify/Immutable.java
```diff
@@ -0,0 +1,16 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Can be used on a variable of type Object, 
+ * to allow assigning values with a record type, or a generic type, to this variable.
+ * However, the `==` operator can not be used on immutable objects.
+ */
+@Retention(RetentionPolicy.SOURCE)
+@Target(ElementType.TYPE_USE)
+public @interface Immutable {
```
alex-chew, line 15: What do you think about renaming this to `ValueObject` to match the Dafny trait? I think it would make the error messages in the tests below more clear.

verifier/src/test/java/com/aws/jverify/verifier/tests/PolymorphismDafnyResolutionErrors3.java
```diff
@@ -0,0 +1,13 @@
...
+
+import com.aws.jverify.Immutable;
+import com.aws.jverify.testengine.JVerifyTest;
+
+@JVerifyTest(exitCode = 2)
+public class PolymorphismDafnyResolutionErrors3 {
+    
+    public static void valueObjectHasNoEqualityBoth(@Immutable Object a, @Immutable Object b) {
+        var c = a == b;
+//                ^ error: '==' is not allowed when both operand's types could be String, a record, or a boxed primitive, unless one of the operands is a null literal
```
alex-chew, line 11: I think it would be helpful to define the term "value object" (which would include `String`, records, boxed primitives, and `@Immutable Object`) in our documentation, and use that term consistently in our error messages and such.

verifier/src/test/java/com/aws/jverify/verifier/tests/PolymorphismDafnyResolutionErrors.java
```diff
@@ -0,0 +1,12 @@
...
+public class PolymorphismDafnyResolutionErrors {
+    
+    public static <T> void refObjectIsNotTop(T value) {
+        Object o = value;
+//      ^^^^^^^^^^^^^^^^^ Error: RHS (of type T) not assignable to LHS (of type Object)
```
alex-chew, line 10: I expect that this error will confuse users, since it doesn't explain that the assignment is forbidden because `T` may be a immutable / value object type.

alex-chew (dismissed):
The implementation looks good in terms of soundness, but I'm worried that the user experience will be confusing. I think it'd make a big difference to have the relevant error messages mention immutable / value object types specifically, so users can search our docs for those terms. (I think it'd be fine to make any changes like that in a follow-up PR, if you choose to.)

verifier/src/test/java/com/aws/jverify/verifier/tests/PolymorphismDafnyResolutionErrors.java
keyboardDrummer: `2x` means that Dafny returned exit code `x`

library/src/main/java/com/aws/jverify/Immutable.java
keyboardDrummer: I think we should either consistently use the terms modifiable and immutable, OR reference and value. With the existence of the `@Modifiable` annotation, I think we're currently using the first two, so let's stick with that for now.

We could rename `Object` to `ModifiableObject` and `ValueObject` to `Object`.

keyboardDrummer:
> The implementation looks good in terms of soundness, but I'm worried that the user experience will be confusing. I think it'd make a big difference to have the relevant error messages mention immutable / value object types specifically, so users can search our docs for those terms. (I think it'd be fine to make any changes like that in a follow-up PR, if you choose to.)

It's really good feedback but we don't have control over the error messages from Dafny atm. First we need to make a change change that couples each one to an error key, so that we can then override them in JVerify.

I'm guessing at some point we should go over all the error messages in our tests and figure out how what they should be and how to get there. It's a bit of functional debt we're incurring by not doing that right away. Do you think it's the wrong order?

alex-chew:
> It's really good feedback but we don't have control over the error messages from Dafny atm. First we need to make a change change that couples each one to an error key, so that we can then override them in JVerify.
>
> I'm guessing at some point we should go over all the error messages in our tests and figure out how what they should be and how to get there. It's a bit of functional debt we're incurring by not doing that right away. Do you think it's the wrong order?

I agree that it'd be a good idea to have Dafny emit more structured errors so that we can re-interpret them as necessary on the JVerify side, and that it's not worthwhile to take that on right away. I wonder if we can use the existing `{:error}` attribute to somewhat improve the error messages, without needing major changes to Dafny or to our translations?

keyboardDrummer:
> I agree that it'd be a good idea to have Dafny emit more structured errors so that we can re-interpret them as necessary on the JVerify side, and that it's not worthwhile to take that on right away. I wonder if we can use the existing `{:error}` attribute to somewhat improve the error messages, without needing major changes to Dafny or to our translations?

Note that Dafny already supports structured errors. The file `Source/DafnyCore/assets/messages.txt` can be used to declare error message keys and the Dafny message for those. The remaining work is to go over all the reported errors and move those into `messages.txt`. JVerify can then override the messages for these keys.

PR #256, opened 2025-07-17
Support unverified record constructors

### LINKED ISSUE #165: Support reading final fields in specifications
There is a snag with the way we write specifications in JVerify:

```java
class Box {
    private final int value;

    public Box(int value_) {
        postcondition(this.value == value_);  // Java error: Variable 'this.value' might not have been initialized
        this.value = value_;
    }
}
```

Java complains that `this.value` isn't set yet if the `postcondition` call is before it is actually initialized, but JVerify will not allow you to put it after because all specification calls have to come before implementation code.

keyboardDrummer:
Fixes https://github.com/aws/jverify/issues/165

### What was changed?
- Moved contract related code from BlockCompiler to ContractCompiler
- Also allow constructor contracts to be specified after the body, to support postcondition contracts on final fields
- Support record constructors, but only if they are not unverified
- Support constructing generic records when the type parameters can not be inferred from surrounding code

### How has this been tested?
- Add a test file `ConstructorsVerified` and `ConstructorErrors`
- Added new cases to the test `Records`


verifier/src/main/java/com/aws/jverify/verifier/compiler/ContractCompiler.java
```diff
@@ -0,0 +1,190 @@
...
+
+    public List<JCTree.JCStatement> translateHeader(JCTree.JCStatement statement, 
+                                                    MethodOrLoopContract contract,
+                                                    boolean allowFooterButNotHeader,
+                                                    boolean reportErrors) {
+        var statements = statement instanceof JCTree.JCBlock block
+                ? block.getStatements()
+                : List.of(statement);
+        return translateHeader(statements, contract, allowFooterButNotHeader, reportErrors);
+    }
```
alex-chew: Is there a significance to the parameter `allowFooterButNotHeader` being named differently from `allowFooter` in the other `translateHeader` signature?

verifier/src/main/java/com/aws/jverify/verifier/compiler/ContractCompiler.java
```diff
@@ -0,0 +1,190 @@
...
+        int headerContracts = addHeaderContracts(statements, contract, reportErrors, startIndex);
+        int footerContracts = addFooterContracts(statements, contract, allowFooter, reportErrors, headerContracts);
+        ArrayList<JCTree.JCStatement> result = new ArrayList<>(statements.subList(headerContracts, footerContracts));
+        if (superOrThis != null) {
+            result.addFirst(superOrThis);
+        }
+        return result;
+    }
+
+    private int addFooterContracts(List<JCTree.JCStatement> statements, MethodOrLoopContract contract, boolean allowFooter, boolean reportErrors, int headerContracts) {
```
alex-chew: It would be nice to add doc comments to these methods (this one and `addHeaderContracts`), in particular since it's not immediately obvious what the return value is.

Also, it looks like all kinds of contract statements are allowed to appear in the footer - do you think it's worth restricting them to only `postcondition`?

verifier/src/main/java/com/aws/jverify/verifier/compiler/ContractCompiler.java
```diff
@@ -0,0 +1,190 @@
...
+
+        JCTree.JCStatement superOrThis = null;
+        var first = statements.isEmpty() ? null : statements.getFirst();
+        if ((first instanceof JCTree.JCExpressionStatement expressionStatement
+                && expressionStatement.getExpression() instanceof JCTree.JCMethodInvocation invocation)) {
+            var isSuperOrThisCall = invocation.getMethodSelect() instanceof JCTree.JCIdent ident && 
+                    (ident.name == ident.name.table.names._super || ident.name == ident.name.table.names._this);
+            if (isSuperOrThisCall) {
+                superOrThis = first;
+            }
```
alex-chew, line 61: nit: There's a `TreeInfo` utility method that does the same thing -
```suggestion
            if (TreeInfo.isConstructorCall(invocation)) {
                superOrThis = first;
            }
```

verifier/src/main/java/com/aws/jverify/verifier/compiler/ContractCompiler.java
```diff
@@ -0,0 +1,190 @@
...
+import com.sun.tools.javac.tree.JCTree;
+import com.sun.tools.javac.tree.TreeInfo;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/*
+Extracts contracts from constructor, method or loop bodies
+ */
+public class ContractCompiler {
```
alex-chew, line 14: Now that postconditions (and perhaps other contract statements) are allowed to appear at the end of a block, do we need to take care to disallow referencing local variables that are defined only within the block (and thus wouldn't be in scope in the translated `ensures` clause)?

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -197,7 +199,18 @@ public void reportError(JCTree tree, String key, Object... args) {
         reportError(positionFromNode(tree, compilationUnit), key, args);
     }
     
+    public <T> T withSkipDiagnostics(Supplier<T> supplier, boolean skipDiagnostics) {
```
alex-chew, line 206: Could you explain in what cases it would be necessary to skip diagnostics?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/RecordCompiler.java
```diff
@@ -71,8 +79,22 @@ public IndDatatypeDecl translateRecord(JCTree.JCClassDecl classDecl, IOrigin ori
...
                 // and the implicit canonical constructor is unneeded to construct datatype values.
                 if (TreeInfo.isConstructor(methodDecl)) {
-                    if (!isSyntheticCanonicalConstructor(methodDecl)) {
-                        compiler.reportError(member, "notSupported", "explicit record constructor");
+                    String resultName = "resultName";
+                    NameSegment resultReference = new NameSegment(origin, resultName, null);
+                    compiler.expressionCompiler.handleThis = (thisExpr, innerOrigin) -> resultReference;
+                    boolean isImplicitCanonicalConstructor = isImplicitCanonicalConstructor(methodDecl);
+                    var dafnyMember = compiler.withSkipDiagnostics(() -> classCompiler.translateMember(member), isImplicitCanonicalConstructor);
+                    compiler.expressionCompiler.handleThis = null;
```
alex-chew: This approach of writing to the public `handleThis` member seems easy to misuse, particularly in case it's used in more places in the future. Would it be feasible to implement that using the same technique as `withSkipDiagnostics`?

verifier/src/main/java/com/aws/jverify/verifier/compiler/ContractCompiler.java
keyboardDrummer: I don't see `isConstructorCall`. I see `isConstructor` but that's not what I need. There's `isSuperCall` which is good but I don't see `isThisCall`

verifier/src/main/java/com/aws/jverify/verifier/compiler/ContractCompiler.java
keyboardDrummer: > It would be nice to add doc comments to these methods (this one and addHeaderContracts), in particular since it's not immediately obvious what the return value is.

Refactored the code so I hope the comment in unnecessary

> Also, it looks like all kinds of contract statements are allowed to appear in the footer - do you think it's worth restricting them to only postcondition?

Would that improve the UX? I think unless there's a benefit, let's not add an extra rule.

verifier/src/main/java/com/aws/jverify/verifier/compiler/ContractCompiler.java
keyboardDrummer: > Now that postconditions (and perhaps other contract statements) are allowed to appear at the end of a block, do we need to take care to disallow referencing local variables that are defined only within the block (and thus wouldn't be in scope in the translated ensures clause)?

That would result in a Dafny error. Question is whether that error can be made to look good, which I don't know. I'm currently leaving the polishing of errors to a later time, since IMO right now we're best served with demonstrating that we can verify interesting Java programs.

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
keyboardDrummer: Good question, added comments

PR #259, opened 2025-07-18
Distinguish primitive types from boxed types

### LINKED ISSUE #233: Distinguish between primitive types and boxed types
The current translation maps both `int` and `Integer` to a Dafny `int32` type. This becomes unsound once we allow using `==` and `equals`, since `1000 == 1000` but `Integer.valueOf(1000) != Integer.valueOf(1000)`, and `Integer.valueOf(1000).equals(Integer.valueOf(1000))` but `1000.equals(1000)` isn't even valid Java.

robin-aws:
### What was changed?
Resolves #233.

No longer maps both `Integer` and `int` to the same Dafny type.

Adds the "LOWER" javac phase to our subset of the javac pipline we run before converting to Dafny. Amongst other features such as nested classes, this handles boxing and unboxing for us.

Also:

* Added many overloads of `JVerify` methods for primitive types. There is plenty of precedent for this approach in the JRE itself, in that specialized types such as `intPredicate` already exist. I only added enough to maintain the current set of tests and examples, but we should add all possible overloads at some point, potentially with some javac AST-driven automation.
* ~Since LOWER outputs loose class definitions instead of rewriting compilation units in place, added `JVerifyCompilationUnit` as a stand-in for JCCompilationUnits to recreate a similar shape of output.~
* Since LOWER rewrites some things we don't want lowered, such as records and switch statements/expressions, added pre- and post-processing in `Suspenders` to remove them from the rewriting scope and then restore them after.
* Since LOWER clears state it consumes from helper classes like `Enter` and `JavacTrees`, created a lightweight replacement for some of that indexing in `JVerifyIndex`.

### How has this been tested?

Existing tests: many had to be adjusted to handle primitive types correctly, a few had existing expected errors change.


verifier/src/main/java/com/aws/jverify/verifier/compiler/JVerifyCompilationUnit.java
```diff
@@ -0,0 +1,37 @@
...
+import com.sun.tools.javac.util.List;
+
+import static com.sun.tools.javac.tree.JCTree.Tag.IMPORT;
+import static com.sun.tools.javac.tree.JCTree.Tag.MODULEDEF;
+import static com.sun.tools.javac.tree.JCTree.Tag.MODULEIMPORT;
+import static com.sun.tools.javac.tree.JCTree.Tag.PACKAGEDEF;
+
+/**
+ * A workaround for the fact that we can't create new JCCompilationUnit instances,
+ * but the LOWER phase
```
keyboardDrummer: This sentence seems to be...

verifier/src/main/java/com/aws/jverify/verifier/compiler/JVerifyCompilationUnit.java
```diff
@@ -0,0 +1,37 @@
...
+import com.sun.tools.javac.tree.JCTree;
+import com.sun.tools.javac.util.List;
+
+import static com.sun.tools.javac.tree.JCTree.Tag.IMPORT;
+import static com.sun.tools.javac.tree.JCTree.Tag.MODULEDEF;
+import static com.sun.tools.javac.tree.JCTree.Tag.MODULEIMPORT;
+import static com.sun.tools.javac.tree.JCTree.Tag.PACKAGEDEF;
+
+/**
+ * A workaround for the fact that we can't create new JCCompilationUnit instances,
```
keyboardDrummer: You can assign to `JCCompilationUnit.defs` though. Could we not use that instead of this wrapper?

verifier/src/main/java/com/aws/jverify/verifier/compiler/JVerifyIndex.java
```diff
@@ -0,0 +1,115 @@
...
+import com.sun.tools.javac.util.Pair;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A replacement for some of the lookup functionality of
+ * indexes like JavacTrees and Elements.
+ */
+public class JVerifyIndex extends JCTree.Visitor {
```
keyboardDrummer, line 32: Can the comment explain why `JavacTrees` and `Elements` are not working?

I'm guessing we could fix the existing Javac index classes so we can use those instead, but we can do that in a follow-up PR.

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -584,6 +585,7 @@ public Name getName(JCTree tree, String name) {
         return getName(tree, name, name.length());
     }
 
+    // TODO: Should be calling toOrigin() instead of duplicating the logic
```
keyboardDrummer: TODO

verifier/src/test/java/com/aws/jverify/verifier/tests/TranslationErrors.java
```diff
@@ -54,14 +54,18 @@ void wrongContractMethod(int x) {
     int switchLabeledStatementGroup(int i) {
         var acc = 0;
         var ret = switch (i) {
-//                ^ error: switch labeled statement group is not supported
             case 0:
+//          ^ error: switch labeled statement group is not supported
                 acc += 100;
                 yield 0;
             case 1, 2, 3:
+//          ^ error: switch labeled statement group is not supported
```
keyboardDrummer, line 62: How come the change in behavior here?

verifier/src/test/java/com/aws/jverify/verifier/tests/ResolutionErrorsTypes.java
```diff
@@ -5,8 +5,8 @@
 
 import static com.aws.jverify.JVerify.check;
 
-@JVerifyTest(exitCode = 22)
-class ResolutionErrorsTypes {
+@JVerifyTest(exitCode = 4, dafnyVerified = 1, dafnyErrors = 1, useBuiltinContracts = true)
+class ResolutionErrorsTypes { // TODO test name is incorrect
```
keyboardDrummer: Fix test name

verifier/src/main/resources/builtin-contracts.java
```diff
@@ -50,6 +51,16 @@ class ShortContract {
 class IntegerContract {
     public static final int MAX_VALUE = 0x7fffffff;
     public static final int MIN_VALUE = 0x80000000;
+
+    public static Integer valueOf(int i) {
```
keyboardDrummer, line 55: Should we add a test for boxed primitives?

verifier/src/main/java/com/aws/jverify/verifier/compiler/Suspenders.java
```diff
@@ -0,0 +1,408 @@
...
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Stream;
+
+import static com.sun.tools.javac.code.Flags.RECORD;
+
+
+/**
+ * A reversible rewriting phase that prevents features like
```
keyboardDrummer: Could you say something about why we need to do the translation of switches to a construct that is semantically equivalent, if we're translating it back anyways? I guess the equivalence helps ensure that the lower pass transforms the code as desired.

verifier/src/main/java/com/aws/jverify/verifier/compiler/Suspenders.java
```diff
@@ -0,0 +1,408 @@
...
+            }
+        }
+
+        super.visitSwitchExpression(tree);
+    }
+
+    /**
+     * Translates a switch expression into an equivalent chained conditional expression.
+     * See also switchExpressionFromConditional for the reverse transformation used
+     * in the UNSUSPECT direction.
```
keyboardDrummer: SUSPECT!!

keyboardDrummer (dismissed):
Amazing :) left some comments. I'll let you decide when to address them

verifier/src/main/java/com/aws/jverify/verifier/compiler/JVerifyCompilationUnit.java
robin-aws: I had the realization that this was all unnecessary because I could just set `unit.defs` directly, then didn't come back to finish it. :) Thanks for pushing that change!

verifier/src/main/java/com/aws/jverify/verifier/compiler/JVerifyCompilationUnit.java
robin-aws: See above - d'oh!

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
robin-aws: Removed as there's non-trivial differences to sort out around ending positions. I think we'll naturally tackle this when there's a concrete need.

verifier/src/test/java/com/aws/jverify/verifier/tests/TranslationErrors.java
robin-aws: See the comments on `Suspenders.swtichToIf` and `Suspenders.switchExpressionToConditional`: they are partial mappings that only cover all types we do support. For cases like this we don't, Suspenders doesn't hide them from LOWER, so cases with multiple labels get split up, and hence you get what looks like duplicated errors.

I don't love this but didn't want to improve it as part of this PR. The best way is probably to make it so that our own early javac-style phases can report errors as well (that aren't flagged as "Java compilation errors"). Cut https://github.com/aws/jverify/issues/266 for tracking.

verifier/src/test/java/com/aws/jverify/verifier/tests/ResolutionErrorsTypes.java
robin-aws: -> `PrimitiveAndBoxedTypes`

verifier/src/main/resources/builtin-contracts.java
robin-aws: Yes, will follow up with more content in the renamed `PrimitiveAndBoxedTypes` test.

verifier/src/main/java/com/aws/jverify/verifier/compiler/Suspenders.java
robin-aws: 🤨

verifier/src/main/java/com/aws/jverify/verifier/compiler/Suspenders.java
```diff
@@ -0,0 +1,417 @@
...
+            }
+        }
+
+        super.visitConditional(tree);
+    }
+
+    ///
+    /// Translates a switch statement into an equivalent chained if statement.
+    /// See also switchFromIf for the reverse transformation used
+    /// in the UNSUSPECT direction.
```
keyboardDrummer, line 154: I still find this comment to be SUSPECT

keyboardDrummer (approved):

verifier/src/main/java/com/aws/jverify/verifier/compiler/Suspenders.java
robin-aws: Ah damn didn't realize there were two of them. :P

PR #263, opened 2025-07-22
Replace immutable with modifiable

keyboardDrummer:
### What was changed?
- Change `@Immutable Object` to `Object`, and `Object` to `@Modifiable Object`
- `@Modifiable Object` has `==` support, but can't be assigned to by Records and type parameters
- A class that extends `Object`, explicitly or implicitly, will extend `@Modifiable Object` instead.
- A type bound that extends `Object` simply extend `Object`.

### How has this been tested?
- Updated existing tests


verifier/src/test/java/com/aws/jverify/verifier/tests/PolymorphismDafnyResolutionErrors2.java
```diff
@@ -1,16 +1,16 @@
...
     public static void valueObjectHasNoEquality() {
         // Annotations on method parameters are broken ATM, so these are locals
-        @Immutable Object a = new Object();
-        Object b = new Object();
-        var c = a == b;
-//              ^ Error: == can only be applied to expressions of types that support equality (got ValueObject)
+        Object a = new Object();
+        @Modifiable Object b = new Object();
+        var c = a == (@Modifiable Object)b;
+//              ^ Error: == can only be applied to expressions of types that support equality (got Object)
```
keyboardDrummer, line 14: This Dafny error will disappear in a future PR, since I'll switch all the code to be ghost, and in ghost code, all Dafny types support equality.

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -383,13 +382,10 @@ public boolean isEnum(JCTree.JCExpression selected) {
         }
         return false;
     }
-
+    
+    // TODO Dangerous. Delete?
```
alex-chew: What makes this dangerous?

verifier/src/main/resources/com/aws/jverify/messages.properties
```diff
@@ -15,5 +15,5 @@ compiler.err.unusedContractMethod={0} is part of a @Contract class, but its sign
 compiler.err.noContractTarget=could not find a target for @Contract class ''{0}''
 compiler.err.annotationOnSourceContractClass=annotation ''{0}'' on @Contract class ''{1}'' is not allowed, because it must be placed on the contractee
 compiler.err.bodylessMethodWithoutContract=method ''{0}'' without a body still needs a contract. Use @Contract to specify a contract class.
-compiler.err.equalityOperatorRestricted=''{0}'' is not allowed when both operand''s types could be String, a record, or a boxed primitive, unless one of the operands is a null literal
+compiler.err.equalityOperatorRestricted=''{0}'' is only allowed when at least one operand''s type is modifiable.
```
alex-chew: Suggestion to clarify that we're referring to the JVerify concept specifically:

```suggestion
compiler.err.equalityOperatorRestricted=''{0}'' is only allowed when at least one operand''s type is @Modifiable
```

verifier/src/test/java/com/aws/jverify/verifier/tests/RecordsDafnyResolutionError.java
```diff
@@ -1,16 +1,15 @@
...
 import static com.aws.jverify.JVerify.postcondition;
 
 @JVerifyTest(exitCode = 22)
 class RecordsDafnyResolutionError {
     record Foo() {}
-    static void assignRecordToRefObject() {
-        Object o = new Foo();
-//      ^^^^^^^^^^^^^^^^^^^^^ Error: RHS (of type Foo) not assignable to LHS (of type Object)
+    static void assignRecordToModifiableObject() {
+        @Modifiable Object o = new Foo();
```
alex-chew, line 12: Aside from assignment, do we also prevent immutable-typed values from being assigned to modifiable-typed variables in other contexts? For example, passing a `new Foo()` as the argument for a `@Modifiable Object` parameter?

keyboardDrummer: Yes that would also be prevented. It will be an error in Dafny's type system.

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
keyboardDrummer: Removed the TODO

PR #265, opened 2025-07-22
Uses ' as specific separator to avoid name potential clashes

### LINKED ISSUE #255: Remove need for `avoidNameCollisions` option
Introduce a naming scheme that leads to Dafny that is both:
- Resolvable because there are no conflicts in names
- Parseable because the rules for identifiers are respected
  - Some identifiers can not start with _
  - Some identifiers are reserved

### Ideas
What Dafny allows for identifiers: https://dafny.org/dafny/DafnyRef/DafnyRef#sec-identifiers (letter, digit, ', _, ?)
Versus Java: https://docs.oracle.com/cd/E19798-01/821-1841/bnbuk/index.html (letters, digit, $, _)

When renaming an identifier, use a character that Dafny allows but Java does not, to prevent conflicts, so `'` or `?`. For example:
- if an identifier starts with _, prefix it with a',
- if an identifier is reserved, prefix it with r'

olivier-aws:
### What was changed?
Fixes #255 

### How has this been tested?
Use existing tests


keyboardDrummer:
Amazing, thanks!

PR #269, opened 2025-07-25
Support anonymous classes

keyboardDrummer:
### What was changed?
- Support local classes. Note that these can only be instantiate in a non-pure context, since they are translated to Dafny classes. I think this is the right design, since local classes can also be extended, unlike Dafny datatypes.
- Support anonymous classes, but not creating them in pure contexts. I think these could potentially be translated to Dafny datatypes, but only if they contain no mutable fields and do not extend/implement anything that's modifiable. I think it would be OK to implicitly determine if they should be a datatype or not, based on the fields.

### How has this been tested?
- Added a test `LocalAndAnonymousClasses`


verifier/src/test/java/com/aws/jverify/verifier/tests/LocalAndAnonymousClasses.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    interface SomethingDoerWithSpec {
+        int doSomething(int x, int y);
+
+        @Contract
+        class SomethingDoerWithSpecContract implements SomethingDoerWithSpec {
+            // TODO: Can't currently put @Nat on the return type because there's
+            // currently no way to indicate that on the lambda expression
+            // since it doesn't declare a return type anywhere.
```
robin-aws, line 155: Not true in this case! :)

verifier/src/test/java/com/aws/jverify/verifier/tests/LocalAndAnonymousClasses.java
```diff
@@ -0,0 +1,164 @@
...
+
+import static com.aws.jverify.JVerify.*;
+import static com.aws.jverify.JVerify.postcondition;
+import static com.aws.jverify.JVerify.precondition;
+
+@SuppressWarnings({"Convert2Lambda", "Anonymous2MethodRef"})
+@JVerifyTest(exitCode = 4, dafnyVerified = 25, dafnyErrors = 3, verifyPrintedDafny = true)
+public class LocalAndAnonymousClasses {
+
+    public void useLambdas() {
```
robin-aws:
```suggestion
    public void useAnonymousClasses() {
```

verifier/src/test/java/com/aws/jverify/verifier/tests/Lambdas.java
```diff
@@ -8,10 +8,14 @@
 import static com.aws.jverify.JVerify.postcondition;
 import static com.aws.jverify.JVerify.precondition;
 
-@JVerifyTest(exitCode = 4, dafnyVerified = 27, dafnyErrors = 3, verifyPrintedDafny = true)
+// TODO move method references to separate test file
```
robin-aws, line 11: Is there a big advantage to that? I found it handy to share the same interfaces to implement when testing the two features.

docs/guide/src/supported_java_features.md
```diff
@@ -31,6 +31,7 @@ JVerify intends to support the full Java language, but is currently being develo
   - Overriding `equals(Object)` or `hashCode()` is not allowed
 - `implements`
 - `extends`
+- Local and anonymous classes
```
robin-aws, line 34: AFAICT we also fully support nested classes now too, which we don't mention in either list. Just need to add references to enclosing class state to the existing `NestedClass` test.

robin-aws:
> Support anonymous classes, but not creating them in pure contexts. I think these could potentially be translated to Dafny datatypes, but only if they contain no mutable fields and do not extend/implement anything that's modifiable. I think it would be OK to implicitly determine if they should be a datatype or not, based on the fields.

Yeah, I think it would be better for the anonymous class to be annotated with something like `@Immutable`, but that doesn't seem syntactically possible. I worry about inferring it and having users confused if they change a class to be mutable and then can no longer create it in an expression context - we'll want a fairly verbose error explaining it.

robin-aws (changes requested):
Thanks, really nice this is all we need on top of reusing the LOWER phase :)

docs/guide/src/supported_java_features.md
keyboardDrummer: Let's add it when we add tests for nested instance classes :-)

verifier/src/test/java/com/aws/jverify/verifier/tests/Lambdas.java
keyboardDrummer: > Is there a big advantage to that? I found it handy to share the same interfaces to implement when testing the two features.

I'll add support for the two features separately, so it might be nice to have separate testfiles.

verifier/src/test/java/com/aws/jverify/verifier/tests/LocalAndAnonymousClasses.java
keyboardDrummer: Updated

PR #270, opened 2025-07-25
Define Object.equals in Dafny translation, more List method contracts

robin-aws:
### What was changed?

* Added `Object.equals` to additional.dfy. See comments for wrinkles however.
* Fleshed out the `Sequence<T>` translation a bit more.
* Added a few more `List` contracts.
  * Because it simplifies the `equals` problem tremendously, for now I'm working to add the necessary collection contracts to support "immutable uses" of these interfaces, meaning only using the subset of the API that does not refer to mutable state. This does mean we will need to return to add `reads()` clauses and otherwise change contracts, particularly on `equals`, to support the remaining API methods and mutable collections.
* Added a `JVERIFY_UPDATE_TEST_ANNOTATIONS=true` feature to the test engine to update the `@JVerifyTest` annotations after running tests. Pretty incomplete and needs careful review afterwards since it doesn't retain all properties, but still saved me some time and will be useful in the future, and can be improved later (especially to update the hat error annotations too).

### How has this been tested?
ListTest.java 

TODO:
* [x] ~(ListTest not passing any more after moving the ListContract back to builtin-contracts.java, and I can't tell why. I also suspect the existing ListContract was never working or broke at some point, because if I delete it all the tests still pass? :) Help!)~ - I just needed to add the new `useBuiltinContracts = true` argument to `@JVerifyTest` :P
* [x] all other test failures are because of the change in the dafnyVerified counts - adding `equals` for each concrete class bumps this up for most of them. Will update all the counts once the reviewer is satisfied on other accounts.


verifier/src/test/java/com/aws/jverify/verifier/tests/ListTest.java
```diff
@@ -0,0 +1,77 @@
...
+        var s  = List.of("one", "two", "three");
+        check(s.size() == 3);
+        check(s.get(1).equals("two"));
+        check(s.contains("two"));
+    }
+}
+
+@Contract
+@Modifiable
+abstract class ListContract<E> implements List<E> {
```
robin-aws: Defined directly in test for easier development, will move to builtin-contracts.java

library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -191,6 +191,13 @@ public static IntSequence sequence(int[] array, int fromIndex, int toIndex) {
     }
 
     public interface Sequence<T> {
+        /**
+         * Returns the elements at index {@code index}.
+         */
+        default T get(int index) {
+            throw new VerificationMethodExecutedException();
```
robin-aws: Question: why are we including a default body that just throws? Why not just leave these as regular interface methods?

keyboardDrummer: I don't know. Maybe they were copy pasted from before there were interface methods.

library/src/main/java/com/aws/jverify/MethodContract.java
```diff
@@ -0,0 +1,12 @@
+package com.aws.jverify;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.SOURCE)
+@Target({ElementType.METHOD})
+public @interface MethodContract {
```
keyboardDrummer: Can you add a comment?

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -527,6 +527,11 @@ public Type translateType(com.sun.tools.javac.code.Type type, IOrigin origin, JC
                     return null;
                 }
 
+                if (className.toString().equals("com.aws.jverify.JVerify$Sequence")) {
```
keyboardDrummer: Magic string

verifier/src/test/java/com/aws/jverify/verifier/tests/EqualsAndHashCode.java
```diff
@@ -0,0 +1,97 @@
...
+import com.aws.jverify.testengine.JVerifyTest;
+
+import static com.aws.jverify.JVerify.check;
+
+@JVerifyTest(dafnyVerified = 8, dafnyErrors = 0)
+public class EqualsAndHashCode {
+
+}
+
+class MyPair<A, B> {
```
keyboardDrummer: Consider making all these classes inner classes of `EqualsAndHashCode`, so they don't conflict with other tests.

verifier/src/test/java/com/aws/jverify/verifier/tests/EqualsAndHashCode.java
```diff
@@ -0,0 +1,97 @@
...
+    private final A a;
+    private final B b;
+
+    MyPair(A a, B b) {
+        this.a = a;
+        this.b = b;
+    }
+
+    @Pure
+    @MethodContract("equals")
```
keyboardDrummer: I don't follow the intention of `@MethodContract` very well yet. The test itself reminds my of function-by-method, since there is both a pure and an impure definition of equals.

I was thinking that for function-by-method, we would let the method body contain a call to `pureBy`, like this:

Java:
```java
class MyClass {
  @Pure
  int soPure(int x, int y) { ... }

  int soImpure(int x, int y) {
    pureBy(MyClass::soPure);

    <body>
  }
}
```

Dafny:
```dafny
function soPure(x: int, y: int): int { .... }

function soImpure(x: int, y: int) returns (x: int) {
  soPure(x, y)
} by method {
  <body>
}
```

library/src/main/java/com/aws/jverify/MethodContract.java
robin-aws: Whoops, I thought I'd deleted this and EqualsAndHashCode, I intended them to land in a later PR :P

verifier/src/test/java/com/aws/jverify/verifier/tests/EqualsAndHashCode.java
robin-aws: Yeah this is exactly how I was thinking as well, but as above it's half baked and meant to be in the next PR.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/workaround/TraitWithConstructorCompiler.java
```diff
@@ -76,6 +76,8 @@ private List<TopLevelDecl> buildTraitAndClassTwin(TraitDecl traitDecl,
         }
 
         if (classNeeded) {
+            classMembers.add(JavaToDafnyCompiler.equalsFunctionDeclaration(traitDecl.getOrigin()));
```
keyboardDrummer, line 79: Is this what generates the new verification conditions? It doesn't seem like something that needs verification. I guess it's verifying that the override has a valid contract, even though the contract is the same as the overridden one, and they're both empty?

robin-aws: Yup, I assumed it was just trivial wellformedness/overrides checks on the declaration.

library/src/main/java/com/aws/jverify/JVerify.java
robin-aws: Removed the defaults!

keyboardDrummer (approved):

PR #271, opened 2025-07-29
Polymorphic anonymous classes

keyboardDrummer:
### What was changed?
- Support polymorphic anonymous classes
- Explicitly provide generic arguments when calling constructors, instead of letting Dafny infer them, which improves reliability
- Move tests into packages to provide more structure
- Rename `ClassesExtendingClassesCompiler` to `TraitWithConstructorCompiler` and refactor code to match the new name, moving some code related to which superTraits to use at the Dafny level from here to `ClassCompiler`.
- Move most code related to `@Modifiable Object` into a new class `ModifiableObjectCompiler`

### How has this been tested?
- Added test `PolymorphicAnonymousClasses`
- Added test `NestedPolymorphism`


verifier/src/main/java/com/aws/jverify/verifier/compiler/BlockCompiler.java
```diff
@@ -357,9 +358,22 @@ public AssignmentRhs toAssignmentRhs(JCTree.JCExpression expr, IOrigin originOve
                 }
                 String ctorNameStr = compiler.nameCompiler.getCompiledName(newClass.constructor);
                 Name ctorName = new Name(origin, ctorNameStr);
-                var baseType = (NameSegment)compiler.expressionCompiler.toExpr(newClass.clazz);
```
keyboardDrummer, line 360: Using `.toExpr` to get a type was wrong because it does not handle generic arguments.

verifier/src/test/java/com/aws/jverify/verifier/tests/localClasses/PolymorphicAnonymousClasses.java
```diff
@@ -0,0 +1,64 @@
+package com.aws.jverify.verifier.tests.localClasses;
+
+import com.aws.jverify.testengine.JVerifyTest;
+
+@JVerifyTest(exitCode = 0, dafnyVerified = 7, dafnyErrors = 0,
+        verifyPrintedDafny = true)
+public class PolymorphicAnonymousClasses {
+
+    static class Anything {}
+    void lambdaForGenericInterfaces() {
```
robin-aws: Nit-picking, but these are all anonymous classes and not "lambda"s :)

verifier/src/test/java/com/aws/jverify/verifier/tests/nestedClasses/NestedInstanceClass.java
robin-aws, line 1: We seem to be missing some cases for non-static nested classes: actually constructing them (which involves handling synthetic fields like `this$1`) and referring to the enclosing state. IIRC I had them in a branch I passed to you but we may have lost them in the actual PR we merged?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/NameCompiler.java
```diff
@@ -111,7 +111,7 @@ private String getClassName(Symbol.ClassSymbol classSymbol) {
             return uncachedGetCompiledName(newTarget);
         }
         var occurrenceCount = classNameOccurrenceCounts.get(classSymbol.name);
-        if (occurrenceCount == null || occurrenceCount > 1) {
+        if (classSymbol.isAnonymous() || occurrenceCount == null || occurrenceCount > 1) {
```
robin-aws: This is necessary because otherwise we try to use `1` as an identifier, right? :) Worth a comment as I don't think that's obvious.

verifier/src/main/java/com/aws/jverify/verifier/compiler/BlockCompiler.java
```diff
@@ -357,9 +358,22 @@ public AssignmentRhs toAssignmentRhs(JCTree.JCExpression expr, IOrigin originOve
...
+                }
+                var baseType = (UserDefinedType)compiler.translateType(type, compiler.toOrigin(newClass));
+                var baseNameSegment = (NameSegment)baseType.getNamePath();
+                var baseName = baseNameSegment.getName();
+                if (baseName.contains(REFERENCE_OR_VALUE_OBJECT_NAME)) {
+                    // 'new Object' should always create a Dafny ModifiableObject
+                    baseName = REFERENCE_OBJECT_NAME;
+                }
+                var classBaseType = new NameSegment(baseNameSegment.getOrigin(), compiler.nameCompiler.CLASS_PREFIX + baseName,
+                        baseNameSegment.getOptTypeArguments());
```
robin-aws: This logic feels like it should live in the NameCompiler to me

robin-aws (changes requested):

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/NameCompiler.java
keyboardDrummer: Otherwise the identifier is `""`

verifier/src/main/java/com/aws/jverify/verifier/compiler/BlockCompiler.java
keyboardDrummer: I would like NameCompiler to be mostly about handling namespaces: removing packages/imports and overloading. We could split it into two so one is the `DafnyNameCompiler` (handles Dafny reserved keywords and identifier rules) and the other is the `NamespaceCompiler` (handles qualifying names to make them unique).

The modifiable/base `Object` code is I think part of a possibly more general feature to map a single Java type to multiple verification types. It would be great if we can extract that to a separate file. I've added a commit that does this just for the Object related type mapping code.

verifier/src/test/java/com/aws/jverify/verifier/tests/nestedClasses/NestedInstanceClass.java
keyboardDrummer: Thanks!! Added two test-cases

robin-aws (approved):

PR #274, opened 2025-07-31
Polymorphic lambdas

keyboardDrummer:
### What was changed?
- Prevent compiled name collisions between different lambdas
- Support lambda expressions that use generic type parameters in various ways
- Support anonymous classes with only final fields to be used as value types
- Support for value types that use outer generic type parameters
- Support for method references is disabled for now

### How has this been tested?
- Added a test `SameLineAndColumnLambdaA`
- Added the testfile `PolymorphicLambdas`
- Added more tests to `Lambdas`


verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaFrontEnd.java
```diff
@@ -136,9 +136,11 @@ public void finished(TaskEvent e) {
                     // Apply the second half of our pipeline as above (4 and onwards).
                     // See the implementation of JavaCompiler.compile() for similar lines,
                     // including the comment "these method calls must be chained to avoid memory leaks"
-                    envs.addAll(unsuspend(lower(suspend(
-                            unsubstitute(unlambda(substitute(
-                                    compiler.flow(compiler.attribute(todo)))))))));
+                    envs.addAll(unsuspend(
+                            lower(suspend(
+                                    unlambda(
+                                    compiler.flow(compiler.attribute(todo))
```
robin-aws:
```suggestion
                    envs.addAll(
                            unsuspend(lower(suspend(
                                    unlambda(
                                            compiler.flow(compiler.attribute(todo))
```

Or potentially other ways, but I think at least keeping `unsuspend(lower(suspend(` together makes sense.

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaFrontEnd.java
```diff
@@ -160,35 +162,14 @@ public void finished(TaskEvent e) {
...
-        TreeMaker localMake = TreeMaker.instance(context).at(Position.NOPOS);
+    private Queue<Env<AttrContext>> unlambda(Queue<Env<AttrContext>> envs) {;
 
         // Note JavaCompiler.desugar has some additional logic to
         // scan for classes that have lambdas first,
         // to not waste time with these traversals.
         // We could do the same here to save time in the future.
         for (Env<AttrContext> env : envs) {
-            env.tree = LambdaToMethod.instance(context).translateTopLevelClass(env, env.tree, localMake);
+            env.tree = new LambdaToAnonymousClassCompiler(context).translate(env.tree);
```
robin-aws: No need to add one instance to the `context` I think, but probably still better to create one instance outside of the loop at least?

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -363,6 +365,10 @@ public boolean isSynthetic(JCTree methodNode, Symbol.MethodSymbol methodSymbol)
         return methodNode.pos == containerPos;
     }
 
+    public static boolean isSynthetic(long flags) {
+        return (flags & Flags.SYNTHETIC) == Flags.SYNTHETIC;
```
robin-aws: nit, but I feel like `!= 0` is more idiomatic?

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -653,9 +659,55 @@ public static Symbol.MethodSymbol getJVerifyMethod(JCTree.JCMethodInvocation inv
         return fromJVerify(methodSymbol) ? methodSymbol : null;
     }
 
-    public boolean useConstructorFunction(JCTree.JCNewClass newClass,
-                                                 Symbol.ClassSymbol classSymbol) {
-        return isRecord(newClass.type) || isImmutableClass(classSymbol);
+    public boolean isValueType(Symbol.ClassSymbol classSymbol) {
+        Symtab symtab = Symtab.instance(context);
+        if (classSymbol.type == symtab.objectType) {
```
robin-aws, line 663: `java.lang.Object` isn't translated to a value type though, right? An Object might be a reference OR a value type.

verifier/src/test/java/com/aws/jverify/verifier/tests/expressions/lambdas/Lambdas.java
```diff
@@ -41,31 +58,66 @@ public void useLambdas() {
         // but if we map lambdas to datatype values incorrectly
         // they could be equal Dafny values.
         check(doer != doer2);
+    }
 
-        makeSomeClass(SomeClass::new);
+    void methodReferences() {
+//        doSomethingTwice(this::add);
+//        doSomethingTwice(Lambdas::staticAdd);
+//        makeSomeClass(SomeClass::new);
```
robin-aws, line 67: We should move "method references" back to the unsupported list in the user guide for now just to keep ourselves honest :)

verifier/src/test/java/com/aws/jverify/verifier/tests/expressions/lambdas/Lambdas.java
```diff
@@ -41,31 +58,66 @@ public void useLambdas() {
...
+            int level2 = 2;
+            SomethingDoer inner = (x2, y2) -> level1 + level2 + x + x2 + y2; 
+//                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Error: value does not satisfy the subset constraints of 'int32'
+
+            return inner.doSomething(1,2);
+        };
+        var result = outer.doSomething(1);
+
+        // TODO lambda result checks, for this and for the other lambdas in the file, should be added in a follow-up 
+        // These also need fixes.
```
robin-aws, line 94: What kind of fixes? Do they pass when they shouldn't?

verifier/src/main/java/com/aws/jverify/verifier/compiler/LambdaToAnonymousClassCompiler.java
robin-aws, line 1: This still seems to be doing more translation work than I expected. I had thought we could just rewrite a lambda like

```java
SomethingDoer doer = (x, y) -> {
  return z;
};
```

into

```java
SomethingDoer doer2 = new SomethingDoer() {
  @Override
  public int doSomething(int x, int y) {
    return z;
  }
};
```

and then we let LOWER take care of identifying captured variables/fields like `z`, creating constructors etc. Is there a reason that simpler approach doesn't work?

robin-aws (changes requested):
Partial review for now

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaFrontEnd.java
keyboardDrummer: Are we talking about `new LambdaToAnonymousClassCompiler(context)` ? It's an O(1) operation so I don't think it matters if we do it more often.

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
keyboardDrummer: Agreed

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
keyboardDrummer: My impression is that from a verification perspective the only important question is whether the thing is mutable. If a type is immutable, that means that the interface that it exposes is stateless. A mutable type may extend an immutable one. As long as the static type of a thing is immutable, you can treat it as such even though the instance type could be mutable. A type that's "either reference or value" can be treated as immutable.

One exception to mutable types extending immutable ones, is that once a type defines structural equality, like a record, then it cannot be used as the base for a reference type any more since then equality would be incorrect.

Anyways, I've renamed `isValueType` to `isImmutable`

verifier/src/test/java/com/aws/jverify/verifier/tests/expressions/lambdas/Lambdas.java
keyboardDrummer: They don't pass when they should. It requires a few different fixes. For example if you have `var x: interfaceType = lambda1234(...)` then Dafny won't be able to use the definition of lambda1234 when reasoning about `x`. I'm not sure how to think about it but it's something like "verification is not able to do valid type casts on its own", so we need to provide the right types pre-verification

verifier/src/test/java/com/aws/jverify/verifier/tests/expressions/lambdas/Lambdas.java
keyboardDrummer: I'll post the PR to bring them back today :-D

verifier/src/main/java/com/aws/jverify/verifier/compiler/LambdaToAnonymousClassCompiler.java
keyboardDrummer: That's a really good point..

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaFrontEnd.java
robin-aws: Sure, but doing an O(1) operation N times instead of once can matter, especially if the constant factor on the O(1) operation gets gradually higher over time (i.e. if the constructor starts to do more work) and N starts to get orders of magnitude larger.

My philosophy is just to encourage the habit of being more efficient when there's no risk or extra effort between the options. I've already spent more time on this specific example than it's worth based on the likely impact :) and I don't actually care if you change it, but I think it's a great example of the philosophy!

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
robin-aws: It depends heavily on how precisely you the "treated as" subtyping relationship here. There is at least two flavors of immutable list interfaces, for example: there's "I cannot mutate this but someone else could" (`Collections.immutableList(x)`) vs. "this value cannot be mutated by anyone" (`List.of(1, 2, 3)`). I think especially for JVerify the difference will end up mattering, and for the latter case it may not be sound to to treat a mutable value as a can-never-be-mutated immutable value.

Not a blocker on this PR, just signaling that I'll likely have to change this myself later.

verifier/src/main/java/com/aws/jverify/verifier/compiler/BlockCompiler.java
```diff
@@ -350,9 +351,10 @@ public AssignmentRhs toAssignmentRhs(JCTree.JCExpression expr, IOrigin originOve
         var origin = Objects.requireNonNullElseGet(originOverride, () -> compiler.toOrigin(expr));
         switch (expr) {
             case JCTree.JCNewClass newClass -> {
+                Symtab symtab = Symtab.instance(compiler.context);
                 Symbol.ClassSymbol classSymbol = (Symbol.ClassSymbol) TreeInfo.symbol(newClass.clazz);
-                if (compiler.useConstructorFunction(newClass, classSymbol)) {
-                    var datatypeValue = RecordCompiler.translateNewRecord(compiler.expressionCompiler, origin, newClass);
+                if (classSymbol.type != symtab.objectType && compiler.isImmutable(classSymbol)) {
```
robin-aws, line 356: This works :)

verifier/src/main/java/com/aws/jverify/verifier/compiler/ExpressionCompiler.java
```diff
@@ -158,8 +157,8 @@ public Expression toExpr(JCTree.JCExpression expr, IOrigin originOverride) {
 
     private Expression translateNew(JCTree.JCExpression expr, JCTree.JCNewClass newClass, IOrigin origin) {
         Symbol.ClassSymbol classSymbol = (Symbol.ClassSymbol) newClass.type.tsym;
-        if (compiler.useConstructorFunction(newClass, classSymbol)) {
-            return RecordCompiler.translateNewRecord(this, origin, newClass);
+        if (compiler.isImmutable(classSymbol)) {
```
robin-aws, line 160: Does this need to special case Object as well?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ExternalContractCompiler.java
```diff
@@ -19,65 +20,59 @@
...
 
-public class ExternalContractCompiler {
+public class ExternalContractCompiler extends TreeScanner {
     final JavaToDafnyCompiler compiler;
     public final Map<Symbol.ClassSymbol, List<JCTree.JCClassDecl>> declarationsForSymbolContract = new HashMap<>();
     public final Map<Symbol.ClassSymbol, ExternalTypeContract> externalContracts = new HashMap<>();
     public Map<com.sun.tools.javac.code.Type, com.sun.tools.javac.code.Type> contractClassTypeToContracteeType = new HashMap<>();
     public LinkedHashMap<Symbol.ClassSymbol, Symbol.ClassSymbol> contractClassToContractee = new LinkedHashMap<>();
-
+    public Map<Symbol.ClassSymbol, JCTree.JCCompilationUnit> foundClasses = new HashMap<>();
```
robin-aws, line 29: I know this was already floating around, but doesn't `JVerifyIndex` already give you this lookup?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ImmutableTypeCompiler.java
```diff
@@ -149,15 +153,36 @@ private void translateConstructor(JCTree.JCClassDecl classDecl, IOrigin origin,
...
+                (classDecl.sym.isAnonymous() || 
+                        JavaToDafnyCompiler.isSynthetic(classDecl.sym.flags()) || 
+                        constructor.getBody() == null || 
+                        !shouldVerify)) {
             Type outType = compiler.translateType(classDecl.type, constructor.getOrigin());
             Formal result = new Formal(origin, new Name(origin, NameCompiler.RETURN_VARIABLE_NAME), outType, false, false, null, null, false, false, false, null);
+
+            List<AttributedExpression> ens = constructor.getEns();
+            if (classDecl.sym.isAnonymous()) {
+                ens = classDecl.getMembers().stream().filter(m -> m instanceof JCTree.JCVariableDecl).map(member ->
```
robin-aws, line 169: I'm a bit surprised these explicit postconditions are necessary given this is a non-opaque function, did you actually find things didn't work without them?

Edit: realized this is bodiless so yes they are definitely necessary, but this also made me wonder why not just provide a `RecordType(a, b, c)` body instead.

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
keyboardDrummer: > and for the latter case it may not be sound to to treat a mutable value as a can-never-be-mutated immutable value.

If a mutable type inherits from an immutable one, then every inherited field remains final. The immutable inherited part remains immutable. It is conceptually similar to an mutable type having a field with an immutable type, so there is no soundness issue.

> There is at least two flavors of immutable list interfaces, for example: there's "I cannot mutate this but someone else could" (Collections.immutableList(x)) vs. "this value cannot be mutated by anyone" (List.of(1, 2, 3))

In JVerify, `List.of(1, 2, 3)` could return an immutable type, but `Collections.immutableList(x)` cannot since its values can change.

verifier/src/main/java/com/aws/jverify/verifier/compiler/LambdaToAnonymousClassCompiler.java
```diff
@@ -0,0 +1,240 @@
...
+                null,
+                List.of(make.Type(lambda.type)),
+                List.from(classBody)
+        );
+        classDef.sym = classSymbol;
+        classDef.type = classSymbol.type;
+        return classDef;
+    }
+
+    private JCMethodDecl createConstructor(Symbol.ClassSymbol classSymbol) {
```
robin-aws, line 128: I take it this is still necessary because we're already past the pipeline phase that makes implicit constructors explicit?

verifier/src/main/java/com/aws/jverify/verifier/compiler/LambdaToAnonymousClassCompiler.java
robin-aws, line 1: Much simpler and cleaner! :) Thanks!

verifier/src/test/java/com/aws/jverify/verifier/tests/expressions/lambdas/namecollision/SameLineAndColumnLambdaB.java
```diff
@@ -0,0 +1,16 @@
+package com.aws.jverify.verifier.tests.expressions.lambdas.namecollision;
+
+public class SameLineAndColumnLambdaB {
+    
+    
+    
+    void foo() {
+        useI(() -> {});
+    } // This needs to be on the same line as this method in A.java
```
robin-aws, line 9: Almost worth an explicit unit test to check this, given how brittle that setup is and how often we've changed the test file syntax.

robin-aws (approved):
Nice surgical fix and appreciate upping the test coverage on lambdas. Some non-blocking thoughts for later.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ExternalContractCompiler.java
keyboardDrummer: I don't know. The map is from a symbol to the compilation unit that the tree of that symbol is in.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ImmutableTypeCompiler.java
keyboardDrummer: > but this also made me wonder why not just provide a RecordType(a, b, c) body instead

That would work as well. That's nicer yes

verifier/src/main/java/com/aws/jverify/verifier/compiler/LambdaToAnonymousClassCompiler.java
keyboardDrummer: I'm not sure. I'll try removing it.

verifier/src/test/java/com/aws/jverify/verifier/tests/expressions/lambdas/namecollision/SameLineAndColumnLambdaB.java
keyboardDrummer: I'm not sure I follow the suggestion. What would you do?

verifier/src/main/java/com/aws/jverify/verifier/compiler/ExpressionCompiler.java
keyboardDrummer: I think so. I'll update this in a separate PR though, where I already added some object tests.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ExternalContractCompiler.java
robin-aws: Oh yeah you can use `JVerifyIndex.getEnv` for that: https://github.com/aws/jverify/blob/main/verifier/src/main/java/com/aws/jverify/verifier/compiler/JVerifyIndex.java#L81

verifier/src/test/java/com/aws/jverify/verifier/tests/expressions/lambdas/namecollision/SameLineAndColumnLambdaB.java
robin-aws: Add a test to something like `TestVerifier` that parses both files and checks that the lambdas are in fact at the same source location?

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
robin-aws: You're right it's fine to inherit immutable state from a concrete supertype. I'm more concerned with implementing an immutable variant of an interface like `List`, where the contract there is that something like `list.contains(x)` can never change value later.

verifier/src/main/java/com/aws/jverify/verifier/compiler/LambdaToAnonymousClassCompiler.java
keyboardDrummer: Didn't work

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
keyboardDrummer: > I'm more concerned with implementing an immutable variant of an interface like List, where the contract there is that something like list.contains(x) can never change value later.

Right, good point. Maybe it will be difficult to create an immutable variant like that. A `MutableList` will read `this` in its reading methods, while an `ImmutableList` does not read or refer to `this` at all. I don't think there can be a direct inheritance relation between those two.

Maybe we could do something like:

```dafny
trait MaybeMutable  {
  function myself(): set<object>

  function get(): int reads myself()
}

datatype Immutable extends MaybeMutable {
  function myself(): set<object> { {} }

  function get(): int reads myself()
}

trait Mutable extends MaybeMutable extends object {
  function myself(): set<object> { {this} }

  function get(): int reads myself()
}
```

robin-aws: Yup this is the direction I'm steering on my PR: https://github.com/aws/jverify/pull/270/files#diff-9a60823d37ee16c1e2fd195fc6b6c47164f9f3b4c96414007688003d192e3c03. (I just used `This()` instead of `myself()` :)

Although I'm actually inclined at this point to make the simplifying assumption that all collections are immutable, and therefore restrict the contracts to only allow the non-mutating methods. That will unblock more progress, and we can return to generalizing the interfaces more later.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ExternalContractCompiler.java
keyboardDrummer: I'll do this in a follow-up for timing reasons

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ImmutableTypeCompiler.java
keyboardDrummer: I'll do this in a follow-up for timing reasons

verifier/src/test/java/com/aws/jverify/verifier/tests/expressions/lambdas/namecollision/SameLineAndColumnLambdaB.java
keyboardDrummer: Sounds good, I'll do this in a follow-up for timing reasons

PR #279, opened 2025-08-04
Improve support for generic and local classes/lambdas

keyboardDrummer:
### What was changed?
1. Enable capturing generic types in anonymous classes and lambdas
2. Add a safety check that ensures a contract class uses the same type parameters as the contractee. Previously you would get Dafny level errors if the names did not match.
3. Improve `@Contract` feature so you don't have to `implements <interface>` to add contracts to an interface
4. Renamed some `@Contract` related tests

### How has this been tested?
- Added case `capturedGenericType` to PolymorphicAnonymousClasses to test change (1)
- Added case `HasGenericArgument` to `InterfaceContractErrors` to test change (2)
- Added case `noImplementsUser` to `InterfaceContractVerification` to test change (3)


verifier/src/main/java/com/aws/jverify/verifier/compiler/TypeDeclarationCompiler.java
```diff
@@ -380,7 +382,9 @@ private MethodOrConstructor translateImpureMethodOrLambda(JCTree source, JCTree.
                     return null;
                 }
             } else {
-                if (!(source instanceof JCTree.JCFieldAccess fa && fa.sym instanceof Symbol.DynamicMethodSymbol) && externalContract != null) {
+                // TODO Remove Symbol.DynamicMethodSymbol ??
```
robin-aws: I think we can yes, they can't appear in Java source code and now that we're not using LambdaToMethod I don't think any javac rewriting will insert them.

I think that means `LambdaCompiler` is dead code as well and should be removed.

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/InstrumentLower.java
```diff
@@ -0,0 +1,58 @@
...
+                            builder
+                            .field(named("make"))
+                                .transform(Transformer.ForField.withModifiers(Visibility.PUBLIC))
+                            .field(named("types"))
+                                .transform(Transformer.ForField.withModifiers(Visibility.PUBLIC))
+                            .method(named("access").
+                                    and(takesArgument(0, named("com.sun.tools.javac.code.Symbol"))))
+                                    .intercept(MethodDelegation.to(LowerInterceptor.class))
+                            .method(named("freevarDefs").
+                                            and(takesArguments(int.class, List.class, Symbol.class, long.class)))
```
robin-aws:
```suggestion
                                    and(takesArguments(int.class, List.class, Symbol.class, long.class)))
```

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/TypesWithoutErasure.java
```diff
@@ -0,0 +1,39 @@
...
+import com.sun.tools.javac.code.Type;
+import com.sun.tools.javac.code.Types;
+import com.sun.tools.javac.util.Context;
+import java.util.function.Supplier;
+
+public class TypesWithoutErasure extends Types {
+
+    public boolean eraseTypes = true;
+    
+    public <T> T withErased(Supplier<T> supplier) {
```
robin-aws: Unused?

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/InstrumentLower.java
```diff
@@ -0,0 +1,58 @@
...
+ * because it depends on erasure for type comparisons and lookups.
+ */
+public class InstrumentLower {
+
+    public static void installModification() {
+        try {
+            Instrumentation instrumentation = ByteBuddyAgent.install();
+
+            new AgentBuilder.Default()
+                    .type(named("com.sun.tools.javac.comp.Lower"))
```
robin-aws, line 34:
```suggestion
                    .type(is(Lower.class))
```

Unless that doesn't work for some reason?

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/InstrumentLower.java
```diff
@@ -0,0 +1,58 @@
...
+                    .transform((builder, typeDescription, classLoader, module, _) ->
+                            builder
+                            .field(named("make"))
+                                .transform(Transformer.ForField.withModifiers(Visibility.PUBLIC))
+                            .field(named("types"))
+                                .transform(Transformer.ForField.withModifiers(Visibility.PUBLIC))
+                            .method(named("access").
+                                    and(takesArgument(0, named("com.sun.tools.javac.code.Symbol"))))
```
robin-aws:
```suggestion
                                    and(takesArgument(0, Symbol.class))
```
?

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/LowerInterceptor.java
```diff
@@ -0,0 +1,71 @@
...
+import net.bytebuddy.implementation.bind.annotation.*;
+
+import java.lang.reflect.Field;
+import java.util.concurrent.Callable;
+
+/*
+Contains several intercepting methods. 
+The lowercased suffix of the name of methods is used to determine what it is intercepting
+If the name does not match anything, then the method is used to intercept the remainder of what needs intercepting
+ */
```
robin-aws: Teensy nit in general:

```suggestion
/*
 * Contains several intercepting methods. 
 * The lowercased suffix of the name of methods is used to determine what it is intercepting
 * If the name does not match anything, then the method is used to intercept the remainder of what needs
 * intercepting
 */
```

Although I've started using `///` markdown comments instead everywhere myself.

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/LowerInterceptor.java
```diff
@@ -0,0 +1,71 @@
...
+public class LowerInterceptor {
+    static Field typesField;
+    static Field makeField;
+    
+    static {
+        try {
+            makeField = Lower.class.getDeclaredField("make");
```
robin-aws, line 28: Worth a comment that these are made public in `InstrumentLower`, since the much more common approach is calling `Field.setAccessible` somewhere.

In fact it might be better to make LowerInterceptor a nested class of InstrumentLower, since they are so tightly coupled.

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/LowerInterceptor.java
```diff
@@ -0,0 +1,71 @@
...
+    /*
+    Intercept the access method
+     */
+    public static JCTree.JCExpression interceptAccess(@SuperCall Callable<JCTree.JCExpression> original,
+                                                @This Object lowerInstance,
+                                                @Argument(0) Symbol sym,
+                                                @Argument(1) JCTree.JCExpression tree,
+                                                @Argument(2) JCTree.JCExpression enclOp,
+                                                @Argument(3) boolean refSuper,
+                                                @AllArguments Object[] args) throws Exception {
```
robin-aws: Do you have to bind `@AllArguments` somewhere? Looks unused and better to avoid the array instantiation.

verifier/src/test/java/com/aws/jverify/verifier/tests/contracts/ExternalContractGhostField.java
```diff
@@ -10,7 +10,7 @@
 
 import static com.aws.jverify.JVerify.*;
 
-@JVerifyTest(useBuiltinContracts = false, exitCode = 0, dafnyVerified = 11, dafnyErrors = 0)
```
robin-aws, line 13: Why this change?

robin-aws (changes requested):
Nice work, the Byte Buddy approach is a pretty clean solution.

Just FYI, I believe AspectJ could handle this level of instrumentation as well and provide better tooling support for navigating the instrumentation declarations, as well as making the instrumentation happen at build time instead of runtime. If we end up having to instrument more and more of the javac code it would be worth considering switching to that approach.

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/LowerInterceptor.java
keyboardDrummer: What's the benefit of those?

verifier/src/test/java/com/aws/jverify/verifier/tests/contracts/ExternalContractGhostField.java
keyboardDrummer: Those were default values, so nothing changed.

robin-aws (approved):

PR #280, opened 2025-08-07
Add missing type contracts for referenced types

keyboardDrummer:
### What was changed?
- Add missing type contracts for referenced types. I also considered an alternative approach: pruning methods that were not verified and not called by anything that's verified, but that's more complicated because methods can also be referenced indirectly if they override a base member. That pruning solution could be more performant though so possibly we could implement it in the future.

### How has this been tested?
- Added test `UnusedUnverifiedContractRequirement.java`


verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/NameCompiler.java
```diff
@@ -5,6 +5,8 @@
 import com.sun.tools.javac.code.Symtab;
 import com.sun.tools.javac.code.Type;
 import com.sun.tools.javac.code.Type.*;
+import io.reactivex.rxjava3.core.Observable;
```
robin-aws: Why not use the built-in `java.util.concurrent.Flow.Publisher`?

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -505,6 +509,9 @@ public Type translateType(com.sun.tools.javac.code.Type type, IOrigin origin, JC
                 }
                 return new UserDefinedType(origin, new NameSegment(origin, "array" + nullableSuffix, List.of(elemType)));
             }
+            case com.sun.tools.javac.code.Type.IntersectionClassType intersectionClassType -> {
+                return null;
```
robin-aws, line 513: Isn't returning null in all of these spots just going to cause downstream NPEs? I'd think throwing an exception immediately would be much friendlier.

verifier/src/main/java/com/aws/jverify/verifier/compiler/TypeDeclarationCompiler.java
```diff
@@ -189,14 +202,27 @@ private TopLevelDeclWithMembers translateInterfaceOrClass(JCTree.JCClassDecl cla
 
         var definingSymbol = getCurrentTypeSymbol(classDecl.sym);
 
+        List<JCTree.JCTypeParameter> javaTypeParams = classDecl.typarams;
+        if (classDecl.sym.isDirectlyOrIndirectlyLocal()) {
+            javaTypeParams = compiler.getAllOwnerTypeParameters(classDecl.sym).toList();
```
robin-aws: Shouldn't this be additive, concatenating the enclosing type parameters with those directly on the class?

(Existing code moving I know)

verifier/src/main/java/com/aws/jverify/verifier/compiler/TypeDeclarationCompiler.java
```diff
@@ -582,7 +613,32 @@ private List<Formal> getIns(Symbol.MethodSymbol methodSymbol, boolean shouldVeri
...
-        var name = new Name(origin, compiler.nameCompiler.RETURN_VARIABLE_NAME);
+        var name = new Name(origin, NameCompiler.RETURN_VARIABLE_NAME);
         return new Formal(origin, name, syntacticType, false, false, null, null, false, false, false, null);
     }
+
+    public void addMissingTypeContracts(List<FileHeader> filesStarts) {
+        var dummyToken = new Token(1, 1);
+        IOrigin dummyOrigin = new TokenRangeOrigin(dummyToken, dummyToken);
+        while(!missingContracts.isEmpty()) {
+            var classSymbol = missingContracts.stream().findFirst().get();
```
robin-aws:
```suggestion
            var classSymbol = missingContracts.iterator().next();
```

Fair bit less overhead compared to the streams version.

verifier/src/main/java/com/aws/jverify/verifier/compiler/TypeDeclarationCompiler.java
```diff
@@ -582,7 +613,32 @@ private List<Formal> getIns(Symbol.MethodSymbol methodSymbol, boolean shouldVeri
...
     private Formal makeReturnFormal(IOrigin origin, Type syntacticType) {
-        var name = new Name(origin, compiler.nameCompiler.RETURN_VARIABLE_NAME);
+        var name = new Name(origin, NameCompiler.RETURN_VARIABLE_NAME);
         return new Formal(origin, name, syntacticType, false, false, null, null, false, false, false, null);
     }
+
+    public void addMissingTypeContracts(List<FileHeader> filesStarts) {
+        var dummyToken = new Token(1, 1);
+        IOrigin dummyOrigin = new TokenRangeOrigin(dummyToken, dummyToken);
+        while(!missingContracts.isEmpty()) {
```
robin-aws, line 642: Just to check (and probably worth a comment): you do this instead of a simple loop because filling in a missing contract for a type could also discover more types missing contracts, correct?

robin-aws (changes requested):

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/NameCompiler.java
keyboardDrummer: I tried it but the API is not comfortable to use

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
keyboardDrummer: It's a valid scenario. If you have a local variable with an inferred intersection type, then `translateType` should return null and the Dafny type should be inferred as well. I think it's also not the first place where this method is returning `null`

verifier/src/main/java/com/aws/jverify/verifier/compiler/TypeDeclarationCompiler.java
keyboardDrummer: yes, because of base types

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
robin-aws: You're right, it was at least already returning null for unsupported types. I'm not confident every CALLER is handling a null result yet but we'll fix that as needed then. :)

verifier/src/main/java/com/aws/jverify/verifier/compiler/TypeDeclarationCompiler.java
keyboardDrummer: Renamed the method to `getOwnAndEnclosedTypeParameters`

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/NameCompiler.java
keyboardDrummer: It seems like `java.util.concurrent.Flow` is missing utilities for using it. I have not found a library that provides those. The common approach seems to be to use `io.reactivex.rxjava3`, which allows exposing its own types as `java.util.concurrent.Flow` types.

robin-aws: As discussed offline, it seems like rxjava is a large and less-widely-used reactive programming solution at this point than the built in `java.util.concurrent.Flow` utilities, so I think it's better to use those. The one thing rxjava seems to have that Java itself doesn't is a simple publisher that just calls onNext synchronously. The builtin `SubmissionPublisher` doesn't quite seem to support that mode AFAICT, so I just added a very simple version that does.

I've pushed the commit, but feel free to revert it if you don't agree and want to discuss further.

robin-aws (dismissed):
LGTM, but there's a chance the CI flakiness is introduced in this PR. We should confirm it already exists before merging, and otherwise get to the bottom of it first.

PR #281, opened 2025-08-07
Verify printed Dafny by default in the test engine

robin-aws:
### What was changed?
Switched the default on `JVerifyTest.verifyPrintedDafny`, since the current test suite doesn't hit any bugs that would block this.

### How has this been tested?
Existing tests.


keyboardDrummer (commented):
I'm not in favor of this change because of the increase in test suite runtime: 2:15 rather than 1:15. An more performant alternative would be adding a mode to the Dafny `--print` option to verify that the printed names are valid. Also, I'm not worried about having a bug in the print Dafny feature, because it's meant for developing JVerify.

PR #283, opened 2025-08-11
User guide updates and test grouping

keyboardDrummer:
### What was changed?
- Organize tests to make it more clear what part of Java we support
- Add an `immutable_types.md` section to the documentation
- Small documentation fixes

### How has this been tested?
Code changes are covered by existing tests. Documentation changes were viewed manually using `mdbook`


library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -49,12 +49,6 @@ public static void invariant(boolean condition) {
     public static <T> void postcondition(Predicate<T> predicate) {
     }
 
-    /**
-     * The given function must evaluate to true after this method call
-     */
-    public static <T> void postcondition(BooleanSupplier predicate) {
```
robin-aws, line 55: Good call!

verifier/src/main/resources/builtin-contracts.java
```diff
@@ -145,6 +145,7 @@ public long longValue() {
 @Contract(Boolean.class)
 class BooleanContract {
 
+    @Pure
```
robin-aws, line 148: This is valid only because we're mapping the primitive boxing types to Dafny types. Is there a test that `==` on the boxing types is forbidden?

docs/guide/src/code_types.md
```diff
@@ -53,7 +51,7 @@ Summary table:
 
 JVerify guarantees that a `@Pure` method is deterministic, meaning that given the same  it returns the same output. To use this property however, we need to know what the input is. In a language with references like Java, it's not enough to know what the method arguments are. 
 
-JVerify requires that a method marked with `@Pure` is explicit about which objects it reads fields from. This can be specified using `reads` calls. The `reads` method takes one or multiple `object` arguments. Example:
+JVerify requires that a method marked with `@Pure` is explicit about which objects it reads non-final fields from. This can be specified using `reads` calls. The `reads` method takes one or multiple `object` arguments. Example:
```
robin-aws:
```suggestion
JVerify requires that a method marked with `@Pure` is explicit about which objects it reads non-final fields from. This can be specified using `reads` calls. The `reads` method takes one or multiple `Object` arguments. Example:
```

docs/guide/src/immutable_types.md
```diff
@@ -0,0 +1,38 @@
+# Immutable types
+
+JVerify divides all types into two categories: mutable and immutable. Compared to immutable types, mutable ones have these properties:
+- Can have non-final fields
+- Can use reference equality `==`
+- Can _not_ be instantiated in a pure context
+
+Immutable types are more reusable than mutable ones, since they can be instantiated in pure contexts. Mutable types are useful when they are used to improve performance, which their mutable fields and reference equality can enable.
+
+Records are always immutable and classes are mutable with an exception. Interfaces are immutable by default, but can be made mutable by annotating them with `@Modifiable`.
```
robin-aws: What exception?

Edit: found it below, I would add something like "(see below)".

docs/guide/src/immutable_types.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Object
+
+In Java an important type is `java.lang.Object`. JVerify distinguishes between these two types of `Object`:
+- `Object`: no support for `==`
+- `@Modifiable Object`: supports `==`, inherits from `Object`
+
+Records, immutable interfaces and type parameters always inherit from `Object`. Classes and mutable interfaces always inherit from `@Modifiable Object`.
+
+Only `@Modifiable Object` can be instantiated directly, so `new Object` is always interpreted as `new @Modifiable Object`.
```
robin-aws:
```suggestion
`Object` cannot be instantiated directly, so `new Object` is always interpreted as `new @Modifiable Object`.
```

Other version sounds like no other types at all can be instantiated directly :)

docs/guide/src/immutable_types.md
```diff
@@ -0,0 +1,38 @@
+# Immutable types
+
+JVerify divides all types into two categories: mutable and immutable. Compared to immutable types, mutable ones have these properties:
+- Can have non-final fields
+- Can use reference equality `==`
+- Can _not_ be instantiated in a pure context
```
robin-aws:
```suggestion
- They can have non-final fields
- They can use reference equality `==`
- They _cannot_ be instantiated in a pure context
```

docs/guide/src/immutable_types.md
```diff
@@ -0,0 +1,38 @@
...
+
+Immutable types are more reusable than mutable ones, since they can be instantiated in pure contexts. Mutable types are useful when they are used to improve performance, which their mutable fields and reference equality can enable.
+
+Records are always immutable and classes are mutable with an exception. Interfaces are immutable by default, but can be made mutable by annotating them with `@Modifiable`.
+
+A mutable type may inherit from an immutable one. This may sound strange, but it only means that the mutable type will have an immutable part. An immutable type can not inherit from a mutable one.
```
robin-aws:
```suggestion
A mutable class may inherit from an immutable class. This may sound strange, but it only means that the mutable type will have an immutable part. An immutable type can not inherit from a mutable one.
```

Being more specific since a mutable type cannot implement an immutable interface. Might be worth showing that case with the error below as well.

docs/guide/src/immutable_types.md
```diff
@@ -0,0 +1,38 @@
...
+
+Here is an example showing the rules for Object types:
+
+```java
+{{#include ../../../examples/src/test/java/com/aws/jverify/examples/ObjectRules.java}}
+```
+
+### Immutable classes
+
+We said before that classes are mutable with an exception. The exception is that classes from external libraries can be marked as immutable, but this must only be done if they behave as such. Examples of immutable classes are the boxed primitive classes such as `Integer` and `Double`, and the `BigInteger` and `BigDecimal` class.
```
robin-aws, line 38: (to debate later) This still feels arbitrarily limiting to me. If someone owns an existing complicated Java class in source that only has final fields and behaves immutably, why not let them mark it as immutable rather than make them rewrite it as a record?

docs/guide/src/supported_java_features.md
robin-aws, line 1: Awesome that it's practical to switch to only what we don't support, great progress! :)

docs/guide/src/supported_java_features.md
```diff
@@ -1,65 +1,37 @@
...
+- `yield` in switch expressions
+
+#### Java 17
+- `permits` clauses are ignored in verification
+- Pattern matches in switch expressions and statements
+- Explicit constructors on records
+- Explicit accessors on records
+- Support `null` value and comparison for record types
+- Using `==` on records. Note that doing so is generally not what you want.
+- Overriding `equals(Object)` or `hashCode()` on records is not supported. Note that doing so is generally not what you want.
```
robin-aws, line 30: We don't support it yet on non-record types either.

robin-aws (changes requested):

verifier/src/main/resources/builtin-contracts.java
keyboardDrummer: `(((Object)10) == ((Object)10)) == true` in Java. Somehow `==` always unboxes primitives before being applied.

docs/guide/src/immutable_types.md
keyboardDrummer: > Being more specific since a mutable type cannot implement an immutable interface.

That's not true. A mutable type can implement an immutable interface. Java does not allow inheriting from records, but that's not a JVerify thing. What JVerify does not allow is for an immutable type to inherit from a mutable one.

For soundness, a pure context must behave deterministically. That means that it can not create objects that have reference equality, nor can it create mutable objects, since mutability allows you to test for reference equality:

```dafny
trait T extends object {
  var x: int
}

method M(t1: T, t2: T) 
  modifies t1, t2
{
  var x1 := t1.x;
  t2.x := 3;
  if (x1 != t1.x) {
    assert t1 == t2; // pass
  }
}
```

docs/guide/src/immutable_types.md
keyboardDrummer: I agree we can allow it. We would only need to check that the class purely has final fields. The `equals`  and hashcode implementations does not need to cover all fields like records do, as long as they don't use `==`.

That said, I think it's low priority to allow it and we can delay it. We can put a GH issue on the board to enable it before GA.

docs/guide/src/supported_java_features.md
keyboardDrummer: We can fix that soon enough it seems :)

verifier/src/main/resources/builtin-contracts.java
robin-aws: Oh careful, that's because SOME common boxed values are cached for performance. E.g.: https://github.com/corretto/corretto-jdk/blob/05b664d50d91e3d1846f170e363cd1b3eb4e5a84/src/java.base/share/classes/java/lang/Integer.java#L1004-L1006

`(((Object)1000) == ((Object)1000)) == false`!

docs/guide/src/immutable_types.md
robin-aws: Ah I think we're both right, because "immutable interface" is ambiguous.

What I'm stuck on (for example) is that if you have something like an `ImmutableList<T>` interface, it will have a method like `get(i)` with `reads {}`, because it cannot read any mutable state. If you have a mutable `ArrayList<T>` class, its `get(i)` will have something like `reads {internalArray}` at the least, or more generally something like `reads Repr()`. `ArrayList<T>` cannot implement `ImmutableList<T>` in this case.

I think we need to clarify in this section that interfaces can either be mutable or "either", and soften the language that all types are either mutable or immutable. Perhaps we should say all concrete/instantiable types are either mutable or immutable?

Then I think what you're really saying here is "a mutable type can inherit from a type that is not necessarily mutable".

verifier/src/main/resources/builtin-contracts.java
keyboardDrummer: Cool, thanks :) I've added the test.

docs/guide/src/immutable_types.md
keyboardDrummer: > ArrayList`<T>` cannot implement ImmutableList`<T>` in this case.

It can but that wouldn't give `ArrayList<T>` the API that you want. You can create another type though `ImmutableAndArrayList<T>` that contains both a mutable and an immutable list, and which implements `ImmutableList<T>`.

I've added added another sentence to the documentation to make it more clear that immutable types have immutable contracts and that these remain immutable when inheriting them.

EDIT: 

Sorry I see the "either" case that you're describing, where there is an immutable interface but it uses "reads *" or something similar (Repr) so it can later be implemented using either immutable or mutable backing values.

The restriction that immutable types can't refer to `this` in modifies or reads clauses seems like it serves just to prevent you from writing unidiomatic code, rather than a hard restriction needed for soundness.

Let me make a PR to update the documentation ^,^

PR #284, opened 2025-08-12
Various fixes

keyboardDrummer:
### What was changed?
- Support lambdas in static methods
- Support `Class::methodName` references in contracts
- Slightly more support for raw types

### How has this been tested?
- Added tests for the above fixes


docs/guide/src/overview.md
```diff
@@ -17,7 +17,7 @@ No, JVerify is not yet ready for production use.
 However, you can use the [installation](./installation.md) and [Running JVerify](./first_verification.md) sections in this guide to explore what it can currently do. Note that you will run into poor error messages, unsupported features and bugs - that are not yours ;-).
 
 #### Does JVerify support the entire Java language?
-Yes, it will support all Java language features. Note that we differentiate between Java the language, as documented in [its specification](https://docs.oracle.com/javase/specs/), and the Java standard library. More on JVerify's support for the latter is further down this FAQ.
+Not yet, but it will support all Java language features. Note that we differentiate between Java the language, as documented in [its specification](https://docs.oracle.com/javase/specs/), and the Java standard library. More on JVerify's support for the latter is further down this FAQ.
```
ssomayyajula, line 20: Good catch...

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -583,6 +472,163 @@ public Type translateType(com.sun.tools.javac.code.Type type, IOrigin origin, JC
...
+        }
+        var typeArguments = typeArgumentsStream.toList();
+        if (typeArguments.stream().anyMatch(Objects::isNull)) {
+            return null;
+        }
+        if (typeArguments.isEmpty()) {
+            if (classType.tsym.getTypeParameters().isEmpty()) {
+                typeArguments = null;
+            } else {
+                // TODO needs a test. Something like the following
```
ssomayyajula, line 541: Do you mean to address this TODO in this PR?

ssomayyajula (commented):
Lambdas and method references make sense to me, but what is a "raw type"? Are they metavariables standing in place for generic type parameters?

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
keyboardDrummer: Jep, thanks

oyendrila-dobe (dismissed):

keyboardDrummer:
> Lambdas and method references make sense to me, but what is a "raw type"? Are they metavariables standing in place for generic type parameters?

Raw types are a Java concept. It's when you use a "generic type", one that's polymorphic, but without applying any types. A raw type has type checking turned off for its type parameters. It's similar to as if you have used TypeScript's "any" type for the type parameters. Given a variable of raw type `List` (it's raw because there's no `<>`), I can add unrelated types to it and Java will not complain.

Raw types are meant as a backwards compatibility feature for when Java didn't have generics yet.

PR #286, opened 2025-08-13
More collections contracts (WIP)

### LINKED ISSUE #54: Add partial contract for collections used by customer B
For `List<>`:
- List.of
- contains
- size
- get
- isEmpty
- ~stream~ - See #275 

For `Set<>`:
- ~stream~ - See #275 
- size
- isEmpty
- contains
- ~findFirst~ - Actually a Stream method, See #275 

~For Stream add a contract for:~  - See #54 and #275
- allMatch
- map
- collect
- toList

~`IntStream.range`, `IntStream.allMatch`, `IntStream.mapToObj`
`Collectors.toSet()`~ - See #54 and #275

For `Map`:
- `Map.of`
- Some more things, depends on how we implement it on the Java side, functional or mutable.
  - `Map.ofEntries`
  - `entrySet`

~`Optional`~  - See https://github.com/aws/jverify/issues/272

`Math.min`

robin-aws:
### What was changed?
Resolves #54.

* Adds `JVerify.Set` and `JVerify.Map` which map to their Dafny counterparts directly.
* Added `JVerify.Set.all` and `JVerify.Set.map` as a way to expose set comprehensions (which are needed for some contracts) but in a more Java-idiomatic way, somewhat similar to the Streams API.
* Adds contracts for a few basic Set and Map types.
* Adds missing `extends Object` supertraits in a few more cases.

TODO:

* [x] The reference to `Map.entry` doesn't yet work because Java has different rules for type parameters on static methods: in Java the method reference looks like `Map.entry<K, V>(K, V)`, but Dafny expects it to be `Map<K, V>.entry(K, V)`.

### How has this been tested?
Added to ListTest.java (to be renamed)


verifier/src/main/java/com/aws/jverify/verifier/compiler/TypeDeclarationCompiler.java
```diff
@@ -206,8 +206,7 @@ private TopLevelDeclWithMembers translateClass(JCTree.JCClassDecl classDecl, IOr
                 })
                 .collect(Collectors.<Type>toList());
 
-        if (definingSymbol == symtab.objectType.tsym || definingSymbol == symtab.recordType.tsym) {
-            superTraits = new ArrayList<>();
+        if (superTraits.isEmpty()) {
```
keyboardDrummer, line 245: Can we add a test for this that does not depend on builtins?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ImmutableTypeCompiler.java
```diff
@@ -46,13 +47,15 @@ public TopLevelDeclWithMembers translateValueType(Symbol.ClassSymbol classSymbol
         var superClass = currentTypeSymbol.getSuperclass();
         if (superClass != null) {
             Symtab symtab = Symtab.instance(typeDeclarationCompiler.compiler.context);
-            if (superClass.tsym == symtab.objectType.tsym) {
+            if (superClass.tsym == symtab.objectType.tsym || superClass.getKind() == TypeKind.NONE) {
```
keyboardDrummer, line 47: Can we add a test for this that does not depend on builtins?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ImmutableTypeCompiler.java
```diff
@@ -46,13 +47,15 @@ public TopLevelDeclWithMembers translateValueType(Symbol.ClassSymbol classSymbol
...
                 traits.addFirst(new UserDefinedType(origin, new NameSegment(origin, JavaToDafnyCompiler.REFERENCE_OR_VALUE_OBJECT_NAME, null)));
             } else {
                 if (compiler.typeHasAContract(superClass)) {
                     traits.addFirst(compiler.translateType(superClass, origin, null));
                 }
             }
+        } else {
+            traits.addFirst(new UserDefinedType(origin, new NameSegment(origin, JavaToDafnyCompiler.REFERENCE_OR_VALUE_OBJECT_NAME, null)));
```
keyboardDrummer, line 55: Can we add a test for this that does not depend on builtins?

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -531,6 +531,14 @@ public Type translateType(com.sun.tools.javac.code.Type type, IOrigin origin, JC
                     var arguments = classType.getTypeArguments().stream().map(a -> translateType(a, origin)).toList();
                     return new SeqType(origin, arguments);
                 }
+                if (className.toString().equals(JVerify.Set.class.getName())) {
```
keyboardDrummer: Maybe these three `ifs` could be moved to a method `JVerifyGhostExpressionCompiler.translateClassType` that returns a nullable Dafny type.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/JVerifyGhostExpressionCompiler.java
```diff
@@ -103,6 +110,30 @@ public Expression jverifyLibMethodToExpr(JCTree.JCMethodInvocation invocation) {
                 var element = expressionCompiler.toExpr(args.getFirst());
                 return new FreshExpr(compiler.toOrigin(invocation), element, null);
             }
+            case "contractOf" -> {
+                return expressionCompiler.toExpr(args.getFirst());
+            }
+            case "all", "map" -> {
+                if (args.size() != 1) {
```
keyboardDrummer: Please extract this "all","map" block to a separate method

verifier/src/test/java/com/aws/jverify/verifier/tests/ListTest.java
```diff
@@ -32,3 +42,210 @@ void preconditionFails() {
...
+    static <E> Set<E> of(E e1, E e2) {
+        postcondition((Set<E> r) ->
+                r.size() == 2 &&
+                r.contains(e1) &&
+                r.contains(e2));
+        throw new ContractException();
+    }
+
+    static <E> Set<E> of(E e1, E e2, E e3) {
+        postcondition((Set<E> r) ->
```
keyboardDrummer: You can write `postcondition((SetContract<E> r) -> r.elements.equals(JVerify.setOf(e1,e2,e3))`, if we add `JVerify.setOf`

robin-aws: Ah interesting - I considered that but at the time didn't see how to reference `elements` in the contract. Hence adding `JVerify.contractOf<T>` to address that in general. But using the contract type directly in the lambda seems good too if that already works.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ImmutableTypeCompiler.java
robin-aws: Not easily, this only affects immutable interfaces, which is currently only reachable via `@Contract(immutable = true)`

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ImmutableTypeCompiler.java
robin-aws: Same as above (this is more defensive programming because I'm not sure it's possible to have a `null` superclass rather than a "none" type superclass).

verifier/src/main/resources/builtin-contracts.java
```diff
@@ -97,8 +103,195 @@ public boolean contains(Object o) {
     }
 }
 
-@Contract(SequencedCollection.class)
-interface SequencedCollectionContract<E> extends Collection<E> { }
+@Contract(value = Set.class, immutable = true)
+abstract class SetContract<E> implements Set<E> {
+
+    JVerify.Set<E> elements;
```
keyboardDrummer, line 109: Based on how you defined the `of` methods, like you could also remove `elements`, leave out the postcondition for `contains` and `size` and add a single lemma that says an empty size means it never contains anything.

I think it would be good to have some of the mutable methods in this contract as well, to get a better feeling of how it will all fit together. We can do that later though.

keyboardDrummer (approved):

keyboardDrummer:
Removed `contractOf` for now and used the other mechanism instead.

robin-aws:
> Removed contractOf for now and used the other mechanism instead.

I'm definitely happy with that. I do think we'll need something like `contractOf` eventually, for cases when you need to reference ghost state from arguments rather than return values for example. But we can cross that bridge when we come to it.

verifier/src/main/resources/builtin-contracts.java
robin-aws: > Based on how you defined the of methods, like you could also remove elements, leave out the postcondition for contains and size and add a single lemma that says an empty size means it never contains anything.

I agree there's at least two different paradigms for defining contracts on collection types:

1. Define a ghost field to hold the equivalent value and relate method contracts to that.
2. Just define the contracts of methods in reference to other methods.

There's a bit of a mix of both currently and it would at least be better to be consistent. I do think I like 2 because it avoids having to work with both the idiomatic Java collections and the Dafny ones, but I suspect we'll also have to add lots of lemmas or postconditions to encode the collection semantics instead. I think we'll know better which approach is best once we actually try to prove things using these contracts.

> I think it would be good to have some of the mutable methods in this contract as well, to get a better feeling of how it will all fit together. We can do that later though.

Yup and I intentionally left out all mutating methods and only support immutable patterns of use of these types, because otherwise you have to add non-empty `reads` clauses to all of them and it gets much more complicated. I'm hoping we can design a good way to treat immutable and mutable collections as different types.

PR #291, opened 2025-08-15
Immutable types v2

keyboardDrummer:
NOTE!!! Instead of reviewing, please reply to this comment: https://github.com/aws/jverify/pull/291#issuecomment-3209582207

### What was changed?
Update documentation and naming related to immutable types. Happy to take comments and undo the renaming.

### How has this been tested?
- Refactoring


docs/guide/src/immutable_types.md
```diff
@@ -1,38 +1,36 @@
-# Immutable types
+# Using 'new' in pure code
```
keyboardDrummer, line 1: I wonder whether we should even do away with the mutable and immutable terminology and just focus on records and classes, so we build more on the context that Java developers already have. The explanation then simply becomes "In pure contexts, you can only instantiate records , or classes that behave like records."

And then later: "You can't use reference equality `==` on records unless they are marked with `@Reference`, but then they can no longer be created in a pure context."

keyboardDrummer:
@robin-aws how'd you feel about replacing the terms immutable and mutable types, with pure and impure types?

- A pure type: 
  - can be created in both pure and impure contexts.
  - does not expose any non-deterministic behavior, so
    - does not support reference equality and may only declare final fields. 
    - may only inherit from pure types
- An impure type:
  - can not be created in pure contexts. 
  - supports reference equality and may declare non-final fields
  - may inherit from both pure and impure types

Mapping:
- Interface with `@Pure`: pure type
- Interface without annotation: impure type
- Record without annotation: pure type
- Class without annotation: impure type
- Class with `@Pure`: pure type. Currently only supported for external classes
- [Possible future]: record with `@Pure(false)`: impure type, to enable reference equality

Relation between pure/impure and immutable/mutable:
- An abstract type that can be either mutable or immutable, can be made pure at the cost of losing reference equality
- An immutable type can be made pure at the cost of losing reference equality
- A mutable type is never pure

Example of the relation pure/impure and immutable/mutable:
```java
package com.aws.jverify.examples;

import com.aws.jverify.*;

@Pure
interface Container<T> {
    default T get() {
            reads(everything()); // we use reads(everything()) so implementing methods can still decide whether to be mutable or not
            decreases(0);
    }
}

interface MutableContainer<T> extends Container<T> {
    default void set(T value) {
        modifies(this); // referring to this is only allowed because this interface is impure
    }
}

@Pure
interface ImmutableContainer<T> extends Container<T> {
    @Pure
    @Override
    default T get() {
        // no reads clause - means its empty
    }
}

/**
 * Records are pure by default, so they can be instantiated in a pure context
 */
record ImmutableContainerImpl<T>(T value) implements ImmutableContainer<T> {
    @Pure
    @Override
    public T get() {
        decreases(0);
        // no reads clause - means its empty
        return value;
    }
    
    @Pure
    public static <T> ImmutableContainerImpl<T> createMe(T value) {
        return new ImmutableContainerImpl<T>(value);
    }
}

/**
 * Classes are impure by default, so they can only be created in an impure context.
 */
class MutableContainerImpl<T> implements MutableContainer<T> {
    T value;

    @Pure
    @Override
    public T get() {
        decreases(0);
        reads(this); // refines the inherited reads clause from 'all' to 'this'
        return value;
    }
    
    public void set(T value) {
        modifies(this);
        this.value = value;
    }
    
    @Pure
    public static <T> MutableContainer<T> createMe() {
        return new MutableContainer<T>();
//             ^ error: using 'new' in an expression to create an instance of an impure type is not supported
    }
}
```

I think changing interfaces to be impure by default is no biggie. When you inherit such an interface with a pure type, the error message will clearly indicate you can resolve it by adding `@Pure` to the interface.

However, without a `@Modifiable` `@Reference` or `@Impure`, we'd have to make `Object` the impure one by default. I'm guessing that most commonly you want a pure object, since you only need an impure object to do reference equality and there are quite a few methods in the Object class (which you can use with a pure object). Would that be enough reason to introduce an `@Impure` ?

robin-aws:
> how'd you feel about replacing the terms immutable and mutable types, with pure and impure types?

Overall I like the use of the terms! I think they would not *replace* immutable and mutable though. The relationship would be that immutable types can be either pure or impure, but mutable types must be impure.

> `reads(everything()); // we use reads(everything()) so implementing methods can still decide whether to be mutable or not`

I think we'll want something tighter than that, since in this case the implementors will have either `reads {}` or `reads this`. Having `reads *` on `Container<T>` means any code using just that interface will not be able to prove nearly as much.

> However, without a @Modifiable @Reference or @Impure, we'd have to make Object the impure one by default. I'm guessing that most commonly you want a pure object, since you only need an impure object to do reference equality and there are quite a few methods in the Object class (which you can use with a pure object). Would that be enough reason to introduce an @Impure ?

I think I lean towards making `Object` the impure one by default in order to avoid needing `@Impure`, because there aren't going to be THAT many references to `Object` in most codebases, so it won't be onerous to have to add `@Pure` to a bunch of them. Most of them will appear in JCL libraries which we'll provide the contracts for ourselves.

I think it would also make sense to support it on type parameters, so `@Pure T` would become `T extends Object` instead of `T extends ModifiableObject`.

keyboardDrummer:
> Having reads * on Container<T> means any code using just that interface will not be able to prove nearly as much.

I don't see an alternative that allows code using Container<T> to prove more. If you replace `reads *` with `reads SomeRepr()` with `SomeRepr` having no body, then it effectively behaves like `reads *`, doesn't it?

PR #292, opened 2025-08-18
Add fp64/double support to JVerify

fabiomadge:
Adds support for Java double type using Dafny's fp64, enabling verification of double-precision floating-point operations with IEEE 754 semantics.

## Key Changes

**Type Translation**
- `double` → `fp64`
- `Double` wrapper → `fp64` (non-nullable)
- `float`/`Float` → Error (not supported yet)

**Method Support**
- `Double.isNaN()` → `fp64.IsNaN()`
- `Double.isInfinite()` → `fp64.IsInfinity()`
- `Double.isFinite()` → `fp64.IsFinite()`

**Operations**
- Int/long to fp64 promotion via `fp64.FromReal()`
- Increment/decrement operators for doubles
- IEEE 754 equality using `fp64.Equal()`

## Testing
- Added comprehensive test: Doubles.java

## Note
Float support will be added when Dafny implements fp32.


verifier/src/main/java/com/aws/jverify/generated/BigFloat.java
```diff
@@ -0,0 +1,13 @@
+package com.aws.jverify.generated;
+
+public class BigFloat {
```
keyboardDrummer: Can you point me to the Dafny branch that this is generated from? It seems that it's not in Dafny master yet.

verifier/src/main/java/com/aws/jverify/verifier/Serializer.java
```diff
@@ -89,6 +90,7 @@ private void serializeValue(Object value, AnnotatedType annotatedType) {
             case Double d -> encoder.writeDouble(d);
             case Boolean b -> encoder.writeBool(b);
             case Character c -> encoder.writeString(new String(new char[] {c}));
+            case BigDec bd -> encoder.writeDouble(bd.getValue());
```
keyboardDrummer: I don't follow. How does a BigDecimal fit into a double? Ah, I guess you should rename `writeDouble` to `writeBigDecimal`

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/BlockCompiler.java
```diff
@@ -166,7 +166,17 @@ private List<Statement> translateVariableDeclaration(IOrigin origin, JCTree.JCVa
                 translatedType, false);
         ConcreteAssignStatement dafnyInitializer = null;
         if (variableDecl.getInitializer() != null) {
-            var rhs = expressionCompiler.toAssignmentRhs(variableDecl.getInitializer(), expressionContext);
+            var initExpr = variableDecl.getInitializer();
+            AssignmentRhs rhs;
+            
+            if (variableDecl.getType().type.hasTag(com.sun.tools.javac.code.TypeTag.DOUBLE) &&
```
keyboardDrummer: I don't follow this yet. Shouldn't the Java compiler insert an implicit cast when you're assigning something that's not a double, to a variable of type double? If it does it seems better to handle that cast. Otherwise, I would imagine it's easy to miss case where you need to add the promotion. What about passing a non-double value to a method call that takes a double?

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/ExpressionCompiler.java
```diff
@@ -611,6 +654,81 @@ public Expression translateMethodInvocation(JCTree.JCMethodInvocation invocation
...
+            var argBindings = invocation.getArguments().stream()
+                    .map(arg -> new ActualBinding(null, toExpr(arg, context.forbidImpure()), false))
+                    .toList();
+
+            return switch (methodName) {
+                case "abs" -> {
+                    if (!argBindings.isEmpty() && invocation.getArguments().get(0).type.getTag() == TypeTag.DOUBLE) {
+                        yield callFp64Method(origin, "Abs", argBindings);
+                    }
+                    reporter.reportError(origin, "notSupported", "Math.abs for integers");
```
keyboardDrummer: Why is it unsupported? Was it not supported (but uninterpreted) before? I think previously the user could still supply a contract for `Math.abs` for integers.

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/ExpressionCompiler.java
```diff
@@ -557,14 +580,34 @@ private Expression translateFieldAccess(JCTree.JCFieldAccess fieldAccess, IOrigi
...
+            var fieldNameStr = fieldAccess.name.toString();
+            return switch (fieldNameStr) {
+                case "NaN" -> fp64Constant(origin, "NaN");
+                case "POSITIVE_INFINITY" -> fp64Constant(origin, "PositiveInfinity");
+                case "NEGATIVE_INFINITY" -> fp64Constant(origin, "NegativeInfinity");
+                case "MAX_VALUE" -> fp64Constant(origin, "MaxValue");
+                case "MIN_VALUE" -> fp64Constant(origin, "MinSubnormal");
+                case "MIN_NORMAL" -> fp64Constant(origin, "MinNormal");
+                default -> {
+                    reporter.reportError(fieldAccess, "notSupported", "Double field " + fieldNameStr);
```
keyboardDrummer: What are the not supported fields? Could you add tests for those? Also, are you sure that we should report an error for them? Can't users still add contracts for them?

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/ExpressionCompiler.java
```diff
@@ -611,6 +654,81 @@ public Expression translateMethodInvocation(JCTree.JCMethodInvocation invocation
...
+                && ownerClass.fullname.contentEquals(Double.class.getName())) {
+            var methodName = methodSymbol.name.toString();
+            
+            Expression result = null;
+            if (methodSymbol.isStatic()) {
+                result = switch (methodName) {
+                    case "isNaN", "isInfinite", "isFinite" -> {
+                        if (invocation.getArguments().isEmpty()) yield null;
+                        var arg = toExpr(invocation.getArguments().get(0), context.forbidImpure());
+                        if (arg instanceof LiteralExpr) arg = new ParensExpression(origin, arg);
```
keyboardDrummer: Why the `ParensExpression`? Isn't that purely useful when printing/parsing, which we're not doing?

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/ExpressionCompiler.java
```diff
@@ -634,15 +752,40 @@ public Expression translateBinary(
...
-        if (leftType.getTag() == TypeTag.FLOAT || leftType.getTag() == TypeTag.DOUBLE) {
-            reporter.reportError(node, "notSupported", "operator " + operator);
-        }
 
         if (opName.equals("+") && leftType.toString().contentEquals(String.class.getName())) {
             var callee = new ExprDotName(origin, left, new Name(origin, "concat"), null);
             return createCall2(origin, callee, Stream.of(right));
         }
 
+        // Promote int/long to double in mixed operations
```
keyboardDrummer: Doesn't Javac add those casts in the Lower.java phase?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MissingContractCompiler.java
```diff
@@ -221,17 +221,24 @@ void visitReference(Symbol symbol, JCTree tree) {
...
             }
-            
+
             if (symbol.owner == symtab.arrayClass) {
                 // Arrays are handled using custom code
                 return;
             }
-            
+
+            // Skip Math and Double methods - they have special fp64 handling in ExpressionCompiler
```
keyboardDrummer: Having Double/Math specific code in MissingContractCompiler, and having code referencing `?static` in ExpressionCompiler, both isn't great because then we're mixing responsibilities. I think ideally we have a registry `NativeSymbols` that maps `VarSymbol` and `MethodSymbol` to functions that build Dafny Expressions, and `MissingContractCompiler` and `MoveStaticMethodsToStaticType` both check that registry and ignore symbols that are in it, so those symbols won't get moved to a static class.

I imagine something like:

```java
class NativeSymbols {
  public void registerField(VarSymbol symbol, Function<IOrigin, Expression> create);

  public void registerMethod(VarSymbol symbol, Function<IOrigin, List<Expression>, Expression create);

  void handleDoubles() {
     var mathClass = symtab.
  }
}
```

And then somewhere in the constructor of ExpressionCompiler or one of the other classes in the dafnygenerator package:
```
    void handleDoubles() {
        var mathClass = syms.getClass(syms.java_base, names.fromString(Math.class.getName()));
        var doubleAbs = mathClass.members().findFirst(names.fromString("abs"), s -> ((Symbol.MethodSymbol)s).params().getFirst().type.getTag() == TypeTag.DOUBLE);
        registerMethod(doubleAbs, (origin, args) -> callFp64Method(origin, "Abs", args));
    }
```

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/ExpressionCompiler.java
```diff
@@ -787,6 +930,102 @@ public static ApplySuffix createCall2(IOrigin origin, Expression callee, Stream<
                 createBindings(bindingList), null);
     }
 
+    public Expression translateFp64Literal(IOrigin origin, double value) {
+        if (Double.isNaN(value)) {
+            return fp64Constant(origin, "NaN");
+        }
+        if (value == Double.POSITIVE_INFINITY) {
```
keyboardDrummer: I'm surprised by some of this code. Can someone type in +infinity or something like that for a Java literal?

keyboardDrummer (commented):
Looks really good! I have a few questions and one significant request.

verifier/src/main/java/com/aws/jverify/generated/BigFloat.java
fabiomadge: It's in master via [boogie](https://github.com/boogie-org/boogie/blob/master/Source/BaseTypes/BigFloat.cs).

keyboardDrummer: I didn't find it in [Syntax.schema-cs in master.](https://github.com/dafny-lang/dafny/blob/master/Source/Scripts/Syntax.cs-schema) though. That's where the Java classes are generated from

fabiomadge: I added it on top of https://github.com/dafny-lang/dafny/pull/6369, as it is still unmerged, but likely needed.

fabiomadge: Actually, you are right. I added this by mistake, it's not generated at all.

verifier/src/main/java/com/aws/jverify/verifier/Serializer.java
fabiomadge: Got rid of it altogether

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/BlockCompiler.java
fabiomadge: Javac doesn't insert explicit JCTypeCast nodes for widening conversions - they remain implicit in the AST. Added FloatingPointCastInserter preprocessing pass to handle all cases: variable declarations, assignments, method calls, returns, ternaries, array initializers, and binary operations.

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/ExpressionCompiler.java
fabiomadge: Fixed. Now only registers the double variant. Integer variants fall through to normal method handling, allowing user contracts.

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/ExpressionCompiler.java
fabiomadge: Removed the error-reporting default case. Now only handles the 6 fp64 constants (NaN, POSITIVE_INFINITY, NEGATIVE_INFINITY, MAX_VALUE, MIN_VALUE, MIN_NORMAL). Other fields fall through to normal handling.

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/ExpressionCompiler.java
fabiomadge: Removed. It was unnecessary.

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/ExpressionCompiler.java
fabiomadge: Lower.java desugars high-level constructs (foreach, try-with-resources, etc.) but doesn't insert widening primitive casts. Those remain implicit in javac's AST - the JVM handles them via bytecode instructions (i2d, l2d). Our FloatingPointCastInserter runs after attribution but before Lower, making implicit conversions explicit for Dafny translation.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MissingContractCompiler.java
fabiomadge: Implemented. Created NativeSymbols with MethodTranslator/FieldTranslator interfaces. ExpressionCompiler registers symbols in constructor. MissingContractCompiler uses isRegistered() to skip native symbols.

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/ExpressionCompiler.java
fabiomadge: Removed that code. Javac doesn't constant-fold 1.0/0.0 - it remains a binary expression in the AST. Only kept MAX_VALUE, MIN_VALUE, MIN_NORMAL constant checks.

verifier/src/main/java/com/aws/jverify/generated/BigFloat.java
keyboardDrummer: > I added it on top of https://github.com/dafny-lang/dafny/pull/6369

Great, thanks

verifier/src/main/java/com/aws/jverify/verifier/Encoder.java
```diff
@@ -31,7 +31,8 @@ public void writeNullable(boolean isNull) {
 
     @Override
     public void writeDouble(double value) {
-        writer.append(value);
+        var bigDecimal = new java.math.BigDecimal(Double.toString(value));
```
keyboardDrummer, line 34: What's the reason for this change?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MissingContractCompiler.java
```diff
@@ -221,17 +224,22 @@ void visitReference(Symbol symbol, JCTree tree) {
...
-            
+
             if (symbol.owner == symtab.arrayClass) {
                 // Arrays are handled using custom code
                 return;
             }
-            
+
+            // Skip native symbols (Math/Double methods) - they have special handling in ExpressionCompiler
+            if (nativeSymbols.isRegistered(symbol)) {
```
keyboardDrummer, line 234: How come we don't need something similar in `MoveStaticMethodsToStaticType` ?

keyboardDrummer (dismissed):
Super! Love the latest changes. I'm frustrated that we need the `FloatingPointCastInserter`, but it seems there's no other way.

verifier/src/main/java/com/aws/jverify/verifier/Encoder.java
fabiomadge: BigDec.FromString supports scientific notation, but the Java serializer (writer.append(value)) produces a format that Dafny's parser rejects. Converting via BigDecimal.toPlainString() produces the correct format that Dafny accepts. Tests fail without this conversion.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MissingContractCompiler.java
fabiomadge: Good catch. Added nativeSymbols.isRegistered() check to MoveStaticMethodsToStaticType - native symbols (Math/Double methods and fields) should stay in their original classes, not be moved to a ?static class.

keyboardDrummer: Shouldn't there be a test for this then?

keyboardDrummer (dismissed):

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MissingContractCompiler.java
fabiomadge: You're right - I added that check by mistake. MoveStaticMethodsToStaticType only processes user code compilation units, and the registry only contains JDK symbols (Math/Double), so the check was unreachable. Removed it. The check in MissingContractCompiler is necessary though - it processes symbols referenced in user code and prevents generating contracts for native methods.

keyboardDrummer (approved):

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MissingContractCompiler.java
keyboardDrummer: Thanks for the explanation, that's very helpful :)

olivier-aws:
[celebrate]     Bouissou, Olivier reacted to your message:
________________________________

PR #293, opened 2025-08-18
Do not make constant fields ghost

keyboardDrummer:
### What was changed?
- Do not make constant fields ghost, since generally they're not.
- Let some of the builtin contracts define a pure body
- Resolve resulting Dafny error that relate to equality not being defined, by using more JVerify.Sequence instead of Set/Map

### How has this been tested?
- Uses existing tests


fabiomadge:

PR #296, opened 2025-08-18
Support static methods nested inside polymorphic types

keyboardDrummer:
### What was changed?
- Support static methods nested inside polymorphic types

### How has this been tested?
- Added a test `PolymorphicStaticsVerification`


verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MoveStaticMethodsToStaticType.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+        
+    }
+
+    class ReferenceUpdater extends TreeTranslator {
+        Symbol.@Nullable ClassSymbol staticContext = null;
+        
+        @Override
+        public void visitSelect(JCTree.JCFieldAccess tree) {
+            if (tree.sym instanceof Symbol.MethodSymbol methodSymbol && staticMethodSymbols.contains(methodSymbol)) {
```
ssomayyajula: Why collect all of the `staticMethodSymbols` ahead of time rather than checking that `methodSymbol` has `STATIC` among its `getModifiers()`?

ssomayyajula (commented):
Just have that one question, otherwise looks good.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MoveStaticMethodsToStaticType.java
keyboardDrummer: That would probably work as well, but this is more defensive since we'll only update references for the methods that we actually moved. In case somehow we didn't touch all static methods this would be safer. Would you prefer I do what you suggest?

ssomayyajula (dismissed):
Oh I get it now. Nah, keep it as it is. Approved!

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MoveStaticMethodsToStaticType.java
ssomayyajula: No worries, SGTM.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ImmutableTypeCompiler.java
```diff
@@ -37,7 +37,6 @@ public TopLevelDeclWithMembers translate(Symbol.ClassSymbol classSymbol, JCTree.
         Symbol.ClassSymbol currentTypeSymbol = typeDeclarationCompiler.getCurrentTypeSymbol(classDecl.sym);
         var traits = currentTypeSymbol
                 .getInterfaces().stream()
-                .filter(type -> JavaToDafnyCompiler.typeHasSource(compiler.index, type.tsym))
```
oyendrila-dobe: why do we not need this check anymore?

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/lambdas/PolymorphicLambdas.java
```diff
@@ -45,7 +61,25 @@ interface MyConsumer<T> {
         int consume(T value);
     }
 
-    interface MyProducer<T> {
```
oyendrila-dobe: why remove the interface test ?

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/nestedClasses/NestedStaticClass.java
```diff
@@ -5,7 +5,7 @@
 
 import static com.aws.jverify.JVerify.check;
 
-@JVerifyTest(exitCode = 4, dafnyVerified = 4, dafnyErrors = 1)
```
oyendrila-dobe, line 8: Are we only mentioning exit codes if its non-zero?

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/InstrumentLower.java
```diff
@@ -41,6 +41,8 @@ public static void installModification() {
                             .method(named("access").
                                     and(takesArgument(0, is(Symbol.class))))
                                     .intercept(MethodDelegation.to(LowerInterceptor.class))
+                            .method(named("makeOuterThisVarSymbol"))
```
oyendrila-dobe: i wasn't able to connect this change to anything else

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/JavaFrontEnd.java
```diff
@@ -99,9 +100,8 @@ public Set<JCTree.JCCompilationUnit> parseResolveAndDesugarJava(VerifierOptions
          * PROCESS(3),
          * ATTR(4),
          * FLOW(5),
-         * SUBSTITUTE,
-         * UNLAMBDA(8),
-         * UNSUBSTITUTE
+         * Method references to lambdas,
```
oyendrila-dobe: What does this do and why doesn't it follow the pattern of the existing entries in the list?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ImmutableTypeCompiler.java
keyboardDrummer: This check was to prevent the Dafny code from referring to types for which we weren't generating a Dafny equivalent. However, now (already changed in a previous PR) we're creating Dafny types for any Java type that's referred to, even if no contract is defined for it.

I'm not sure why this line is being deleted in this PR but it's obsolete in any case.

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/nestedClasses/NestedStaticClass.java
keyboardDrummer: Yes, zero is the default value

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/lambdas/PolymorphicLambdas.java
keyboardDrummer: Instead of `MyProducer` I'm using `Supplier` now, which is a standard library type.

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/InstrumentLower.java
keyboardDrummer: `Lower` is a pass from the Java compiler, which contains this method. You can find it here: https://github.com/openjdk/jdk/blob/f0498c2aed761d4023917bc9cd1f852a02ce977a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Lower.java#L1460

Note the call to `erasure(types)`, which takes a regular possibly applied Java type and removes all type arguments from it. This type erasure is useful in the Java compiler which wants to generate Java bytecode without any applied types, but since we want to generate Dafny we need the types to stay applied. This change makes it so that no type erasure occurs during execution of `makeOuterThisVarSymbol`.

I don't recall why this change is needed in this PR, but since this PR is about polymorphic types I'm not surprised it makes a change so that Lower.java keeps more applied polymorphic types.

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/JavaFrontEnd.java
keyboardDrummer: Updated the comment to make it more clear.

oyendrila-dobe (approved):
LGTM

PR #297, opened 2025-08-19
Added contract for Optional Class

### LINKED ISSUE #272: Add contracts for the Optional class
Methods used in customer B's prototype: 
- get
- ifPresent
- map
- orElse

oyendrila-dobe:
### What was changed?
- Added contract for a subset of methods in Optional class 
- Modified relevant line number for expected error in `Collections.test`

### How has this been tested?
- Added relevant test cases in `verifier/tests/Optionals.java`


verifier/src/main/resources/builtin-contracts.java
```diff
@@ -489,3 +491,53 @@ public int signum() {
...
+        throw new ContractException();
+    }
+
+    static <T> OptionalContract<T> empty() {
+        postcondition((OptionalContract<T> o) -> !o.isSet);
+        throw new ContractException();
+    }
+
+    //Currently the same as of because we are not able to handle null cases
+    static <T> OptionalContract<T> ofNullable(T val) {
```
keyboardDrummer: Probably better to comment out this contract since it can't effectively be used.

verifier/src/main/resources/builtin-contracts.java
```diff
@@ -489,3 +491,53 @@ public int signum() {
...
 }
+
+@Contract(value = Optional.class, immutable = true)
+class OptionalContract<T> {
+
+    private T value;
+    private boolean isSet;
+
+    //Throws a NullPointerException if the val is null
+    static <T> OptionalContract<T> of(T val) {
```
keyboardDrummer: I prefer `value` over `val`. `val` makes me think of the immutable variable declaration from `scala`. In general, try to avoid acronyms.

verifier/src/main/resources/builtin-contracts.java
```diff
@@ -489,3 +491,53 @@ public int signum() {
     }
 
 }
+
+@Contract(value = Optional.class, immutable = true)
+class OptionalContract<T> {
```
keyboardDrummer, line 528: A few `@Pure` annotations are missing on methods

verifier/src/main/resources/builtin-contracts.java
keyboardDrummer: Would you be interested in enabling this as well?

I think on the JVerify side, we would write `@Pure static <T extends @Modifiable Object> OptionalContract<T> ofNullable(T val) { return val == null ? : ... : ... } `

which in Dafny should translate to:

```dafny
function ofNullable<T extends ModifiableObject>(T val): Optional<T> {
  ...
}

trait ModifiableObject extends Object, object
```

A remaining problem is then to translate the `==` in the body. If you translate that to `val == null` then Dafny complains, since it isn't very smart when it comes to constraints on type parameters. I think because of the `extends ModifiableObject`, Dafny should be able to know that `T` supports equality, and then `null` should be interpreted as a `ModifiableObject`. However,  Dafny only uses `extends ModifiableObject` to verify type tests and conversions, but nothing else. Given that, here's a workaround: if you translate it the `==` comparison to `(val as ModifiableObject?) == null` in Dafny then it's fine.

`==` in Java is always reference equality, and everything that has reference equality should inherit from `ModifiableObject`, so in the translation I think you could do something like: if the type of the operand of `==` is a type parameter, then add the conversion ` as ModifiableObject?` to it.

keyboardDrummer (dismissed):
Awesome! Left a few comments

oyendrila-dobe:
TODO: yet to add contract for `map`. Will add it in a later PR along with the null checks modification

verifier/src/main/resources/builtin-contracts.java
```diff
@@ -488,4 +490,61 @@ public int signum() {
...
+}
+
+@Contract(value = Optional.class, immutable = true)
+class OptionalContract<T> {
+
+    private T value;
+    private boolean isSet;
+
+    //Throws a NullPointerException if the value is null
+    //TODO: add precondition for null
```
keyboardDrummer, line 533: What happens if you pass in `null` and run JVerify on this?

verifier/src/main/resources/builtin-contracts.java
```diff
@@ -488,4 +490,61 @@ public int signum() {
...
+        throw new ContractException();
+    }
+
+    @Pure
+    T orElse(T value) {
+        postcondition((T b) -> isPresent()? b == this.value : b == value);
+        throw new ContractException();
+    }
+
+    // Throws a NoSuchElementException if this is an empty object
```
keyboardDrummer: I don't understand the comment. JVerify prevents the exception from ever being thrown right?

verifier/src/main/resources/builtin-contracts.java
oyendrila-dobe: Currently, it complains that `type of 'null' is a reference type, but it is used as Object` 
(code used: `Optional.of( null)`)

verifier/src/main/resources/builtin-contracts.java
oyendrila-dobe: My bad. This is related to the actual behaviour of the `Optional.get` wrt Java semantics not Jverify. Removing this and the comment on line 501 to avoid confusion

verifier/src/main/resources/builtin-contracts.java
keyboardDrummer: Then I don't understand what the TODO would entail. It seems like this method's contract is done. What you could do is add a test that checks the behavior you describe above. 

I think later we should remap the error message to "type of 'null' is an impure type, but it is used as a pure Object", but you can leave that out of this PR.

keyboardDrummer (approved):

PR #298, opened 2025-08-20
Handle `@Contract` as a separate pass

keyboardDrummer:
### What was changed?
- Handle `@Contract` as a separate pass, so all other passes don't need to be aware of `@Contract` semantics. 
  - I started this PR because I was working on https://github.com/aws/jverify/pull/296 and realized there too I had to consider contracts. I think with this refactoring, lots of future PRs will be easier. 
  - The separate pass tries to make it seem as if all contracts defined by a `@Contract` class, were defined on the contractee class. It inserts `@Verify(false)` annotations so that the bodies of those methods does not need to be checked, so the `throw new ContractException()` is ignored.
- Move some reporting code from `JavaToDafnyCompiler` to a new class `Reporter`

### How has this been tested?
- Uses existing tests


verifier/src/test/java/com/aws/jverify/verifier/tests/jcl/PrimitiveAndBoxedTypesVerification.java
```diff
@@ -1,21 +1,27 @@
 package com.aws.jverify.verifier.tests.jcl;
 
+import com.aws.jverify.Nullable;
 import com.aws.jverify.Pure;
 import com.aws.jverify.testengine.JVerifyTest;
 
 import static com.aws.jverify.JVerify.check;
 
-@JVerifyTest(exitCode = 4, dafnyVerified = 3, dafnyErrors = 1, useBuiltinContracts = true)
+@JVerifyTest(exitCode = 4, dafnyVerified = 4, dafnyErrors = 1, useBuiltinContracts = true)
```
oyendrila-dobe, line 9: Is it the case that we don't anymore need exitcode if we are not using builtIn Contracts?

oyendrila-dobe (approved):
LGTM

verifier/src/test/java/com/aws/jverify/verifier/tests/jcl/PrimitiveAndBoxedTypesVerification.java
keyboardDrummer: Based on the change, which does not affect exitCode, I don't understand your question. Were you looking at an outdated diff? In any case, the answer is no, exit code and builtin contracts are unrelated.

oyendrila-dobe: I was trying to understand where are we not using exit codes anymore. Combining these changes and the ones in another PR, i guess we can skip exit codes if it is zero and mention it otherwise.

PR #299, opened 2025-08-25
Add missing contract members

keyboardDrummer:
### What was changed?
- Fix a name collision issue that could occur when child and parent class had same named methods with different signatures
- Automatically add missing members. When members are missing this creates warnings, but the code can still be verified.
- Change order in the GH built yml so the test suite runs before an integration test (which would prevent the test suite from running, reducing the amount of feedback).
- Move code for adding missing types to a separate class
- Remove field `typeForWhichCurrentClassIsDefiningContract` since it was no longer used

### How has this been tested?
- Add a case to `AvoidNameCollisionsTest`
- Added the test `MissingContracts`


verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/JVerifyMaker.java
```diff
@@ -0,0 +1,42 @@
...
+import com.aws.jverify.Verify;
+import com.sun.tools.javac.code.Symtab;
+import com.sun.tools.javac.model.JavacElements;
+import com.sun.tools.javac.tree.JCTree;
+import com.sun.tools.javac.tree.TreeMaker;
+import com.sun.tools.javac.util.Context;
+import com.sun.tools.javac.util.List;
+import com.sun.tools.javac.util.Names;
+
+public class JVerifyMaker {
```
ssomayyajula, line 12: Are you naming this "JVerifyMaker" because it's meant to store methods for creating any kind of resolved node, not just `{:verify false}`?

ssomayyajula (dismissed):
Trusting the tests and the small fixes focusing on `MissingContractCompiler`, which from what I can tell, adds `{:verify false}` to members that are not annotated and verified. Approved!

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/JVerifyMaker.java
keyboardDrummer: Yes, I've named it after javac's `TreeMaker`, which is a class for creating Java nodes. This one is for creating JVerify specific nodes.

Note that `{:verify false}` is a Dafny thing that JVerify is not using at all. `@Verify(false)` is implemented using bodyless members at the moment

PR #302, opened 2025-08-26
Better support for inferred generics

keyboardDrummer:
### What was changed?
- Improve behavior when constructing objects with implicit generic arguments

### How has this been tested?
- Added a new test file

### DIFF
--- a/verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ImmutableTypeCompiler.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ImmutableTypeCompiler.java
@@ -222,11 +222,7 @@ public class ImmutableTypeCompiler {
             typeArgs = compiler.getOwnAndEnclosedTypeParameters(newClass.type.tsym).map(
                     tp -> compiler.translateType(tp.type, compiler.toOrigin(tp))).collect(Collectors.toList());
         }
-        
-        typeArgs.addAll(newClass.typeargs.map(expressionCompiler.compiler::translateType));
-        if (newClass.clazz instanceof JCTree.JCTypeApply typeApply) {
-            typeArgs.addAll(typeApply.arguments.map(expressionCompiler.compiler::translateType));
-        }
+        typeArgs.addAll(newClass.type.getTypeArguments().map(t -> expressionCompiler.compiler.translateType(t, origin)));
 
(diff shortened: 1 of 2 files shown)


verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ImmutableTypeCompiler.java
```diff
@@ -222,11 +222,7 @@ public static Expression translateNewRecord(ExpressionCompiler expressionCompile
             typeArgs = compiler.getOwnAndEnclosedTypeParameters(newClass.type.tsym).map(
                     tp -> compiler.translateType(tp.type, compiler.toOrigin(tp))).collect(Collectors.toList());
         }
-        
-        typeArgs.addAll(newClass.typeargs.map(expressionCompiler.compiler::translateType));
-        if (newClass.clazz instanceof JCTree.JCTypeApply typeApply) {
-            typeArgs.addAll(typeApply.arguments.map(expressionCompiler.compiler::translateType));
-        }
+        typeArgs.addAll(newClass.type.getTypeArguments().map(t -> expressionCompiler.compiler.translateType(t, origin)));
```
oyendrila-dobe, line 225: Just curious: The static class JCNewClass (type of newClass) seems to use `type` which is a non-static member of the enclosing abstract class. How are we able to use the `.type` on `newClass`?

oyendrila-dobe (approved):
LGTM

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ImmutableTypeCompiler.java
keyboardDrummer: If you nest a class inside another class in Java, and do not put the `static` keyword in front of the nested class, then the nested one gets access to all instance fields of the outer class. At runtime it maintains a reference to the enclosing class. Could this feature explain what you're seeing?

oyendrila-dobe: That makes sense. However, the nested class here is marked with `static` which is why i was not expecting it to be able to refer to the `non-static` fields in the enclosing class. 
Turns out `JCNewClass` extends classes that ultimately extends JCTree

PR #311, opened 2025-08-28
Improve `@Verify` annotation and more

keyboardDrummer:
### What was changed?
- Refactoring: move code from the incorrectly named `OverrideFinder` into `ExternalContractCompiler`, and simplify it by using more from the javac class `Types`.
- Pureness of methods is now always inherited from base methods
- Refactoring: "lambda to local class" phase is now agnostic to pureness. Previously it copied it from the SAM method, but in cases of external contracts that did not work because external contracts had not been compiled yet. Lambda remains the first phase done after type checking.
- Added a new phase "verify annotation" that comes after "contract annotation" and before "missing contracts", which deletes any implementations that do not need to be verified according to `@Verify` annotations. This means you'll never get warnings/errors from the implementations of unverified code. This new phase means `JavaToDafnyCompiler` and `TypeDeclarationCompiler` are now simpler because they become agnostic to `@Verify` 
- Refactoring: Added a new "extract contracts from methods and loops" phase. The extraction is done by giving each method body two block statements, the first containing the contract and the second containing the implementation. This phase occurs right after compiling lambdas. The purpose of this phase is to prevent having to figure out where the contract statements are more than once, since both "verify annotation" and Dafny generation need to know that.

### How has this been tested?
- Add test for an inherited pure method that's not verified.
- Add test to see if pure on methods is inherited
- Add test to see if Lambda is pure when the SAM gets its pureness from an external contract
- Add a test to check that we can't rely on the body of pure unverified methods
- Add test to check that inside a `@Verify(false)` we won't emit JVerify warnings/errors


verifier/src/main/java/com/aws/jverify/verifier/compiler/BlockCompiler.java
```diff
@@ -13,12 +13,14 @@
 public class BlockCompiler {
 
     public final JavaToDafnyCompiler compiler;
+    MethodOrLoopContractCompiler methodOrLoopContractCompiler;
```
oyendrila-dobe, line 16: Why isn't this compiler final as well?

verifier/src/main/java/com/aws/jverify/verifier/compiler/TypeDeclarationCompiler.java
```diff
@@ -293,7 +275,7 @@ public Field translateField(JCTree.JCVariableDecl variableDecl) {
     private MethodOrConstructor translateImpureMethod(JCTree.JCMethodDecl method,
                                                       boolean shouldVerify,
                                                       MethodOrLoopContract contract,
-                                                      List<JCTree.JCStatement> postHeader) {
+                                                      com.sun.tools.javac.util.List<JCTree.JCStatement> postHeader) {
```
oyendrila-dobe, line 278: Is there a name clash we are avoiding by not importing this class?

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/JavaFrontEnd.java
```diff
@@ -170,27 +176,39 @@ public void finished(TaskEvent e) {
...
+        for(var unit : result) {
+            Env<AttrContext> env = enter.getTopLevelEnv(unit);
+            for(var def : unit.defs) {
+                index.index(env, def);
+            }
+        }
+        return result;
+    }
+    
+    private Set<JCTree.JCCompilationUnit> unlambda(Set<JCTree.JCCompilationUnit> envs) {;
```
oyendrila-dobe: why do we have this dangling `;` ?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ExternalContractCompiler.java
```diff
@@ -159,15 +169,16 @@ private void handleSourceContract(JCTree.JCClassDecl classDecl, Map<String, JCTr
...
+                        var baseSource = (JCTree.JCMethodDecl)index.getTree(contracteeMethod);
                         if (baseSource.getBody() != null) {
                             reporter.reportError(methodDecl, "internalAndExternalContractForMethod", methodSymbol.name.toString());
                         } else {
+                            contractSymbolToContractee.put(methodSymbol, contracteeMethod);
                             baseSource.body = methodDecl.body;
-                            baseSource.mods.annotations = methodDecl.mods.annotations;
-                            baseSource.mods.annotations = baseSource.mods.annotations.append(verifyMaker.getVerifyFalseAnnotation());
+                            baseSource.mods.annotations = methodDecl.mods.annotations.append(jverifyUtils.getVerifyFalseAnnotation());
+                            jverifyUtils.addVerifyFalseToMethodSymbol(methodSymbol, contracteeMethod);
```
oyendrila-dobe, line 181: Clarification: why are we adding the verifyFalse both on the contractee and the base source?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ExternalContractCompiler.java
```diff
@@ -217,17 +207,85 @@ private void handleLibraryContract(JCTree.JCClassDecl classDecl, Symbol.ClassSym
...
+
+            updateLibraryContractAnnotations(methodDecl, baseMethod, methodSymbol);
+        }
+
+        private void updateLibraryContractAnnotations(JCTree.JCMethodDecl methodDecl, 
+                                                      Symbol.MethodSymbol baseMethod, 
+                                                      Symbol.MethodSymbol methodSymbol) {
+
+            ListBuffer<Attribute.Compound> newAnnotations = new ListBuffer<>();
+            // newAnnotations.addAll(baseMethod.getAnnotationMirrors());
```
oyendrila-dobe: Are we keeping this for any specific reason?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MethodOrLoopContractCompiler.java
```diff
@@ -1,71 +1,177 @@
...
+import com.sun.tools.javac.util.List;
 
 import java.util.ArrayList;
-import java.util.List;
+import java.util.Set;
 
 /*
 Extracts contracts from constructor, method or loop bodies
+Does not support lambdas.
+TODO: could be slightly simplified by compiling DoWhile loops first
```
oyendrila-dobe, line 23: I'm assuming this is for a later PR?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MethodOrLoopContractCompiler.java
```diff
@@ -82,36 +188,60 @@ private static JCTree.JCStatement getSuperOrThis(List<JCTree.JCStatement> statem
...
+            return false;
+        }
+        
+        var methodName = jverifyMethod.getQualifiedName().toString();
+        switch (methodName) {
+            case "check" -> {
+                // not a header method, so stop here
+                return false;
+            }
+            case Common.PRECONDITION, "postcondition", "invariant", "decreases", "reads", "modifies" -> {
```
oyendrila-dobe, line 222: Why do we treat precondition specially?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MethodOrLoopContractCompiler.java
```diff
@@ -82,36 +188,60 @@ private static JCTree.JCStatement getSuperOrThis(List<JCTree.JCStatement> statem
         return superOrThis;
     }
 
-    private List<JCTree.JCStatement> addHeaderContracts(List<JCTree.JCStatement> statements,
-                                                        MethodOrLoopContract contract) {
-        int headerContracts;
+    private java.util.List<JCTree.JCStatement> addHeaderContracts(
```
oyendrila-dobe, line 191: Why did we remove the List import and add the package to all invocations of List?

oyendrila-dobe (dismissed):
Looks good

verifier/src/main/java/com/aws/jverify/verifier/compiler/TypeDeclarationCompiler.java
keyboardDrummer: Yes. There's also `java.util.List`

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ExternalContractCompiler.java
keyboardDrummer: javac translates a program to both `JCTree` nodes (syntactic), and `Symbol` nodes (semantic). The first line creates the syntactic `@Verify(false)` JCTree, and the second creates its `Symbol`

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MethodOrLoopContractCompiler.java
keyboardDrummer: yes

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MethodOrLoopContractCompiler.java
keyboardDrummer: Because there's 18 usages of `com.sun.tools.javac.util.List` and 9 of `java.util.List`, so we're qualifying the second and importing the first.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MethodOrLoopContractCompiler.java
keyboardDrummer: It's referenced more often

PR #315, opened 2025-09-01
String contract

### LINKED ISSUE #307: Support casting String to any of its base types, like CharSequence
Support

```java
var s: String = "hello";
java.time.Instant.parse(s)
```

which currently gives: `Error: incorrect argument type for function parameter 'arg0' (expected CharSequence, found DString)`

keyboardDrummer:
Fixes https://github.com/aws/jverify/issues/307

### What was changed?
- Make `object-contract.java` be treated as a library by Dafny, improving performance
- Fix a bug in `getBodylessMethods`
- Cleanup `record MapEntry` from `builtin-contracts.java`
- Remove custom support for Strings and use `@Contract` mechanism instead. This enables downcasting `String`s to its basetypes, which the previous custom code did not allow.
- Add a JVerify type `CharJSequence` since Java does not allow `Sequence<char>`. Maybe if we support pure arrays such as `@Pure char[]` and translate that to a Dafny `seq<char16>` that could replace `CharJSequence`.
- Add JVerify function `concat`

### How has this been tested?
- Add a test for downscasting String to `CharSequence`, one of its basetypes


verifier/src/main/resources/object-contract.java
```diff
@@ -20,12 +21,88 @@ public boolean equals(Object obj) {
...
+                : chars.get(result) == ch && JVerify.forall((int j) -> JVerify.implies(0 <= j && j < result, chars.get(j) != ch))
+        );
+        throw new ContractException();
+    }
+
+    @Pure
+    public int length() {
+        postcondition((int result) -> result == chars.size());
+        throw new ContractException();
+    }
```
olivier-aws, line 61: Why not `return chars.size()` similar to the `isEmpty` method?

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/records/RecordsVerified.java
```diff
@@ -122,7 +122,7 @@ record Pair<A, B>(A a, B b) {}
 record UnusedTypeParameter<A, B>(int x) {}
 
 /** Recursion */
-record BasicConsList(String head, @Nullable Wrapper<BasicConsList> tail) {}
+record BasicConsList(int head, @Nullable Wrapper<BasicConsList> tail) {}
```
olivier-aws, line 125: Why did you need to change from `String` to `int` there?

olivier-aws (approved):
LGTM. Just two questions.

verifier/src/main/resources/object-contract.java
keyboardDrummer: I tried that, but `chars.size()` is a Dafny `int` while `length` needs to return a Dafny `int32`. The current implementation effectively assumes that `chars.size()` is an int32.

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/records/RecordsVerified.java
keyboardDrummer: That was needed only in a previous version of this PR, but I also prefer it like this because now this test, which is not about String, does not depend on it.

PR #318, opened 2025-09-02
Array compiler

keyboardDrummer:
### What was changed?
- Refactoring: reduce code duplication in the Dafny generator that related to generating calls.
- Use ghost code everywhere in the generated Dafny, this makes the generated Dafny code more permissive while remaining sound. Some things that are unlocked is equality on all types and type tests on pure types. A problem with this change is that Dafny arrays can no longer be used because their data field is not ghost, and non-ghost fields can not be assigned from ghost code.
- Stop using Dafny arrays but instead use a JVerify defined type `GhostArray<T>`. Caveats: 
  - This change pushes up the "verified count" number for all the tests, which is bad for performance. We should find a way to avoid Dafny verifying the contracts of builtins code, but I've left that out of scope for this PR. 
  - There is a verification completeness issue related to using `GhostArray<T>` that I've filed here: https://github.com/dafny-lang/dafny/issues/6356 . It would be nice to understand that better.
- Refactoring: replace the array specific code in the Dafny generator with a Java source transformation, located in a new file `ArrayCompiler`.

### How has this been tested?
- Tests were adapted to the changes, some messages were changed, but conceptually the tests have remained the same.


robin-aws:
> Use ghost code everywhere in the generated Dafny, this makes the generated Dafny code more permissive while remaining sound. 

I think this is a good idea and was already tempted to do this myself earlier. I'm still a bit worried that there are cases where this is unsound though: at least in older versions of Dafny, there were checks that were only run when the code was translated/executed, such as checking for missing bodies. I know we moved a lot of them into the verifier phase but I'd want to double check there aren't some left over that we need here as well. I added a bullet to https://github.com/aws/jverify/issues/234 to revisit this.

> This change pushes up the "verified count" number for all the tests, which is bad for performance.

I don't think we should over-index on that exact metric. The cost of the individual VCs matters a lot more than the raw number of them. That's part of why I'm still not in favor of having tests assert them explicitly (but I'm at least less frustrated now that I added the `JVERIFY_UPDATE_TEST_ANNOTATIONS` utility, and hopefully you were able to use that too :)

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/JavaFrontEnd.java
```diff
@@ -143,16 +143,18 @@ public void finished(TaskEvent e) {
...
+                                    arrayCompiler.transform(
                                     unsuspend(lower(suspend(
                                                 missingContractCompiler.compile(
                                                         verifyAnnotationCompiler.transform(
                                                                 externalContractCompiler.apply(
                                                         newMethodContractCompiler.transform(
                                                                 unlambda(
                                                         toUnits(compiler.flow(compiler.attribute(todo)))
-                    ))))))))));
+                    )))))))))));
```
robin-aws, line 157: Getting close to Lisp-level nesting here :D

Would like to clean up the indenting above but not in scope for this PR - maybe there's a way of ensuring GC of the intermediate trees between phases without this lexical mess I copied from the javac code.

verifier/src/main/resources/object-contract.java
```diff
@@ -1,11 +1,50 @@
...
+//    @Pure
+//    @Override
+//    public boolean equals(Object obj)
+//    {
+//        return JVerify.jequals(this, obj);
+//    }
+}
+
+@Verify(false)
+class GhostArray<TArrayElement> {
```
robin-aws: Two suggestions:

1. I would name this `JArray`, since the important thing is that it is used to represent Java arrays in Dafny. The ghostness is pretty orthogonal.
2. Wouldn't it be simpler to define the image of this directly in additional.dfy instead, since there are no explicit references to it from Java code?

verifier/src/main/resources/object-contract.java
```diff
@@ -1,11 +1,50 @@
...
+    public static <TCreateArrayElement> GhostArray<TCreateArrayElement> create(int size) {
+        precondition(size >= 0);
+        //noinspection ConstantValue
+        postcondition((GhostArray<TCreateArrayElement> r) -> r.size() == size && fresh(r));
+        throw new ContractException();
+    }
+
+    @Pure
+    public @Nat int size() { // TODO rename to length?
```
robin-aws: Yeah that would be slightly better, since that's the name Java uses (i.e. `a.length`)

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/ArraysVerification.java
```diff
@@ -35,44 +33,26 @@ static void intArrayOfSize10() {
...
+        a[0] = zeroPoint;
         for (int i = 1; i < a.length; i++) {
             modifies(a);
             // We want a better invariant like
             // invariant(Forall((Integeer j) -> Implies(0<=j && j<i,a[i]!=null && a[i].getA()==i)));
             // This does not work now as we cannot pass a non final variable in a lambda
             invariant(a[0] != null);
             invariant(a[0].getA()==1);
-            a[i] = new Point(i,i+1);
+            invariant(a[0] == zeroPoint); // should not be necessary
```
robin-aws: Why shouldn't it be necessary? Yes the assignment is right there immediately before the loop, but the loop body does modify `a` and there's nothing specifying it never modifies `a[0]`.

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/generics/PolymorphismDafnyResolutionErrors2.java
```diff
@@ -3,14 +3,13 @@
...
-@JVerifyTest(exitCode = 22)
+@JVerifyTest(dafnyVerified = 11, dafnyErrors = 0)
 public class PolymorphismDafnyResolutionErrors2 {
     
     public static void valueObjectHasNoEquality() {
         // Annotations on method parameters are broken ATM, so these are locals
         Object a = new Object();
         @Modifiable Object b = new Object();
         var c = a == (@Modifiable Object)b;
-//              ^ Error: == can only be applied to expressions of types that support equality (got Object)
```
robin-aws, line 14: Isn't this a regression? Yes Dafny will allow `==` check in ghost code, but it's not consistent with Java semantics so it shouldn't.

verifier/src/main/java/com/aws/jverify/verifier/compiler/TypeDeclarationCompiler.java
```diff
@@ -173,16 +173,20 @@ public TraitDecl getTraitDecl(IOrigin origin, Name name,
                 typeParameters, members, superTraits, false);
     }
 
-    private static TypeParameter getTypeParameter(IOrigin origin, com.sun.tools.javac.util.List<@Nullable Type> bounds, Name name) {
-        if (bounds.isEmpty()) {
+    private static TypeParameter getTypeParameter(IOrigin origin, 
+                                                  com.sun.tools.javac.util.List<@Nullable Type> bounds, 
+                                                  Name name) {
+        boolean arrayParameter = name.getValue().equals("TArrayElement");
+        boolean createArrayParameter = name.getValue().equals("TCreateArrayElement");
```
robin-aws, line 180: Oof, this seems like even more reason to define what you want in additional.dfy instead.

robin-aws:
> A problem with this change is that Dafny arrays can no longer be used because their data field is not ghost, and non-ghost fields can not be assigned from ghost code.

That is unfortunate and feels like something we should address in Dafny. It's weirdly inconsistent to allow ghost code to instantiate a ghost array value but then forbid writing to it!

robin-aws (changes requested):
I think this is a good change to make, a couple of points to address:

1. Can you comment on specifically what additional desirable permissiveness we get from this change? I may have missed it but it looks like the tests were just refactored to still pass. Just a comment in the PR description for posterity would be enough though.
2. I made a couple of comments about defining GhostArray in addtional.dfy instead. I then realized that you implemented this as a Java-level pass which is why you need it defined in Java code. It does seem to me that it would be a lot cleaner to define it in additional.dfy and just translate array references directly. Since this is working I don't want to block on that, but I'd like to at least understand if my preference isn't possible for some reason.

keyboardDrummer:
> > A problem with this change is that Dafny arrays can no longer be used because their data field is not ghost, and non-ghost fields can not be assigned from ghost code.
> 
> That is unfortunate and feels like something we should address in Dafny. It's weirdly inconsistent to allow ghost code to instantiate a ghost array value but then forbid writing to it!

My impression is that this is a fundamental property that doesn't relate to arrays. Fields have to be declared either ghost or non-ghost, and non-ghost fields can only be assigned from non-ghost code.

```dafny
method foo() {
  var a := new int[100];
  bar(a); // we want to erase this call because it's to a ghost method, but doing would be unsound
  assert a[12] == 45;
}
ghost method bar(a: Array<int>) 
  ensures a[12] == 45
{
  a[12] := 45;
}
```

verifier/src/main/resources/object-contract.java
keyboardDrummer: > Wouldn't it be simpler to define the image of this directly in additional.dfy instead, since there are no explicit references to it from Java code?

Simpler how? Is this not the most simple? This way you get better precondition errors, and it is Dafny agnostic.

verifier/src/main/java/com/aws/jverify/verifier/compiler/TypeDeclarationCompiler.java
keyboardDrummer: I don't want to move GhostArray/JArray to additional.dfy because I think the less Dafny specific code the better. If you want me to resolve the above hacks to a mechanism, I can add an annotation for it, but I'd rather do that once we need this more often.

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/generics/PolymorphismDafnyResolutionErrors2.java
keyboardDrummer: What's not consistent here? It's a reference comparison between a ref type (ModifiableObject) and its non-ref supertype (Object)

robin-aws:
> My impression is that this is a fundamental property that doesn't relate to arrays. Fields have to be declared either ghost or non-ghost, and non-ghost fields can only be assigned from non-ghost code.

Sure, I was thinking of an example like this:

```dafny
ghost method bar() 
{
  var a := new int[100];
  a[12] := 45;  // Error: assignment to array element is not allowed in this context, because this is a ghost method
}
```

It *feels* obvious in that context that the state is ghost and should be writable. I get that in general the issue is knowing whether an arbitrary array value is allocated "on the ghost heap".

verifier/src/main/resources/object-contract.java
robin-aws: > This way you get better precondition errors, and it is Dafny agnostic.

Fair points. I was thinking you could avoid the special-casing on the type parameters and some of the other noise like the class/trait twin, but on second review it doesn't seem as bad.

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/generics/PolymorphismDafnyResolutionErrors2.java
robin-aws: I was worried this was an indication we were allowing `==` on value types when we shouldn't, but I realize now if that was the case other test cases would change behavior, so I think we're okay.

verifier/src/main/java/com/aws/jverify/verifier/compiler/TypeDeclarationCompiler.java
robin-aws: Agreed on perhaps creating an annotation later if this comes up, works for me.

robin-aws (dismissed):

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/JavaFrontEnd.java
keyboardDrummer: Yes this needs cleanup, and there's more than just this list that I'd like to refactor, but I'll leave that for another PR.

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/ArraysVerification.java
keyboardDrummer: It wasn't necessary before this PR. It's not clear to me what's different now from a verification perspective. I think the issue is that if `a == a[0]`, then the assignment `a[i] = ...`, which modifies `a`, would modify `a[0]`, and that could change `a[0].getA()`. With an extra `invariant(a != (Object)a[0])` and some checks, I can also make verification pass, without needing `invariant(a[0] == zeroPoint)`, but it's not clear to me why no extra invariant was needed before this PR.

keyboardDrummer:
> Sure, I was thinking of an example like this:
> 
> ```
> ghost method bar() 
> {
>   var a := new int[100];
>   a[12] := 45;  // Error: assignment to array element is not allowed in this context, because this is a ghost method
> }
> ```
> 
> It _feels_ obvious in that context that the state is ghost and should be writable. I get that in general the issue is knowing whether an arbitrary array value is allocated "on the ghost heap".

Good explanation, thanks. Another way to phrase it might be, that we only have 1 heap so objects are never ghost or non-ghost, but objects that are only reachable from ghost-variables should be assignable from a ghost context.

PR #319, opened 2025-09-03
Enable verifying `.equals` for records

keyboardDrummer:
### What was changed?
- Enable verifying `.equals` on records

### How has this been tested?
- Added additional testcase


library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -109,6 +109,10 @@ public static int old(int value) {
         throw new ContractException();
     }
 
+    public static <T> boolean jequals(T left, T right) {
```
robin-aws, line 121: I think this needs quite a bit more documentation to explain how this behaves and why you need it on top of `Object.equals`. Possibly even some content in the user guide as well. And possibly a better name too. :) But fine to address all that later.

robin-aws (dismissed):
Nice!

library/src/main/java/com/aws/jverify/JVerify.java
keyboardDrummer: Yeah.. Tbh I'm not even sure this should be user facing. Having it is more of an experiment ATM. I'll add a comment

PR #320, opened 2025-09-03
Enable inheriting unverified methods

keyboardDrummer:
### What was changed?
- Enable inheriting unverified methods

### How has this been tested?
- Added test InheritedUnverifiedMembers.java


test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -204,9 +204,9 @@ public static void verifyFile(SourceFile sourceFile, JVerifyTest annotation, Lis
                 () -> assertThat("exit code",
                         verificationResults.getExitCode(),
                         is(annotation.exitCode())),
-                () -> assertThat("Dafny verified count",
-                        verificationResults.getDafnyVerifiedCount(),
-                        is(expectedDafnyVerifiedCount)),
+//                () -> assertThat("Dafny verified count",
+//                        verificationResults.getDafnyVerifiedCount(),
+//                        is(expectedDafnyVerifiedCount)),
```
robin-aws: I assume not intended :)

verifier/src/main/java/com/aws/jverify/verifier/compiler/JavaToDafnyCompiler.java
```diff
@@ -623,23 +623,6 @@ public static Symbol.MethodSymbol getJVerifyMethod(JCTree.JCMethodInvocation inv
...
     }
 
-    /// The root Object type is currently defined directly in additional.dfy,
-    /// and includes the declaration of the `Object.equals` function.
-    /// But many Java types don't explicitly implement `equals`
-    /// because it has a default implementation based on reference equality.
-    /// Not having a class provide a body for a trait method isn't necessarily a problem in Dafny,
-    /// but it IS an error to not at least repeat the bodiless declaration.
-    /// Hence, this synthetic declaration we have to add to every constructable class.
-    public static Function equalsFunctionDeclaration(IOrigin origin) {
```
robin-aws, line 637: Woot, very glad to see this special case generalized!

robin-aws (changes requested):
Great change!

For the record, would it be more accurate to call this "enable overriding unverified methods"?

keyboardDrummer:
> For the record, would it be more accurate to call this "enable overriding unverified methods"?

I don't think so because previously if you inherited an unverified method that you did not override, you would get a Dafny error.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
keyboardDrummer: Actually I got tired of having to update those numbers, even with your automation ^,^ I'll update them for this PR, but we should make a PR that turns our library code into a proper Dafny library, so we can change our library without these numbers changing and also to improve performance.

PR #321, opened 2025-09-04
Enable early exits in pure methods

keyboardDrummer:
### What was changed?
- Enable early exits in pure methods

### How has this been tested?
- Added testcases

### DIFF
--- a/verifier/src/main/resources/builtin-contracts.java
+++ b/verifier/src/main/resources/builtin-contracts.java
@@ -420,7 +420,8 @@ class HelperForBigIntegerContract {
             return -stringToInt(v.substring(1));
         }
 
-        return (v.charAt(v.length()-1)-'0') + 10*stringToInt(v.substring(0,v.length()-1));
+        return v.charAt(v.length()-1)-'0' + 
+                10*stringToInt(v.substring(0,v.length()-1));
     }
 
     @Erased
--- a/verifier/src/main/resources/builtin-contracts.java
+++ b/verifier/src/main/resources/builtin-contracts.java
@@ -403,17 +403,24 @@ class HelperForBigIntegerContract {
                 ((v.charAt(0) == '+' || v.charAt(0) == '-') && isAllDigits(v.substring(1))) ||
                 isAllDigits(v));
     }
+    
     @Erased
     @Pure
     static @Unbounded int stringToInt(String v) {
         decreases(v.length());
-        return
-                v.length() == 0 ? 0 :
-                        (v.charAt(0) == '+' ?
-                                stringToInt(v.substring(1)) :
-                                (v.charAt(0) == '-' ? -stringToInt(v.substring(1)) :
-                                        (v.charAt(v.length()-1)-'0') + 10*stringToInt(v.substring(0,v.length()-1))
-                                ));
+        if (v.length() == 0) {
+            return 0;
+        }
+
+        if (v.charAt(0) == '+') {
+            return stringToInt(v.substring(1));
+        }
+
+        if (v.charAt(0) == '-') {
+            return -stringToInt(v.substring(1));
+        }
+
+        return (v.charAt(v.length()-1)-'0') + 10*stringToInt(v.substring(0,v.length()-1));
     }
 
     @Erased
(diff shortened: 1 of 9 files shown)


verifier/src/test/java/com/aws/jverify/verifier/tests/verification/pure/PureMethodErrors.java
```diff
@@ -26,6 +26,17 @@ void pureWithoutReturn() {
...
     }
+
+    @Pure
+    int pureEndingWithIfNotElse(int x) {
+        if (x > 2) {
+            if (x > 3) {
+//          ^ error: a pure block may not end in an if statement without an else block    
+                return 3;
+            }
+        }
```
olivier-aws, line 37: I don't understand why we don't allow this. It is equivalent to
```
if (x>2 && x>3) {
    return 3;
}
return 2;
```
which would be accepted. What blocks us from accepting such a case?

keyboardDrummer: It is equivalent and we could accept it if we did that translation, but we don't do that translation. I think supporting that would require adding a rather specific matching rule that doesn't increase general expressive power.

What about:
```java
        if (x > 2) {
            var y = x + 3;
            if (y > 3) {
//          ^ error: a pure block may not end in an if statement without an else block    
                return 3;
            }
        }
```

How would you support that?

olivier-aws: Agree that there are complex cases that would require lots of analysis to be transformed into a single
```
if (...)
   return ...;
```

My more general question would actually be: why don't we allow any method that only contains if and return statements? Is it because we don't know how to translate that into a single Dafny expression?

keyboardDrummer: > why don't we allow any method that only contains if and return statements? Is it because we don't know how to translate that into a single Dafny expression?

You would get duplication if you do it in a mechanical way.

```java
if A {
  if B {
     return C;
  }
}
return D;
```

==>

```java
return A ? (B ? C : D) : D
```

Notice D was duplicated

verifier/src/main/resources/builtin-contracts.java
```diff
@@ -403,17 +403,25 @@ static boolean isValidString(String v) {
...
+        if (v.charAt(0) == '+') {
+            return stringToInt(v.substring(1));
+        }
+
+        if (v.charAt(0) == '-') {
+            return -stringToInt(v.substring(1));
+        }
+
+        return v.charAt(v.length()-1)-'0' + 
+                10*stringToInt(v.substring(0,v.length()-1));
```
olivier-aws, line 424: Love how much simpler it is.

olivier-aws (approved):
LGTM. Just a question out of curiosity. What makes that this passes
```
if (x>0) {
    if (x>1) {
        return 1;
    }
    return 2;
}
return 2;
```
while this does not
```
if (x>0) {
    if (x>1) {
        return 1;
    }
}
return 2;
```

From the translator code, I imagine that in the Java AST, there is an else branch in the first case and not in the second? Is this else branch added by some javac transform?

keyboardDrummer:
> LGTM. Just a question out of curiosity. What makes that this passes
> 
> ```
> if (x>0) {
>     if (x>1) {
>         return 1;
>     }
>     return 2;
> }
> return 2;
> ```
> 
> while this does not
> 
> ```
> if (x>0) {
>     if (x>1) {
>         return 1;
>     }
> }
> return 2;
> ```
> 
> From the translator code, I imagine that in the Java AST, there is an else branch in the first case and not in the second? Is this else branch added by some javac transform?

There's no else branch in the first case either. The question is: can we transform the program to give each `if` statement an else branch?

For the first example the answer is yes:
```java
if (x>0) {
  if (x > 1) {
    return 1;
  } else {
    return 2; 
  }
} else {
  return 2;
}
```

but for the second we can't

PR #322, opened 2025-09-05
Enable defining a @Contract in different locations

keyboardDrummer:
### What was changed?
- Enable using `@Contract` on the same target type multiple times. 
- TODO: This requires changing the way the Uri part of source locations is sent to Dafny, since instead of multiple top level files, there needs to be a single tree within which the Uri can be changed at arbitrary subtrees.

### How has this been tested?


fabiomadge:

PR #328, opened 2025-09-19
Support nullable pure types

keyboardDrummer:
### What was changed?
- Support @Nullable on pure types. Note that this support is temporary in the sense that it fundamentally does not support the combination of @Nullable with polymorphic types, and the implementation will have to be completely redone to support that.

### How has this been tested?
- Added a test `NullableInterfacesVerification`


robin-aws:
> the implementation will have to be completely redone to support that.

Not COMPLETELY. :) Translating many things to the `Nullable<T>` datatype and its operations should stay the same AFAICT.

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/records/RecordsErrors.java
```diff
@@ -79,7 +79,6 @@ static class WantsContract {}
 
     // This is a limitation of the current implementation; we'd like to allow matching Java semantics more precisely.
```
robin-aws: Comment out of date (yay!)

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/interfaces/NullableInterfacesVerification.java
robin-aws, line 1: Honest question, is it worth having an explicit test of a concrete `@Nullable` class or record type, and not just pure interfaces?

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/ExpressionContext.java
```diff
@@ -2,15 +2,24 @@
...
 import com.aws.jverify.Nullable;
 import com.aws.jverify.generated.Statement;
-import net.bytebuddy.implementation.LoadedTypeInitializer;
+import com.sun.tools.javac.code.Type;
 
 import java.util.function.Consumer;
 
-public record ExpressionContext(@Nullable Consumer<Statement> statementWriter, boolean allowImpure, @Nullable BlockCompiler blockCompiler) {
-    public final static ExpressionContext Pure =  new ExpressionContext(null, false, null);
+enum Nullability { Nullable, NonNullable }
```
robin-aws: Unused?

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/BlockCompiler.java
```diff
@@ -59,40 +60,22 @@ public List<Statement> translateStatement(JCTree.JCStatement statement, IOrigin
...
-            case JCTree.JCBreak jcBreak -> {
-                yield translateBreak(jcBreak);
-            }
-            case JCTree.JCSwitch jcSwitch -> {
-                yield translateSwitchStatement(jcSwitch, expressionContext);
-            }
-            case JCTree.JCSkip _ -> {
-                yield List.of();
-            }
+            case JCTree.JCExpressionStatement expressionStatement ->
```
robin-aws, line 63: (teensy nit) For the record I actually found the old version a bit easier to read and to visually compare the different branches to each other, especially since some expressions then wrap lines.

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/NullableGenerator.java
```diff
@@ -34,7 +39,16 @@ private static String getNullableSuffix(boolean isNullable) {
         return isNullable ? "?" : "";
     }
 
-    private boolean isNullable(com.sun.tools.javac.code.Type type, JCTree.JCModifiers additionalModifiers) {
+    private boolean isNullable(JCTree.JCExpression expression) {
+        if (expression instanceof JCTree.JCIdent identifier) {
```
robin-aws, line 43: Surprised we lasted this long without handling these cases :)

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/NullableGenerator.java
```diff
@@ -48,10 +62,10 @@ public UserDefinedType translateArrayType(com.sun.tools.javac.code.Type.ArrayTyp
                                               JCTree.JCModifiers additionalModifiers) {
         var isNullable = isNullable(arrayType, additionalModifiers);
         var result = super.translateArrayType(arrayType, origin, additionalModifiers);
-        return makeUserDefinedTypeNullable(result, isNullable);
+        return addQuestionMarkToUserDefinedType(result, isNullable);
     }
 
-    private static UserDefinedType makeUserDefinedTypeNullable(UserDefinedType original, boolean nullable) {
+    private static UserDefinedType addQuestionMarkToUserDefinedType(UserDefinedType original, boolean nullable) {
```
robin-aws: (nit) I would have found it more natural to keep `makeUserDefinedTypeNullable` and add the `"if immutable" case to it instead.

robin-aws (dismissed):
Very nice, thanks!! This will unblock a lot of common uses. Nothing but nits from me.

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/interfaces/NullableInterfacesVerification.java
keyboardDrummer: Added them.

keyboardDrummer:
> Not COMPLETELY. :) Translating many things to the Nullable<T> datatype and its operations should stay the same AFAICT.

I was planning on adding a transformation before the attribute phase (basically during or after parsing) that replaces occurrences of `@Nullable TypeFoo` to `Nullable<TypeFoo>`, where `Nullable` is defined in Java. I would override `Types.isCompatible` to allow `Nullable<TypeFoo>` to be compatible with `TypeFoo`, and then possibly in Lower extend the code that adds casts between compatible but different types.

So then NullableGenerator would become entirely obsolete.

PR #329, opened 2025-09-23
Report count of java methods verified

### LINKED ISSUE #317: Let JVerify report how many Java methods were present, attempted to verify, verified and failed to verify

oyendrila-dobe:
### What was changed?
- Reporting 3 counts along with Dafny method counts:
     - Number of in-source methods that are verifiable
     - Number of in-source methods currently being verified (based on parsing of @Verify annotation)
     - Number of java methods failing to verify

- Added sourcefile to new static class created 
     - Using this source file details to create interval search tree of locations of methods in the file. Since Dafny reports this source file path in its error messages, it allows an easy connection.
 

### How has this been tested?
Modified a few existing test cases to add expected counts for java methods 


test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTest.java
```diff
@@ -56,4 +56,16 @@
...
+     * Expected number of Java methods to fail verification
+     */
+    int javaErrors()  default -1;
+
+    /**
+     * Expected number of Java methods to be verified
+     * Note: add +1 for each class you are verifying to account
+     * for the implicit constructor unless you have it specific explicitly
+     */
+    int javaVerified() default -1;
```
keyboardDrummer, line 70: Cool!

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -384,6 +415,82 @@ public String[] additionalFiles() {
...
+            }
+
+            @Override
+            public int javaVerified() {
+                return -1;
+            }
+        };
+    }
+
+    public static JVerifyTest makeJVerifyTestAnnotation(boolean verifyByDefault, int exitCode,
```
keyboardDrummer: Seems unused

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -85,11 +89,21 @@ public static int verifyJavaFiles(
...
     ) throws IOException {
         var verificationResults = verifyJavaFiles(readFiles, verifierOptions);
-        outputVerificationResults(verificationResults, verifierOptions, outputWriter);
+        var intervalTreeMap = getIntervalTreeMap(verificationResults);
+        outputVerificationResults(verificationResults, verifierOptions, outputWriter, intervalTreeMap);
         return verificationResults.getExitCode();
     }
 
-    private static void outputVerificationResults(VerificationResults verificationResults, VerifierOptions verifierOptions, Writer outputWriter) throws IOException {
+    public static HashMap<URI, MethodIntervalTree> getIntervalTreeMap(VerificationResults verificationResults) {
```
keyboardDrummer: Consider moving this to `MethodIntervalTree`

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -104,11 +118,36 @@ private static void outputVerificationResults(VerificationResults verificationRe
...
             outputWriter.write(verificationResults.getDafnyFinishedMessage());
         }
+
+        if (verificationResults.getJavaInSourceMethods() != null) {
+            outputWriter.write('\n');
+            outputWriter.write("Total Java methods in source code: ");
+            outputWriter.write(String.valueOf(verificationResults.getJavaInSourceMethods().size()));
+
+            outputWriter.write('\n');
+            outputWriter.write("Total Java methods under verification: ");
```
keyboardDrummer: I find the meaning of "under verification" not immediately clear. It sounds a little like those methods are still being verified. Consider rewriting to: "Attempted to verify {x} methods" or "Number of methods that JVerify tried to verify: {x}". Also, instead of this number, consider reporting the number of successfully verified methods: "Successfully verified {x} methods"

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -104,11 +118,36 @@ private static void outputVerificationResults(VerificationResults verificationRe
...
+            outputWriter.write(String.valueOf(verificationResults.getJavaInSourceMethods()
+                    .stream().filter(JavaMethodDetails::isVerifiable)
+                    .toList().size()));
+
+            outputWriter.write('\n');
+            outputWriter.write("Total Java methods not verified: ");
```
keyboardDrummer: This reads as if it also includes the methods that it did not attempt to verify. Consider rewriting to: "Failed to verify {x} methods" or "Number of method that failed to verify: {x}"

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -104,11 +118,36 @@ private static void outputVerificationResults(VerificationResults verificationRe
...
+
             }
         }
         if (verificationResults.getDafnyFinishedMessage() != null) {
             outputWriter.write(verificationResults.getDafnyFinishedMessage());
         }
+
+        if (verificationResults.getJavaInSourceMethods() != null) {
+            outputWriter.write('\n');
+            outputWriter.write("Total Java methods in source code: ");
```
keyboardDrummer: Consider reporting the number of methods for which verification was skipped instead, because that's the number they (should) want to drive to 0, while the total number of methods has no meaning on its own.

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -325,4 +364,22 @@ private static void parseDafnyJsonOutput(NameCompiler nameCompiler,
      */
     @JsonIgnoreProperties({"pos"})
     private static abstract class DafnyJsonPosition {}
+
+    private static HashMap<URI,MethodIntervalTree> getIntervalsFromVerifiableMethods(ArrayList<JavaMethodDetails> javaInSourceMethods) {
```
keyboardDrummer: Considering moving this to `MethodIntervalTree`

verifier/src/main/java/com/aws/jverify/verifier/JavaMethodDetails.java
```diff
@@ -0,0 +1,58 @@
...
+import com.sun.tools.javac.tree.JCTree;
+
+public class JavaMethodDetails {
+    private JCTree.JCMethodDecl methodTree;
+    private Token startPosition;
+    private Token endPosition;
+    private boolean verifiable;
+    private boolean verified;
+
+    public JavaMethodDetails(JCTree.JCMethodDecl methodTree, Token startPosition, Token endPosition, boolean verifiable, boolean verified) {
```
keyboardDrummer: Instead of having a start and end token, consider using a `TokenRange` which already has a start and end.

What does `verifiable` mean? Are there methods that can not be verified? 

Instead of the booleans `verifiable` and `verified`, how about an `enum VerificationStatus { Skipped, VerifiedSuccessfully, FailedToVerify }` ?

verifier/src/main/java/com/aws/jverify/verifier/JavaMethodDetails.java
```diff
@@ -0,0 +1,58 @@
...
+    public void setVerified(boolean verified) {
+        this.verified = verified;
+    }
+
+    public void setPosition(Token startPosition, Token endPosition) {
+        this.startPosition = startPosition;
+        this.endPosition = endPosition;
+    }
+
+    public void setMethodSymbol(JCTree.JCMethodDecl methodTree) {
```
keyboardDrummer: The method name `symbol` does not correspond to the argument since that's a `JCTree` not a `Symbol`

verifier/src/main/java/com/aws/jverify/verifier/MethodIntervalTree.java
```diff
@@ -0,0 +1,62 @@
+package com.aws.jverify.verifier;
+
+class IntervalTreeNode {
```
keyboardDrummer: Consider making this an static inner class of IntervalTree

verifier/src/main/java/com/aws/jverify/verifier/MethodIntervalTree.java
```diff
@@ -0,0 +1,62 @@
...
+    int maxEnd;
+    IntervalTreeNode left, right;
+
+    IntervalTreeNode(JavaMethodDetails method) {
+        this.method = method;
+        this.maxEnd = method.getEndPosition().getLine();
+    }
+}
+
+public class MethodIntervalTree {
```
keyboardDrummer: Have you considered making this generic? Like `IntervalTree<Key extends Comparable, Value>`. I know you only use it with methods now but it can make it easier to read if you know that it does not depend on properties of method.

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/JavaToDafnyCompiler.java
```diff
@@ -50,13 +54,13 @@ public JavaToDafnyCompiler(Context context) {
...
         context.put(DiagnosticListener.class, diagnostics);
-        
+
         reporter = Reporter.instance(context);
         enter = Enter.instance(context);
         this.dafnyGenerator = DafnyGenerator.getGenerator(context);
     }
 
-    public @Nullable FilesContainer analyzeJavaCode(VerifierOptions options, java.util.List<JavaFileObject> files) {
+    public @Nullable FilesContainer analyzeJavaCode(VerifierOptions options, java.util.List<JavaFileObject> files, VerificationResults results) {
```
keyboardDrummer: I don't like how you're using `VerificationResults` here. You're not using it to carry verification results. It's only carrying the `setJavaInSourceMethods` property. How about you create a field in `JavaToDafnyCompiler` where you set that property? 

Even better, only have a field `compilationUnits` that is set in `JavaToDafnyCompiler`, which is used to compute `javaInSourceMethod` from outside `JavaToDafnyCompiler`, using `JavaToDafnyCompiler`, `VerifyAnnotationCompiler` and `MissingContractsCompiler`, all of which should be stored in `Context`. `JavaToDafnyCompiler` currently is not stored in `Context` but you can add a call to `context.put(JavaToDafnyCompiler.class, this)` in the constructor.

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/JavaToDafnyCompiler.java
```diff
@@ -288,4 +295,38 @@ private Set<JCTree.JCCompilationUnit> lower(Set<JCTree.JCCompilationUnit> envs)
         }
         return envs;
     }
+
+    private @Nullable ArrayList<JavaMethodDetails> processVerifyingMethods(ArrayList<JCTree.JCMethodDecl> shouldVerifyList,
```
keyboardDrummer: Does not relate to `JavaToDafnyCompiler`, so shouldn't be in this class.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MissingContractCompiler.java
```diff
@@ -33,6 +34,8 @@ public class MissingContractCompiler {
     private final Set<Symbol> foundSymbols = new HashSet<>();
     private final Reporter reporter;
 
+    private ArrayList<JCTree.JCMethodDecl> canVerifyMethodList;
```
keyboardDrummer: What does it mean when a method can be verified? Can't every method be verified? Is this the list of all source methods?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MissingContractCompiler.java
```diff
@@ -45,17 +48,20 @@ public MissingContractCompiler(Context context) {
         jverifyUtils = JVerifyUtils.instance(context);
         reporter = Reporter.instance(context);
         internalContractCompiler = MethodOrLoopContractCompiler.instance(context);
+        canVerifyMethodList = new ArrayList<JCTree.JCMethodDecl>();
     }
     
     public Set<JCTree.JCCompilationUnit> compile(Set<JCTree.JCCompilationUnit> units) {
         var finder = new SymbolReferenceFinder();
         for(var unit : units) {
+            ArrayList<JCTree.JCMethodDecl> currentCanVerifyMethodList = new ArrayList<JCTree.JCMethodDecl>(canVerifyMethodList);
```
keyboardDrummer: Oof ! You're copying the entire list here? That turns the compilation time complexity into `O(N^2)` where N is the size of the input program.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MissingContractCompiler.java
```diff
@@ -45,17 +48,20 @@ public MissingContractCompiler(Context context) {
         jverifyUtils = JVerifyUtils.instance(context);
         reporter = Reporter.instance(context);
         internalContractCompiler = MethodOrLoopContractCompiler.instance(context);
+        canVerifyMethodList = new ArrayList<JCTree.JCMethodDecl>();
```
keyboardDrummer: What do you need `MissingContractCompiler` for? It seems like you only use it to traverse all methods, but `VerifyAnnotationCompiler` does that as well. Why not use only that class?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MissingContractCompiler.java
```diff
@@ -313,4 +320,15 @@ public boolean isRecordAccessor(Symbol.MethodSymbol method) {
...
         return false;
     }
+
+    public ArrayList<JCTree.JCMethodDecl> getCanVerifyMethodList() {
+        return canVerifyMethodList;
+    }
+
+    private void adjustCanVerifyMethodList(JavaFileObject sourceFile,
+                                           ArrayList<JCTree.JCMethodDecl> previousCanVerifyMethodList) {
+        if (sourceFile.toUri().getScheme().equals("string")) {
```
keyboardDrummer: This is not a proper way to detect if something is a library, and the code is also not self-documenting. Please introduce a method `isLibrary` in `JavaToDafnyCompiler` that uses the `builtinSources` property.

keyboardDrummer:
How do you handle methods for which an external contract was defined? After `ExternalContractCompiler` runs, they will have a JCMethodDecl and a user provided source file. Do you count those as source methods? I think that wouldn't make sense to the user. For something to count as a verifiable method, I think it has to have a body - so we know it's not an abstract or interface method - and it has to have an implementation (not just a contract): `method.body != null && MethodOrLoopContractCompiler.hasImplementation(method)`

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MissingContractCompiler.java
oyendrila-dobe: The would require `buildinSources` to be static though. That should not affect any usage of the class IMO

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MissingContractCompiler.java
oyendrila-dobe: Should have thought better. Found a way to avoid this.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MissingContractCompiler.java
oyendrila-dobe: Yes. Renamed

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/JavaToDafnyCompiler.java
oyendrila-dobe: ah this is so much cleaner and more beautiful!

oyendrila-dobe:
For a java method that has a body, but is annotated with `@Verify(false)`, `VerifyAnnotationCompiler` removes its body after parsing. If I only use VerifyAnnotationCompiler, on adding the methodDecl to the list, the above removal makes this method bodyless, which causes it to be filtered out during method count as `verifiable`. 

Do we have a way to deepcopy the method tree?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MissingContractCompiler.java
keyboardDrummer: > The would require buildinSources to be static though. 

Why not get an instance of `JavaToDafnyCompiler` ? Put one in the `Context` in the constructor of `JavaToDafnyCompiler` and retrieve it when needed.

keyboardDrummer:
> For a java method that has a body, but is annotated with `@Verify(false)`, `VerifyAnnotationCompiler` removes its body after parsing. If I only use VerifyAnnotationCompiler, on adding the methodDecl to the list, the above removal makes this method bodyless, which causes it to be filtered out during method count as `verifiable`.

How about doing that count before the body is removed?

> Do we have a way to deepcopy the method tree?

Let's avoid copying for performance's sake

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -189,6 +189,18 @@ public static void verifyFile(SourceFile sourceFile, JVerifyTest annotation, Lis
                 .sorted()
                 .toList();
 
+        verificationResults.getOutputs().stream()
+                .filter(dafnyOutput -> dafnyOutput instanceof DafnyDiagnostic)
+                .forEach(dafnyOutput -> {
+                    var unverifiedJavaMethod = intervalTreeMap.get(((DafnyDiagnostic) dafnyOutput).getSource())
```
keyboardDrummer: unverifiedJavaMethod sounds we did not attempt to verify it. Consider renaming to `methodWithError` or `failedVerificationMethod`

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -209,7 +224,37 @@ public static void verifyFile(SourceFile sourceFile, JVerifyTest annotation, Lis
...
+                                verificationResults.getJavaInSourceMethods().stream()
+                                        .filter(method ->  method.getVerificationStatus() == JavaMethodDetails.VerificationStatus.Verified)
+                                        .toArray().length,
+                                is(expectedJavaVerifiedCount));
+                    }
+                },
+                () -> {
+                    if (expectedJavaErrorCount != null) {
+                        assert verificationResults.getJavaInSourceMethods() != null;
+                        assertThat("Java unverified method count",
```
keyboardDrummer: -> "Java verification failed method count"

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -52,11 +60,16 @@ public static VerificationResults verifyJavaFiles(
...
         TypesWithoutErasure.preRegister(context);
         context.put(VerifierOptions.class, verifierOptions);
-        
+
         var messages = JavacMessages.instance(context);
         messages.add("com.aws.jverify.messages");
 
         var dafnyEquivalent = new JavaToDafnyCompiler(context).analyzeJavaCode(verifierOptions, readFiles);
+
+        verificationResults.setJavaInSourceMethods(processVerifyingMethods(context.get(VerifyAnnotationCompiler.class).getJavaMethodUnderVerification(),
```
keyboardDrummer: Why are you computing this here instead of where you use it, in `outputVerificationResults` ? You can pass the `context` to there.

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -85,11 +98,15 @@ public static int verifyJavaFiles(
             Writer outputWriter
     ) throws IOException {
         var verificationResults = verifyJavaFiles(readFiles, verifierOptions);
-        outputVerificationResults(verificationResults, verifierOptions, outputWriter);
+        var intervalTreeMap = IntervalTree.getIntervalTreeMap(verificationResults);
```
keyboardDrummer: Why compute `intervalTreeMap` here instead of where you use it?

verifier/src/main/java/com/aws/jverify/verifier/JavaMethodDetails.java
```diff
@@ -0,0 +1,47 @@
+package com.aws.jverify.verifier;
+
+import com.aws.jverify.generated.TokenRange;
+import com.sun.tools.javac.tree.JCTree;
+
+public class JavaMethodDetails {
```
keyboardDrummer: This name is too generic. This is more like `JavaMethodVerificationStatus`

verifier/src/main/java/com/aws/jverify/verifier/IntervalTree.java
```diff
@@ -0,0 +1,101 @@
...
+
+        if (node.right != null && point.compareTo(node.start) >= 0) {
+            return search(node.right, point);
+        }
+
+        return null;
+    }
+
+
+    public static HashMap<URI, IntervalTree<Integer, JavaMethodDetails>> getIntervalTreeMap(VerificationResults verificationResults) {
```
keyboardDrummer: Now that `IntervalTree` no longer relates to Java methods, better move this code to `Driver`

verifier/src/main/java/com/aws/jverify/verifier/IntervalTree.java
```diff
@@ -0,0 +1,101 @@
...
+        HashMap<URI, IntervalTree<Integer, JavaMethodDetails>> intervalTreeMap = null;
+        if (verificationResults.getJavaInSourceMethods() != null) {
+            intervalTreeMap = getIntervalsFromVerifiableMethods(verificationResults.getJavaInSourceMethods());
+        }
+        return intervalTreeMap;
+    }
+
+    private static HashMap<URI,IntervalTree<Integer, JavaMethodDetails>> getIntervalsFromVerifiableMethods(ArrayList<JavaMethodDetails> javaInSourceMethods) {
```
keyboardDrummer: Now that `IntervalTree` no longer relates to Java methods, better move this code to `Driver`

verifier/src/main/java/com/aws/jverify/verifier/VerificationResults.java
```diff
@@ -29,6 +29,12 @@ public final class VerificationResults {
      */
     private @Nullable String dafnyFinishedMessage;
 
+    /**
+     * Tracks the java in-source methods that can be verified,
+     * their location in the source file, and if they are being verified
+     */
+    private @Nullable ArrayList<JavaMethodDetails> javaInSourceMethods = null;
```
keyboardDrummer: Why did you put this property in this class? I don't think it belongs here.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/VerifyAnnotationCompiler.java
```diff
@@ -67,6 +80,10 @@ public void visitClassDef(JCTree.JCClassDecl tree) {
     @Override
     public void visitMethodDef(JCTree.JCMethodDecl tree) {
         boolean shouldVerify = processVerifyAnnotationAndPop(tree.sym);
+        if (considerCurrentEnv) {
```
keyboardDrummer: `considerCurrentEnv`, consider for what? I know the answer but only because I know what this PR is about. Without this PR that name is too generic. I suggest you replace `considerCurrentEnv` with `javaToDafnyCompiler.isLibrary(reporter.compilationUnit)`, storing `javaToDafnyCompiler` in a field, and setting `reporter.compilationUnit` where you are currently setting `considerCurrentEnv`

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -325,4 +385,45 @@ private static void parseDafnyJsonOutput(NameCompiler nameCompiler,
      */
     @JsonIgnoreProperties({"pos"})
     private static abstract class DafnyJsonPosition {}
+
+    private static @Nullable ArrayList<JavaMethodDetails> processVerifyingMethods(HashMap<JCTree.JCMethodDecl, VerifyAnnotationCompiler.MethodStatus> listOfMethodsUnderVerification,
```
keyboardDrummer: The term `underVerification` is ambiguous to me. To me the name implies that these methods are being verified at this very moment. Please use something more explicit.

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -325,4 +385,45 @@ private static void parseDafnyJsonOutput(NameCompiler nameCompiler,
...
+                                                                           Set<JCTree.JCCompilationUnit> units) {
+        ArrayList<JavaMethodDetails> combinedList = new ArrayList<>();
+        if (!listOfMethodsUnderVerification.isEmpty()) {
+
+            var subset = listOfMethodsUnderVerification.entrySet().stream()
```
keyboardDrummer: Is the variable `subset` used?

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -325,4 +385,45 @@ private static void parseDafnyJsonOutput(NameCompiler nameCompiler,
...
+                    .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+            combinedList = listOfMethodsUnderVerification.entrySet().stream()
+                    .filter( entry -> !BaseDafnyGenerator.isSynthetic(entry.getKey().getModifiers().flags)
+                            && entry.getValue().hasImpl())
+                    .map(entry -> new JavaMethodDetails(entry.getKey(), null,
+                            entry.getValue().hasVerify() ? JavaMethodDetails.VerificationStatus.Unknown
+                                    : JavaMethodDetails.VerificationStatus.Skipped))
+                    .collect(Collectors.toCollection(ArrayList::new));
+        }
+        var nonLibraryClasses = units.stream()
```
keyboardDrummer: `nonLibraryUnits` ?

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -325,4 +385,45 @@ private static void parseDafnyJsonOutput(NameCompiler nameCompiler,
...
+                .filter(unit -> !unit.getSourceFile().toUri().getScheme().equals("string"))
```
keyboardDrummer: this is not the right way to check whether something is a library

verifier/src/main/java/com/aws/jverify/verifier/VerificationResults.java
```diff
@@ -74,4 +80,21 @@ public void setDafnyVerifiedCount(@Nullable Integer dafnyVerifiedCount) {
...
+
+    public ArrayList<JavaMethodDetails> getJavaInSourceMethods() {
+        return javaInSourceMethods;
+    }
+
+    public void setVerifiedVerificationStatus() {
+        // marking any unskipped method that was not associated with any failure message to be successful
+        if (getJavaInSourceMethods() != null) {
+            getJavaInSourceMethods().stream()
+                    .filter(method -> method.getVerificationStatus() == JavaMethodDetails.VerificationStatus.Unknown)
```
keyboardDrummer: Why first mark the status as unknown and later change that to Verified, instead of marking it as Verified right away? Can a method ever keep the status unknown after the algorithm runs? If not, I think you can remove `VerificationStatus.Unknown`

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
keyboardDrummer: Why do you need to do this processing, instead of building an `HashMap<URI, IntervalTree<Integer, JavaMethodDetails>>` in `VerifyAnnotationCompiler` directly? I don't see the need for an intermediate `HashMap<JCTree.JCMethodDecl, VerifyAnnotationCompiler.MethodStatus>`

keyboardDrummer:
> it stores the methods as well their verification statuses. The name is misleading, but i think the data it carries actually belongs to this class

Reading the latest version of the code, I agree with you. Sorry for the bad feedback I gave before. Would you mind undoing that change so that the intervalTree is again in `VerificationResults` ?

keyboardDrummer:
Great stuff! I love it. I added one more comment, but I'll already approve this as is.

keyboardDrummer (dismissed):

oyendrila-dobe:
> > it stores the methods as well their verification statuses. The name is misleading, but i think the data it carries actually belongs to this class
> 
> Reading the latest version of the code, I agree with you. Sorry for the bad feedback I gave before. Would you mind undoing that change so that the intervalTree is again in `VerificationResults` ?

I'm currently, creating a [record](https://github.com/aws/jverify/pull/329/files#diff-50b4e2aa56d6761bda59550fa392d60607972c220b7547261b32673e255aa60cR31-R32) in `Driver`, to save the verification result and interval tree map. In this current context, i don't think we should add the interval maps to the `VerificationResult` class

keyboardDrummer (approved):

PR #330, opened 2025-09-25
Rename `@Modifiable` to `@Impure`

keyboardDrummer:
### What was changed?
- Rename `@Modifiable` to `@Impure`, and update documentation

### Future work
- Replace the `pure` field of `@Contract` with more usages of `@Pure`
- Consider replacing `@Impure` by changing defaults and adding more usages of `@Pure`. I tried doing this already but it's not such a trivial change and I wasn't entirely convinced by the effect on UX, so I'd prefer to do that in a separate PR. One unexpected UX effect is that when `@Pure` can be used both on methods and on type references (which would be needed to allow `@Pure Object`), then it will bind to the return type of methods instead of the method. Given that, we can't replace `@Impure` with more usages of `@Pure`, although we could introduce something like `@PureType`

### How has this been tested?
- Updated existing tests


library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -96,6 +96,13 @@ public static void reads(Object object) {
     public static void modifies(Object object) {
     }
 
+    /**
+     * Can be used in reads and modifies clauses to refer to all objects
```
robin-aws: Only reads, `modifies *` is not valid Dafny and I don't think we want to change that.

robin-aws (changes requested):
LGTM, agreed this is strictly better even if there's more to do. Just one important correction.

PR #332, opened 2025-09-26
Flow typing

keyboardDrummer:
### What was changed?
- Support `<Expr> instanceof <Type> <Identifier>` in combination with `if <expr> { } else {}`, `<expr> ? <expr> : <expr>` and with `<expr> && <expr>`

### How has this been tested?
- Added test `FlowTyping.java`

### DIFF
--- a/verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/DafnyGenerator.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/DafnyGenerator.java
@@ -4,6 +4,7 @@ import com.aws.jverify.generated.*;
 import com.aws.jverify.verifier.compiler.dafnygenerator.base.BaseDafnyGenerator;
 import com.aws.jverify.verifier.compiler.dafnygenerator.base.BlockCompiler;
 import com.aws.jverify.verifier.compiler.dafnygenerator.base.ExpressionContext;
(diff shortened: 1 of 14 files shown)


verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/DafnyGenerator.java
```diff
@@ -17,7 +18,11 @@ public interface DafnyGenerator {
     
     List<Statement> translateStatementAfterLabel(BlockCompiler blockCompiler, JCTree.JCStatement statement, List<Label> labels, IOrigin originOverride);
 
-    Expression toExpr(JCTree.JCExpression expr, IOrigin originOverride, ExpressionContext context);
+    default Expression toExpr(JCTree.JCExpression expr, IOrigin originOverride, ExpressionContext context) {
```
oyendrila-dobe, line 21: why do we still need this? It seems most of the existing usage of this method is replaced by the new `toExprWithFlow` method

oyendrila-dobe (approved):
LGTM

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/DafnyGenerator.java
keyboardDrummer: Many calls don't care about the flow, so this is a utility to immediately throw that flow information away

PR #334, opened 2025-09-29
Fix array copy

### LINKED ISSUE #333: Crash when copying an array cell into another array cell
Analyze this code with JVerify:
```java
    static void arrayCopy() {
        int[] tab = new int[4];
        int[] tab2 = new int[4];
        tab2[1] = 10;
        tab[1] = tab2[1];
        check(tab[1]==10);
    }
```
There is a crash
```
java.lang.RuntimeException: not supported. should have already been lowered
	at com.aws.jverify.verifier.compiler.dafnygenerator.base.ExpressionCompiler.translateArrayAccess(ExpressionCompiler.java:374)
...
```

olivier-aws:
### What was changed?
Fix for issue #333 

### How has this been tested?
Added a test case

### DIFF
--- a/verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ArrayCompiler.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ArrayCompiler.java
@@ -65,7 +65,7 @@ public class ArrayCompiler extends TreeTranslator {
     public void visitAssign(JCTree.JCAssign tree) {
         if (tree.lhs instanceof JCTree.JCArrayAccess arrayAccess) {
             maker.pos = tree.pos;
-            tree.rhs = super.translate(tree.rhs);
+            tree.rhs = translate(tree.rhs);
 
             result = maker.App(maker.Select(arrayAccess.getExpression(), setMethodSymbol),
                     List.of(arrayAccess.getIndex(), tree.rhs));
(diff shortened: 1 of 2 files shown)


verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ArrayCompiler.java
```diff
@@ -65,12 +65,16 @@ public void visitNewClass(JCTree.JCNewClass tree) {
     public void visitAssign(JCTree.JCAssign tree) {
         if (tree.lhs instanceof JCTree.JCArrayAccess arrayAccess) {
             maker.pos = tree.pos;
-            result = maker.App(maker.Select(arrayAccess.getExpression(), setMethodSymbol), 
-                    List.of(arrayAccess.getIndex(), tree.getExpression()));
+            tree.rhs = super.translate(tree.rhs);
```
keyboardDrummer: Should it be `super.translate` or `translate` ? Also, should we call `super.visitAssign(tree)` at the top instead of having this?

olivier-aws: We can't call super.visitAssign(tree) at the top. This would translate
```java
  tab1[i] = tab2[i];
```
into
```dafny
  tab1.get(i) := tab2.get(i);
```
I think it is the same to call `translate` or `super.translate` because `ArrayCompiler` does not override the `translate` method anyway, but you are right that it seems more natural to call `translate`. Updated.

keyboardDrummer (approved):

keyboardDrummer:
Nice, thanks :)

PR #341, opened 2025-10-14
Various fixes Oct14

keyboardDrummer:
#### UI
- Let the repository VSCode task "verify" also dump the Dafny program to a temporary file
- Let `getExtraS` work well for the zero elements case
- Make the method verification summary work when working with relative URIs
- Improve debugging by stopping compilation early if there are errors

#### Semantic
- Handle more cases related to `@Nullable` 
- Enable calling non-void methods and not using their result
- Enable using `check` and `assume` in pure code
- Small improvements to expected type code that's used in expression compilation

#### Documentation
- Make SumCache example work.
- Add DataConsistency example.

### How has this been tested?
Added tests related to the 4 semantics changes:
- Slightly extended `NullableInterfacesVerification.java`
- Add `ignoreCallResult` test
- Add `pureMethodWithAssumeAndCheck` result
- Add `sumNullableString` test


examples/src/test/java/com/aws/jverify/examples/SumCache.java
```diff
@@ -1,36 +1,36 @@
...
+    int sum = 0;
 
+    @Pure
     @Invariant
     boolean validSum() {
-        reads(this);
-        reads(numbers);
-//          ^^^^^^^^ Related: this is the invariant that could not be proven
-        return numbers.stream().reduce(0, Integer::sum) == sum;
+//          ^^^^^^^^ Related location: this is the postcondition that could not be proved
```
oyendrila-dobe, line 15: Note for improvement of error messages in a later CR: Not sure if this message is helpful to the user in identifying what was it that we had issues proving

examples/src/test/java/com/aws/jverify/examples/SumCache.java
```diff
@@ -1,36 +1,36 @@
...
-        precondition(!sumWillOverflow(sum, value));
-        modifies(this);
-        modifies(numbers);
-        numbers.add(value);
-        // sum = sum + value;
+        sum = sum + value;
+        // numbers.add(value);
         return;
-//      ^^^^^^ Error: could not prove invariant on return    
+//      ^^^^^^^ Error: a postcondition could not be proved on this return path
```
oyendrila-dobe, line 29: Similar to above, would we be able to point out to the `modifies(numbers)` in the error message?

library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -103,6 +109,21 @@ public static Object everything() {
         throw new VerificationMethodExecutedException();
     }
 
+    /**
+     * Can be used in a reads or modifies contract to make that clause abstract
+     * A subclass of the current clause can implement that abstract clause
+     * By redefining it any filling in a value
```
oyendrila-dobe: any? redefining it anytime?

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/NullableGenerator.java
```diff
@@ -116,59 +120,25 @@ public ExpressionWithFlows translateLiteral(JCTree.JCLiteral literal, IOrigin or
 
     @Override
     public ExpressionWithFlows toExprWithFlows(JCTree.JCExpression expr, IOrigin originOverride, ExpressionContext context) {
```
oyendrila-dobe, line 122: Confused by this refactoring: why did we get rid of the `JCTree.JCMethodInvocation` case?

oyendrila-dobe (dismissed):
LGTM other than the clarification comment in `NullableGenerator`

examples/src/test/java/com/aws/jverify/examples/SumCache.java
keyboardDrummer: I think it's better to consider improving error messages out of scope of this PR. It's quite a complicated topic and not particularly related to the changes here.

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/NullableGenerator.java
keyboardDrummer: The new case is more general so it replaces several cases, both the `JCMethodInvocation` and `JCAssign` one. Both the old and the new cases operated on the arguments of the invocation and the assign, but the new one matches on the argument and determines whether it needs conversion, while the old ones matches on the argument containers (invocation and assign). The new case also applies to more container than just invocation and assign so it fixes some issues.

examples/src/test/java/com/aws/jverify/examples/SumCache.java
keyboardDrummer: Maybe! We'd have to think about how to do that.

PR #353, opened 2025-10-29
Gradle verify task on sample project, support precondition(BooleanSupplier)

robin-aws:
### What was changed?
* Added a `precondition(BooleanSupplier)` overload, as block-based preconditions are often much easier to read than expressions using `?:`.
  * Also changed the return type of `isAbstract()` back to `Object` and added an `IsAbstractBoolean()` version as well, since the additional `precondition` overload made `isAbstract()` ambiguous and the translation broken.
* Added a `verify` Gradle task to the build for the `sample-project`, which provides the classpath automatically.
* Changed the `--jar` option to `--classpath`, since this is more consistent with the `javac` CLI, the option was already being used for non-jar classpath entries, and accepting classpaths rather than just individual jars made the above Gradle task much simpler to implement.

### How has this been tested?
* Updated CI to invoke `gradle verify` on the sample project
* Added a couple of basic tests of `precondition(BooleanSupplier)`


keyboardDrummer (approved):
Looks great, thanks

PR #354, opened 2025-10-29
Ignore errors in library methods when fetching mapping to original java code

oyendrila-dobe:
### What was changed?
Added check for source of error message to ensure it is from a non-library source when fetching mapping to original Java code

### How has this been tested?
Locally tested using code snippet in [this issue](https://github.com/aws/jverify/issues/352)

### DIFF
--- a/verifier/src/main/java/com/aws/jverify/verifier/Driver.java
+++ b/verifier/src/main/java/com/aws/jverify/verifier/Driver.java
@@ -28,6 +28,8 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 public class Driver {
+
+    public static Context context;
     public record VerificationResultsWithIntervalTreeMap(VerificationResults verificationResults, HashMap<URI, IntervalTree<Integer,
             JavaMethodVerificationStatus>> sourceFileToMethodIntervals) {}
 
@@ -54,7 +56,7 @@ public class Driver {
         var verificationResults = new VerificationResults();
 
         InstrumentLower.installModification();
-        Context context = new Context();
+        context = new Context();
         TypesWithoutErasure.preRegister(context);
         context.put(VerifierOptions.class, verifierOptions);
 
@@ -120,9 +122,7 @@ public class Driver {
                     continue;
                 }
 
-                var relativeURIstring = relativeUri.toString();
-                if (relativeURIstring.equals(SourceFile.URIprefix + JavaToDafnyCompiler.builtinFile) ||
-                        relativeURIstring.equals(SourceFile.URIprefix + JavaToDafnyCompiler.objectFile)) {
+                if (context.get(JavaToDafnyCompiler.class) != null && context.get(JavaToDafnyCompiler.class).isLibrary(relativeUri)) {
                     continue;
                 }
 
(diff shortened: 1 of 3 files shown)


verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -115,10 +115,17 @@ private static void outputVerificationResults(VerificationResultsWithIntervalTre
                     }
                 }
                 var relativeUri = dafnyDiagnostic.getSource();
-                var failedJavaMethod = verificationResultsWithIntervalTreeMap.sourceFileToMethodIntervals.get(relativeUri)
-                        .findAtPoint((int) dafnyDiagnostic.getLineNumber());
-                if (failedJavaMethod != null) {
-                    failedJavaMethod.setVerificationStatus(JavaMethodVerificationStatus.VerificationStatus.Failed);
+                if (relativeUri != null) {
```
keyboardDrummer: Please invert the condition to reduce nesting

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -115,10 +115,17 @@ private static void outputVerificationResults(VerificationResultsWithIntervalTre
...
+                    var relativeURIstring = relativeUri.toString();
+                    if (!relativeURIstring.equals(SourceFile.URIprefix + JavaToDafnyCompiler.builtinFile) &&
```
keyboardDrummer: Checking against `builtinFile` and `objectFile` in this location seems too specific. Do you even need this now that you've also added the null check?

oyendrila-dobe: Yes we do. The null check cannot catch a diagnostic error message due to code in the library files. I can try to make a set of these files somewhere and use them for the check to enable adding if any new library file in the future.

keyboardDrummer: > Yes we do. The null check cannot catch a diagnostic error message due to code in the library files.

Ah, thanks

> I can try to make a set of these files somewhere and use them for the check to enable adding if any new library file in the future.

Sounds good.

In `VerifyAnnotationCompiler` you are calling `if (!javaToDafnyCompiler.isLibrary(reporter.compilationUnit)) {` in the code related to method verification status, so it seems good to try and call the same method now. Seems like you would need to change the signature to use a `URI` instead of a `CompilationUnit` though.

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -127,8 +138,19 @@ private static void outputVerificationResults(VerificationResultsWithIntervalTre
...
+
+        /*
+         * checks for the following:
+         *  - the presence of non-library methods,
+         *  - lack of errors in generated Dafny code
+         *  - Dafny did not fail to run
+         *  - lack of errors in Dafny run
+         */
+        if (verificationResultsWithIntervalTreeMap.sourceFileToMethodIntervals != null //there exists
+                && verificationResultsWithIntervalTreeMap.verificationResults.getExitCode() != CommandLine.ExitCode.USAGE
```
keyboardDrummer, line 150: Nit: I would extract these subexpressions into variables with clarifying names, so that you don't need the comment.

keyboardDrummer (approved):

PR #357, opened 2025-11-07
Abstract contracts

keyboardDrummer:
### What was changed?
- Support the `isAbstract` JVerify method, that enables making method contract clauses assignable values on the type.
- Support `preconditionOf` which allows referring to the precondition of another method.
- Introduce `JVerify.range` that returns an erased `IntSequence`
- Refactoring of contract extraction in `MethodOrLoopContractCompiler`, so the Dafny specific part is moved to `BaseDafnyGenerator`

### How has this been tested?
- Added three test files for isAbstract and `preconditionOf`. `JVerify.range` is also used here


examples/src/test/java/com/aws/jverify/examples/Fibonacci.java
```diff
@@ -14,7 +14,7 @@ class Fibonacci {
     {
         postcondition((int result) -> result == spec(n));
         
-        // this precondition prevents overflow later in this method
+        // this preconditions prevents overflow later in this method
```
robin-aws: A bunch of these seem like the result of an overly-wide find-and-replace? :)

library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -137,13 +149,15 @@ public static int old(int value) {
...
-     * To clarify when it resolves to:
-     * - reference equality
-     * - shallow equality
-     * - undefined equality
+     * Provides observational equality. Two values are observationally equal if they can not be distinguished.
+     * For primitive values, jequals behaves the same as `==`
+     * For objects such as records and classes:
+     * - If the type is impure, jequals behaves like `==`, reference equality
+     * - If the type is pure, jequals behaves as structural equality, 
+     * recursively testing for observational equality on all its fields.
```
robin-aws:
```suggestion
     *   recursively testing for observational equality on all its fields. 
```
(Actually fairly important to read this line as only applying to "if the type is pure...")

library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -114,7 +126,7 @@ public static Object everything() {
      * A subclass of the current clause can implement that abstract clause
      * By redefining it and filling in a value
      */
-    public static Object isAbstract() {
+    public static <T> T isAbstract() {
```
robin-aws, line 129: Just curious, why did this help?

verifier/src/test/java/com/aws/jverify/verifier/tests/verification/abstractContracts/AllMatchContract.java
```diff
@@ -0,0 +1,60 @@
...
+    static abstract class IntStreamContract implements IntStream {
+        // Temporarily assigning this fixed value here, since Dafny does not support abstract constants in traits
+        // And JVerify does not yet support writing this as an abstract method
+        public final JVerify.IntSequence values = JVerify.range(0, 10);
+
+        @Pure
+        public boolean allMatch(IntPredicate predicate) {
+            precondition(JVerify.forall((int i) ->
+                    implies(values.contains(i),
+                            preconditionOf(predicate.test(i)))
```
robin-aws, line 46: The meaning of `preconditionOf` feels weird to me. I'm used to thinking of a precondition as an expression attached to a callable like a method or a function, rather than an expression such as a call to a method or function. In other words, I'd expect the usage to be `preconditionOf(predicate)(i)` instead.

Perhaps a term like `isWellFormed(predicate.test(i))` would be less confusing?

verifier/src/test/java/com/aws/jverify/verifier/tests/verification/abstractContracts/AllMatchContract.java
```diff
@@ -0,0 +1,60 @@
...
+            precondition(isAbstract());
+            throw new ContractException();
+        }
+    }
+
+    @Contract(IntStream.class)
+    static abstract class IntStreamContract implements IntStream {
+        // Temporarily assigning this fixed value here, since Dafny does not support abstract constants in traits
+        // And JVerify does not yet support writing this as an abstract method
+        public final JVerify.IntSequence values = JVerify.range(0, 10);
```
robin-aws: What's the version you were hoping to write? Having to work around things is fine, but I'm not yet seeing how we're hoping to eventually attach this to the particular instance of `IntStream` on line 19 rather than all `IntStream`s ever.

Could we make it a ghost `values()` method instead, and then the postcondition on `IntStream.range()` would be:

```java
postcondition((IntStreamContract r) -> jequals(r.values(), JVerify.range(startInclusive, endExclusive)));
```

library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -114,7 +126,7 @@ public static Object everything() {
      * A subclass of the current clause can implement that abstract clause
```
robin-aws: (Meant for the line above) Should say it can be used in `precondition` as well now?

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/JVerifyGhostExpressionCompiler.java
```diff
@@ -185,6 +211,10 @@ private Expression translateJVerifyMethodInvocation(JCTree.JCMethodInvocation in
                 var right = expressionCompiler.toExpr(args.get(1), ExpressionContext.Pure);
                 return new BinaryExpr(origin, BinaryExprOpcode.Eq, left, right);
             }
+            case "isAbstract" -> {
+                reporter.reportError(invocation.getMethodSelect(), "notSupported", "isAbstract() in clauses other than a preconditions");
```
robin-aws: ...or reads or modifies?

verifier/src/test/java/com/aws/jverify/verifier/tests/verification/abstractContracts/AllMatchContract.java
```diff
@@ -0,0 +1,60 @@
...
+            return value == 2;
+        });
+        check(true);
+    }
+
+    @Contract(IntPredicate.class)
+    static class IntPredicateContract implements IntPredicate {
+        @Pure
+        public boolean test(int value) {
+            precondition(isAbstract());
```
robin-aws, line 28: I think I get `isAbstract()` and it seems to read well. I assume this is necessary because not declaring any `precondition` clauses is equivalent to `precondition(true)` and therefore can't be overridden? I think that's worth documenting somewhere explicitly.

verifier/src/test/java/com/aws/jverify/verifier/tests/verification/abstractContracts/AllMatchContract.java
```diff
@@ -0,0 +1,60 @@
...
+import static com.aws.jverify.JVerify.*;
+
+@JVerifyTest(dafnyVerified = 9, dafnyErrors = 0)
+public class AllMatchContract {
+
+    @SuppressWarnings("ConstantValue")
+    void foo() {
+        var temp = IntStream.range(0, 10);
+        var r = temp.allMatch(value -> {
+            precondition(cast(temp, IntStreamContract.class).values.contains(value));
```
robin-aws: This does make total sense to me as a solution!

robin-aws (commented):
Just reviewed the user-facing API changes to start, will review the implementation in a second pass.

library/src/main/java/com/aws/jverify/JVerify.java
keyboardDrummer: When something expects a boolean, like `precondition`, `Object` doesn't cut it.

examples/src/test/java/com/aws/jverify/examples/Fibonacci.java
keyboardDrummer: Went through all of them.

verifier/src/test/java/com/aws/jverify/verifier/tests/verification/abstractContracts/AllMatchContract.java
keyboardDrummer: > I'm used to thinking of a precondition as an expression attached to a callable like a method or a function

Don't operators like `.` or `/` have preconditions too?

`preconditionOf` could have as its meaning: return the precondition syntactically required by its expression argument. Currently it only works if the argument is a method call though.

> Perhaps a term like isWellFormed(predicate.test(i)) would be less confusing?

I would rather avoid the term wellformed altogether. I had never encountered this term in a programming context before I got into formal methods, so I don't think it works well for the general programmer.

verifier/src/test/java/com/aws/jverify/verifier/tests/verification/abstractContracts/AllMatchContract.java
keyboardDrummer: > Could we make it a ghost values() method instead

Yes that's what I meant with the comment, but that requires an extra change to allow `@Contract` classes to define methods that don't relate to the contractee. I can add that change though.

verifier/src/test/java/com/aws/jverify/verifier/tests/verification/abstractContracts/AllMatchContract.java
keyboardDrummer: Yay :)

library/src/main/java/com/aws/jverify/JVerify.java
keyboardDrummer: Updated it so it only mentions `precondition`, since for now this only works for precondition clauses. I will expand it to `reads` clauses as well, but later. I'm not a 100% sure it makes sense to support this for all clauses.

verifier/src/test/java/com/aws/jverify/verifier/tests/verification/abstractContracts/AllMatchContract.java
keyboardDrummer: > I think that's worth documenting somewhere explicitly.

Yes, I'll add a section to the user guide.

verifier/src/test/java/com/aws/jverify/verifier/tests/verification/abstractContracts/AllMatchContract.java
robin-aws: > Don't operators like . or / have preconditions too?

Totally! I'm just more used to saying that the operator `/` has a precondition, not that the expression `3 / x` has a precondition. I'll fully admit it might be just fine to use it that way in general though. We may still want to adjust the way we use it in the user guide too, to say that calling `precondition(...)` is a way to specify what the precondition of a an expression that calls the method will be.

> I would rather avoid the term wellformed altogether. I had never encountered this term in a programming context before I got into formal methods, so I don't think it works well for the general programmer.

Yeah I agree.

verifier/src/test/java/com/aws/jverify/verifier/tests/verification/abstractContracts/AllMatchContract.java
robin-aws: Good call, I think that features be generally very useful for many cases anyway.

docs/guide/src/abstract_contracts.md
```diff
@@ -0,0 +1,37 @@
+# Abstract contracts
+When working with abstract methods, you might not yet be sure what the contract of the method will be. For this situation,  you can define abstract contracts using the method `isAbstract`. Here's an example where an abstract contract is used:
+
+```java
+@Contract(IntPredicate.class)
+class IntPredicateContract implements IntPredicate {
+    @Pure
+    public boolean test(int value) {
+        // Having the following line is different from not calling 'precondition' at all,
+        // because that would mean there is precondition, equivalent to 'precondition(true)'
```
robin-aws:
```suggestion
        // because that would mean there is a precondition, equivalent to 'precondition(true)'
```

library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -66,6 +66,15 @@ public static void postcondition(BooleanPredicate predicate) {
     public static void postcondition(IntPredicate predicate) {
     }
 
+    /**
+     * Return the precondition of the given expression
+     * Currently only works when the expression is a method call
+     * and only works for method calls that return a value.
+     */
+    public static <T> boolean preconditionOf(T value) {
```
robin-aws, line 74: I'm definitely warming to this more general use of "precondition". I wonder if we could even eventually support expressions with no result via something like `preconditionOf(this::methodReturningVoid)` or `preconditionOf(() -> { methodReturningVoid(1, 2); }`.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -196,6 +196,9 @@ public static void verifyFile(SourceFile sourceFile, JVerifyTest annotation, Lis
                 .forEach(dafnyOutput -> {
                     URI source = ((DafnyDiagnostic) dafnyOutput).getSource();
                     var methodIntervals = verificationResultsWithMethodIntervals.sourceFileToMethodIntervals().get(source);
+                    if (methodIntervals == null) {
```
robin-aws: When does this happen? I'm nervous about silently dropping verification results if something went wrong.

docs/guide/src/abstract_contracts.md
robin-aws, line 1: Nice!

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/JavaToDafnyCompiler.java
```diff
@@ -192,24 +192,25 @@ public void finished(TaskEvent e) {
                     // Apply the second half of our pipeline as above (4 and onwards).
                     // See the implementation of JavaCompiler.compile() for similar lines,
                     // including the comment "these method calls must be chained to avoid memory leaks"
```
robin-aws: Comment out of date with this refactoring. Not objecting to the refactoring since this version is obviously cleaner, and if we start to hit memory leaks we can always change it back.

robin-aws (changes requested):
A couple of cleanup things to take care of. Need a bit more time to skim the rest of the implementation but not anticipating finding anything blocking, the testing coverage looks good.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/IsAbstractCompiler.java
```diff
@@ -0,0 +1,151 @@
...
+            return null;
+        }
+        return (JCTree.JCMethodDecl) index.getTree(baseSymbol);
+    }
+
+    private void updatePrecondition(JCTree.JCMethodDecl tree, MethodOrLoopContract contract, 
+                                    Symbol.MethodSymbol preconditionFunction) {
+        var thisSymbol = utils.thisSymbol(tree.sym);
+
+        JCTree.JCMethodInvocation application = treeMaker.Apply(null /* TODO */,
```
robin-aws: TODO?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/PreconditionOfCompiler.java
```diff
@@ -0,0 +1,103 @@
...
+import com.sun.tools.javac.code.*;
+import com.sun.tools.javac.tree.*;
+import com.sun.tools.javac.util.*;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Handle 'preconditionOf' calls
```
robin-aws, line 17: Worth describing what transformation it actually does

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/Property.java
```diff
@@ -0,0 +1,57 @@
...
+            @Override
+            public T get() {
+                return list.get(index);
+            }
+
+            @Override
+            public void set(T value) {
+                List<T> current = list;
+                //noinspection StatementWithEmptyBody
+                for(var i = 0; i < index; i++, current = list.tail) {
```
robin-aws: `current = list.tail` is a bug, should be `current = current.tail`.

And stylistically, I know List.get itself does the empty body for loop thing, but I'd much prefer:

```dafny
var i = 0;
while (i < index) {
  list = list.tail;
  i++;
}
```

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/Property.java
```diff
@@ -0,0 +1,57 @@
+package com.aws.jverify.verifier.compiler.simplifications;
+
+import com.sun.tools.javac.util.List;
+
+import java.util.ArrayList;
+
+public interface Property<T> {
```
robin-aws, line 7: Consider having this and `finalProperty` in the common library, and just the `fromElement` implementations in this package.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MethodOrLoopContractCompiler.java
```diff
@@ -239,121 +256,90 @@ public JCTree.JCBlock getDefaultBody() {
...
-            // Casts like (IntPredicate) are sometimes necessary to disambiguate
-            handlePostcondition(baseGenerator, header, typeCast.getExpression());
-        } else {
-            var dafnyExpr = baseGenerator.expressionCompiler.toExpr(expr, ExpressionContext.Pure);
-            header.postconditions.add(new AttributedExpression(dafnyExpr, null, null));
+    public void checkEmptyExpressions(JCTree tree,
+                                      java.util.List<Property<JCTree.JCExpression>> expressions,
+                                      String typeName,
+                                      String containerName) {
+        for(var _ : expressions) {
```
robin-aws: I don't understand what this is doing. Whatever it is, a for loop that prints the same error N times without using the expressions at all can't be right.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MethodOrLoopContract.java
```diff
@@ -1,47 +1,43 @@
...
+    public static JCTree.JCExpression combineClauses(Context context, 
+                                                     List<Property<JCTree.JCExpression>> clauses) {
+        var treeMaker = TreeMaker.instance(context);
+        Attr attr = Attr.instance(context);
+        Enter enter = Enter.instance(context);
+        Reporter reporter = Reporter.instance(context);
+
+        return clauses.stream().map(Property::get).
+                reduce((a,b) -> {
+                    // Use temporary arguments to binary so we can easily attribute Binary
```
robin-aws: Don't follow why this helps

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/IsAbstractCompiler.java
```diff
@@ -0,0 +1,151 @@
...
+import com.sun.tools.javac.code.*;
+import com.sun.tools.javac.model.JavacElements;
+import com.sun.tools.javac.tree.*;
+import com.sun.tools.javac.util.*;
+import java.util.Set;
+
+/**
+ * Must run before lower
+ * 
+ * Handles 'isAbstract()' calls that can be used at the root of contract clauses
```
robin-aws: Worth documenting how they are handled exactly

robin-aws (changes requested):
Finished reviewing the implementation.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/MethodOrLoopContract.java
keyboardDrummer: Provided a much more complete explanation.

library/src/main/java/com/aws/jverify/JVerify.java
keyboardDrummer: > I wonder if we could even eventually support expressions with no result

Yes we should

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/Property.java
keyboardDrummer: Why?

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
keyboardDrummer: I don't know. I'll remove it for now.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/Property.java
robin-aws: Pretty useful generic interface (I was a bit surprised something like this isn't in the JCL at this point). But can be moved later if we find more uses.

robin-aws (approved):

PR #361, opened 2025-11-20
Lambda in postcondition

keyboardDrummer:
### What was changed?
- Enable using lambdas inside contracts

### How has this been tested?
 - Added an additional test


verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/LambdaToAnonymousClassCompiler.java
```diff
@@ -36,12 +36,38 @@ public LambdaToAnonymousClassCompiler(JCCompilationUnit compilationUnit, Context
...
         if (jverifyMethod == null) {
-            super.visitApply(invocation);
+                super.visitApply(invocation);
         } else {
+            invocation.meth = translate(invocation.meth);
+            if (invocation.args != null) {
+                for (List<JCExpression> l = invocation.args; l.nonEmpty(); l = l.tail) {
+                    // Our pipeline might have inserted a typecast at this point
+                    // So that postcondition((int x) -> x == 2)) becomes
+                    // postcondition((IntPredicate)(int x) -> x == 2);
```
robin-aws, line 55: FYI these kinds of casts are also sometimes necessary in user code to disambiguate overloads like `BooleanSupplier` vs. `Supplier<Boolean>`, so it's good if we can handle them in general.

robin-aws (approved):
LGTM!

PR #363, opened 2025-11-25
Various fixes

keyboardDrummer:
### What was changed?
1. Add `--include-filter-dependencies` option to allow verifying an entire dependency tree
2. Make `--position-filter` better handle different line ranges
3. Allow use of `? extends super` in method signatures. This was needed to support `Stream.map`.
4. Add `EmptyContract` concept.
5. When lambdas capture variables, allow the captured values to be used for verification
6. Show compilation time information when running with `--track-time`
7. Fail earlier when there are errors in `additional.dfy`
8. Tighten checks on the implementations that occur in contract methods
9. Enable returning a contract class in a method, like `StreamContract.fromStream`
10. Let parameter annotations, like `@Nat`, be used in external contract classes.
11. Improve support for IntSequence
12. Improve support for relative filepaths
13. Support long string literals by no longer verifying the encoding of Dafny String literals, since it is correct by construction.

### How has this been tested?
1. Added test for `--include-filter-dependencies`
2. No test added. Relates strongly to the Dafny integration
3. Updated WildcardsNotSupported to use `? super <type>` and add Stream.map definition to built-ins
4. Added use of `@EmptyContract` in tests
5. Extended the Lambdas test with cases that would previously fail
6. Added test for `--track-time`
7. No test added
8. Added test-case in `LibraryContractGhostFieldErrors`
9. Added test-case for `StreamContract.fromStream` to `LibraryContractGhostFieldVerification`
10. Add test-case related to `@Nat` usage to `LibraryContractGhostFieldVerification`
11. No test added
12. No test added
13. Added long String literal to `Strings.java`


library/src/main/java/com/aws/jverify/EmptyContract.java
```diff
@@ -0,0 +1,14 @@
+package com.aws.jverify;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Can be placed on a bodyless method to indicate that this method has no pre- or postconditions.
```
robin-aws, line 9: Should specify explicitly what "empty" means for the other clauses as well: reads/modifies/decreases etc.

library/src/main/java/com/aws/jverify/JVerify.java
```diff
@@ -44,7 +40,7 @@ public static void precondition(boolean condition) {
      * in the same way as two tuples would be compared, 
      * also known as a lexicographical comparison.
      */
-    public static void decreases(Object... values) {}
+    public static void decreases(Object value) {}
```
robin-aws, line 43: Why this change? That's strictly less powerful.

docs/guide/src/interfaces.md
```diff
@@ -63,3 +63,18 @@ abstract class ListContract<T> implements List<T> {
 We only wanted to provide a contract for `size`, so we made ListContract an _abstract_ class. This way, the Java compiler is satisfied even though we did not implement all methods from the interface `List<T>`.
 
 Contrary to the `interface I` example from before, because `List<T>` is not defined in source, annotations such as `@Modifiable` are taken from the contract class.
+
+### Empty contracts
+
+You can indicate that a bodyless method has an empty contract using the `@EmptyContract` attribute, to avoid having to define a contract class. Example:
```
robin-aws, line 69: Nice feature!

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -362,7 +369,7 @@ public static VerifierOptions getVerifierOptions(JVerifyTest annotation, Positio
                 },
                 annotation.verifyByDefault(),
                 annotation.continueOnErrors(),
-                positionFilter, false
+                positionFilter, true, false
```
robin-aws, line 372: I think we'll want to make a builder class for `VerifierOptions` soon, starting to worry about mixing up arguments in the future...

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/wildcards/WildcardsNotSupported.java
```diff
@@ -5,32 +5,39 @@
 import com.aws.jverify.Pure;
 import com.aws.jverify.testengine.JVerifyTest;
 
-@JVerifyTest(exitCode = 2)
+import java.nio.file.ReadOnlyFileSystemException;
```
robin-aws: I have to assume autocorrect went haywire and this is unused :)

verifier/src/test/java/com/aws/jverify/verifier/tests/jcl/Strings.java
```diff
@@ -14,8 +14,10 @@
         "OnlyOneElementUsed",
         "StringOperationCanBeSimplified"
 })
-@JVerifyTest(exitCode = 4, dafnyVerified = 14, dafnyErrors = 6)
+@JVerifyTest(exitCode = 4, dafnyVerified = 13, dafnyErrors = 6)
 class Strings {
+    static final String veryLongLiteral = "aeriuhgrae;igheraiugher;ughae;oigjaer;oigjaei;ogjera;ogj;eaorjga";
```
robin-aws: Cat-like typing detected. :)

Do you need a reference to this somewhere to make sure it isn't trimmed out in the future?

verifier/src/test/java/com/aws/jverify/verifier/tests/verification/abstractContracts/AllMatchContract.java
```diff
@@ -3,8 +3,10 @@
 import com.aws.jverify.*;
 import com.aws.jverify.testengine.JVerifyTest;
 
+import java.util.function.IntFunction;
```
robin-aws: Unused?

We should hook up spotless in the Gradle build.

robin-aws (changes requested):
Reviewed about half, will finish tomorrow

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -119,17 +128,19 @@ private static void outputVerificationResults(VerificationResultsWithIntervalTre
...
 
-                if (context.get(JavaToDafnyCompiler.class).isLibrary(relativeUri)) {
+                if (context.get(JavaToDafnyCompiler.class).isBuiltin(relativeUri)) {
                     continue;
                 }
 
-                var failedJavaMethod = verificationResultsWithIntervalTreeMap.sourceFileToMethodIntervals.get(relativeUri)
-                        .findAtPoint((int) dafnyDiagnostic.getLineNumber());
+                var uriMethods = verificationResultsWithIntervalTreeMap.sourceFileToMethodIntervals.get(relativeUri);
+                if (uriMethods == null) {
```
robin-aws: Why does this happen? (Again paranoid about skipping verification)

verifier/src/main/java/com/aws/jverify/verifier/Main.java
```diff
@@ -52,10 +49,13 @@ class AppCommand implements Callable<Integer> {
     
     @Option(names = "--paths", description = "Show file paths instead of names")
     private boolean paths;
-
+    
     @Option(names = "--filter-position", description = "Filter what gets verified based on a source location. The location is specified as a file path suffix, optionally followed by a colon and a line number or line range. For example, `jverify verify source1.java source2.java --filter-position=source1.dfy:5-7` will only verify things that between (and including) line 5 and 7 in the file `source1.dfy`. You can also use `:5`, `:5-`, `:-5` to specify individual lines or open ranges.")
     private String filterPositionString;
     
+    @Option(names = "--include-filter-dependencies", description = "When filtering on just a file, also verify its transitive dependencies.")
```
robin-aws, line 56: Would it be cleaner to just make this an independent `--include-dependencies` option that could be used without a filter? We could just block using it together with a line filter if that's not supported.

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/TypeDeclarationCompiler.java
```diff
@@ -258,7 +258,7 @@ public Field translateField(JCTree.JCVariableDecl variableDecl) {
 //        if (varFlags.contains(Modifier.FINAL)) {
 //            Expression rhs = null;
 //            if (variableDecl.getInitializer() != null) {
-//                rhs = compiler.expressionCompiler.toExpr(variableDecl.getInitializer(), ExpressionContext.Pure);
+//                rhs = expressionCompiler.toExpr(variableDecl.getInitializer(), ExpressionContext.Pure);
```
robin-aws, line 261: Why is the commented out code checked in the first place?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ExternalContractCompiler.java
```diff
@@ -278,32 +280,60 @@ private void updateLibraryContractAnnotations(JCTree.JCMethodDecl contracter,
...
             newAnnotations.addAll(contracterSymbol.getAnnotationMirrors());
-            if (!shouldVerify(contracter, contracterSymbol)) {
-                contracter.mods.annotations = contracter.mods.annotations.append(jverifyUtils.getVerifyFalseAnnotation());
-                newAnnotations.add(jverifyUtils.getVerifyAnnotation());
-            }
             contracteeSymbol.resetAnnotations();
             contracteeSymbol.setDeclarationAttributes(newAnnotations.toList());
+
+            com.sun.tools.javac.util.List<Symbol.VarSymbol> parameters = contracteeSymbol.getParameters();
+            for (int i = 0; i < parameters.size(); i++) {
```
robin-aws, line 289: Could be a simple enhanced for loop? The whole thing feels very AI :)

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/IsAbstractCompiler.java
```diff
@@ -4,6 +4,7 @@
 import com.aws.jverify.Nullable;
 import com.aws.jverify.verifier.compiler.Reporter;
 import com.aws.jverify.verifier.compiler.frontend.JVerifyIndex;
+import com.aws.jverify.verifier.compiler.frontend.JavaToDafnyCompiler;
```
robin-aws: Unused?

library/src/main/java/com/aws/jverify/EmptyContract.java
robin-aws: Also wondering if "DefaultContract" would be more accurate? Given the defaults for some clauses like `reads` and `decreases` aren't really "empty"

verifier/src/main/resources/additional.dfy
```diff
@@ -39,8 +40,8 @@ function toSequence<T>(arr: JArray<T> ): (r: seq<T>)
   
 datatype Nullable<T> = NonNull(value: T) | Null
 
-function intSeqRange(from: int, to: int): seq<int> 
-  requires to >= from
+ghost function intSequenceRange(inclusiveFrom: int, exclusiveTo: int): seq<int>
```
robin-aws, line 43: Just curious, why was this necessary? Should we just be making all translated Dafny code ghost?

verifier/src/main/resources/builtin-contracts.java
```diff
@@ -54,6 +67,25 @@ abstract class StreamContract<T> implements Stream<T> {
...
+                        forall((int index) -> implies(0 <= index && index < elements.size(),
+                                jequals(mapper.apply(elements.get(index)), r.elements.get(index)))));
+        throw new ContractException();
+    }
+    
+    @Erased
+    @Pure
+    public static <T> StreamContract<T> fromStream(Stream<T> stream) {
+        return JVerify.<Stream<T>, StreamContract<T>>cast(stream);
+    }
```
robin-aws, line 88: Bit off topic/ranty, but I'm not loving the UX of this approach. I don't think the name "cast" matches the semantics very well, since it's not really down- or upcasting in the type hierarchy, it's retrieving a parallel object. And it seems like many contracts will have to declare a static method like this in practice. I don't like using "from" in the static method because it implies it's creating a new object, but it was implicitly created at the same time as the "actual" object.

Probably something we should think about independently, maybe an issue for it?

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/DropUnreachableUnits.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+
+    private Set<JCTree.JCCompilationUnit> scanForDirectDependencies(JCTree.JCCompilationUnit unit) {
+        DependencyVisitor visitor = new DependencyVisitor();
+        unit.accept(visitor);
+        return visitor.getDependencies();
+    }
+
+    private class DependencyVisitor extends TreeScanner {
```
robin-aws, line 84: I'm skeptical this is complete enough. How can we test thoroughly enough to know?

Perhaps something to add to https://github.com/aws/jverify/issues/234?

robin-aws (changes requested):

library/src/main/java/com/aws/jverify/JVerify.java
keyboardDrummer: The previous API wasn't working because we don't support varargs.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
keyboardDrummer: Shouldn't the IDE show inline argument names instead? Seems like it would provide the same benefits at a lower cost. Note that Q managed to do this refactoring on its own, so either way is fine with me.

verifier/src/test/java/com/aws/jverify/verifier/tests/jcl/Strings.java
keyboardDrummer: I'll add `@SuppressWarnings("unused")`

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
keyboardDrummer: Don't recall. I'll remove it.

verifier/src/main/java/com/aws/jverify/verifier/Main.java
keyboardDrummer: What would it do when used without `--filter` ? Right now it causes less to be filtered, so if you're not filtering it wouldn't do anything.

verifier/src/main/java/com/aws/jverify/verifier/compiler/dafnygenerator/base/TypeDeclarationCompiler.java
keyboardDrummer: I sometimes comment this back in only to realize I can't do that yet ^^

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ExternalContractCompiler.java
keyboardDrummer: `i` is referred to twice, so I think not unless we zip first.

> The whole thing feels very AI :)

I don't recall but looking at it, it looks good to me.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/IsAbstractCompiler.java
keyboardDrummer: Happy to process these, but we can also leave them in until we add an automatic solution :)

verifier/src/main/resources/additional.dfy
keyboardDrummer: > Should we just be making all translated Dafny code ghost?

I think we already are! Which might also be the reason that this had to be ghost as well.

verifier/src/main/resources/builtin-contracts.java
keyboardDrummer: > Bit off topic/ranty, but I'm not loving the UX of this approach.

That's fine :) 

> Probably something we should think about independently, maybe an issue for it?

How about we rename both to `asContractType` ? The inner cast could also be used in the other direction, but it seems rare to use it like that.

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/DropUnreachableUnits.java
keyboardDrummer: > I'm skeptical this is complete enough.

You're right it's definitely missing cases.

> Perhaps something to add to https://github.com/aws/jverify/issues/234?

How would such a bug affect soundness? If we delete a dependency then downstream compilation will fail, won't it?

library/src/main/java/com/aws/jverify/EmptyContract.java
keyboardDrummer: There's no fixed default also for Dafny right? I think it might infer a decreases clause based on a reads clause, and besides that I'm not sure. Do reads/modifies always stay empty? I'm also not sure what we'll replicate with Laurel.

Without more information, I find it hard to decide on the right name and documentation now, but it's still useful to have this, even for ourselves.

library/src/main/java/com/aws/jverify/JVerify.java
robin-aws: Okay good reason :)

library/src/main/java/com/aws/jverify/EmptyContract.java
robin-aws: I mean it to read as "the default contract for this method" rather than a fixed default. But I can live with EmptyContract meaning "the contract you get when the method header is empty. :)

verifier/src/main/java/com/aws/jverify/verifier/Main.java
robin-aws: I was thinking about this like Dafny's `--verify-includes` option. But I realize now it's not the same, since the dependencies here means the other source files the filter depends on, but not the libraries, so never mind.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/ExternalContractCompiler.java
robin-aws: Oh I didn't spot the one-letter difference between `contracteeSymbol` and `contracterSymbol` so it looked unnecessary. Disregard.

verifier/src/main/resources/builtin-contracts.java
robin-aws: I definitely like `asContractType` better! I originally wanted to call this `contractOf` which I might like even better. :)

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/DropUnreachableUnits.java
robin-aws: True, the worst that can happen is errors you didn't expect or much longer processing.

robin-aws (approved):

PR #366, opened 2025-12-11
Add fp32 support and use unchecked methods for floating-point operations

fabiomadge:
Adds fp32 (Java float) support and adapts both fp32/fp64 to use Dafny's unchecked static methods, bypassing wellformedness checks while maintaining Java's IEEE 754 semantics.

**Changes:**
- Add fp32 type support
- Use unchecked methods (fp32.Add, fp64.Less, etc.) for all floating-point operations
- Optimize int literal casts (0 → 0.0)
- Add comprehensive fp32 tests
- Refactor registration code to eliminate duplication

**Testing:** All 88 float/double tests pass


verifier/src/test/java/com/aws/jverify/verifier/tests/jcl/Floats.java
```diff
@@ -0,0 +1,112 @@
+package com.aws.jverify.verifier.tests.jcl;
```
keyboardDrummer: Seems inconsistent to have this file here instead of in `javasupport/expressions`

keyboardDrummer (dismissed):
Looks great, added just one comment

verifier/src/test/java/com/aws/jverify/verifier/tests/jcl/Floats.java
fabiomadge: A lot of this test is about Java Class Library features. Should we also move the Double equivalent?

keyboardDrummer: I don't have a strong opinion on this, but even though strictly speaking the boxed primitives are part of the JCL, it might be simplest to consider everything related to primitives and boxed primitives to be part of the language instead of the JCL, so that you don't get a split between boxed and unboxed primitives.

Currently there's a lot related to primitives in the non-JCL part of the tests:

image

so I would suggest we keep all the numeric tests there. If you feel strongly that something should be in the JCL part of the tests I'm also happy to go in that direction but then we should have clear rules that allow us to be consistent.

keyboardDrummer (approved):

PR #367, opened 2025-12-16
Contract path CLI option

robin-aws:
### What was changed?
Added a `--contract-path` CLI option, which can be used to provide source directories or source jars containing contract definitions. This should work well in the general Java ecosystem, as repositories such as Maven Central have deep support for publishing the source of compiled libraries as parallel `<library>-source.jar` artifacts, and IDEs automatically fetch these jars in order to display the source of libraries. This provides a natural way for users to read the contracts of libraries: performing an "Open Type" action on "BigIntegerContract" in IntelliJ for example, will display the contract source.

Moved the `builtin-contract.java` resource into a proper separate Gradle package with the simple `withSourcesJar()` directive, and removed the `--builtin-contracts` option. This might be controversial and I'm more open to debate on it:
  * Pros
    * More "drinking our own champagne" by maintaining the JCL contracts in the same way users will write contracts
    * The built-in contracts can be developed with normal idiomatic Java tooling (e.g. there was an unused import that was only flagged in my IDE after this change)
    * Natural testing coverage of the `--contract-path` feature without having to create more dummy packages
  * Cons
    * Makes the "getting started" command even more unwieldy. I'm not personally bothered by this because I don't expect any serious users of JVerify to call the CLI directly, but to use our to-be-written Gradle plugin to calculate both the --classpath` and `--contract-path` arguments.
    * Users will need to be aware of needing to add `builtin-contracts` to rely on any contracts on the JCL.

Note that we could still have a `--builtin-contracts` option, but would need to add more build smarts to embed the package as a resource in the `verifier` package. I would still prefer to keep a cleaner architecture and let the Gradle plugin or other higher-level tooling provide this option instead.

### How has this been tested?
Used for the `builtin-contracts` package in all the tests, either the source directory (in the unit tests) or the built source jar (in the `sample-project`).


verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/JavaToDafnyCompiler.java
```diff
@@ -59,15 +62,54 @@ public JavaToDafnyCompiler(Context context) {
...
+                            )
+                            .forEach(source -> {
+                                files.add(source);
+                                builtinSources.add(source);
+                            });
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
+            } else {
+                try {
```
keyboardDrummer, line 89: Is this else branch just for testing purposes? Would it also be possible to use the then branch for testing instead?

keyboardDrummer (approved):
Once again, @robin-aws your PR looks amazing!

I look forward to splitting up builtin-contracts.java into separate files as well. I'm fine with removing `--builtin-contracts`. Regarding having a separate or unified `--classpath` and `--contract-path` option, I'm not sure.

Regarding the choice for source jars. For the built-in contracts, having all the original source as part of the contract library seems fine. For libraries where the contracts and sources are specified in the same file, I can imagine it would be more ideal if opaque bodies in the library were not part of the contract library content.

What I had initially imagined is that we would put the contract information in the runtime jars, similar to how non-erased method type signatures are already stored in there. However, the contract information is much larger in size than method contracts, and for methods with transparent bodies, would be similar or the same as the source, so maybe putting that information in source-jars makes more sense. I think this does break with the intention of source-jars. They are intended just for reading sources, while runtime jars and meant for running _and_ building.

Regardless of what I said above, this PR implements an important feature in a concise way, so that's great. Let's try this approach and if somehow it isn't what we want, we can think about it more.

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/JavaToDafnyCompiler.java
```diff
@@ -59,15 +62,54 @@ public JavaToDafnyCompiler(Context context) {
...
+            if (resolvedContractsPath.toString().endsWith(".jar")) {
+                try (var jarFile = new JarFile(resolvedContractsPath.toFile())) {
+                    jarFile.stream()
+                            .filter(jarEntry -> jarEntry.getName().endsWith(".java"))
+                            .map(jarEntry ->
+                                    new JarFileEntry(resolvedContractsPath, jarEntry)
+                            )
+                            .forEach(source -> {
+                                files.add(source);
+                                builtinSources.add(source);
```
keyboardDrummer, line 83: Maybe the variable `builtinSources` should be renamed.

keyboardDrummer (approved):

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/JavaToDafnyCompiler.java
robin-aws: I think it's also a reasonable option for local development of multiple packages. It's useful to support directories on the regular classpath so you can have one local package that depends on the classes of another one build without having to create a jar first. I expect when we have a Gradle plugin we can leverage Gradle to do this automatically for the contract path too.

It would be somewhat tricky to use jars in testing, because running `gradle test` on the `verifier` won't automatically build `builtin-contracts` deeply enough to produce the source jar. It works for the sample project build in CI because it does `gradle publishToMavenLocal` first (which is accurately imitating how users will consume that contract library).

verifier/src/main/java/com/aws/jverify/verifier/compiler/frontend/JavaToDafnyCompiler.java
robin-aws: True, will fix in a followup PR.

robin-aws:
> I look forward to splitting up builtin-contracts.java into separate files as well.

Definitely, had the same thought. In fact I'll do that in the follow-up PR right away. The longer we wait the more potential annoying merge conflicts. :)

> I think this does break with the intention of source-jars. They are intended just for reading sources, while runtime jars and meant for running and building.

Yeah I agree to a degree. But I also think the source jars are pretty essential for using and understanding libraries in practice, so I'm not too worried about static analysis requiring them as well. As I said above I think having the contract source published is important anyway, so if we change our minds we'd want to keep doing that anyway.

PR #369, opened 2025-12-29
Support of native assert in JVerify

olivier-aws:
## Summary

JVerify now supports verification of Java `assert` statements in addition to `JVerify.check()` calls. Both produce identical `assert` statements and are equivalently verified.

### Problem
Java's `assert` statements are transformed by javac's LOWER compilation phase into conditional throw statements before JVerify can process them. This prevented JVerify from verifying assertions and the verification of every program using an `assert` would end in an exception `error: statement JCThrow is not supported`

### Solution
The support for `assert` statements is implemented using the Suspenders mechanism in `verifier/src/main/java/com/aws/jverify/verifier/compiler/Suspenders.java`:

1. **SUSPEND Phase** (before LOWER): Transform `assert condition;` into a marked if-statement that LOWER won't recognize
2. **LOWER Phase**: Runs without transforming our suspended asserts
3. **UNSUSPEND Phase** (after LOWER): Transform the marked if-statements back to `assert condition;`
4. **Translation**: `BlockCompiler.java` translates `JCTree.JCAssert` to Dafny `AssertStmt`

The transformation uses a special marker name `$assertMarker` to identify suspended assert statements:
- SUSPEND: `assert condition;` → `if ($assertMarker) { condition; }`
- UNSUSPEND: `if ($assertMarker) { condition; }` → `assert condition;`

This is similar to how Suspenders handles switch statements and records to prevent unwanted lowering.

### Code Changes

**Suspenders.java**:
- Added `assertMarkerName` field to mark suspended assertions
- Added `visitAssert()` method to suspend assert statements:
  - Creates a synthetic `VarSymbol` with the marker name
  - Transforms `assert condition;` to `if ($assertMarker) { condition; }`
- Modified `visitIf()` method to unsuspend marked if-statements back to asserts

## Usage

You can use either `JVerify.check()` or Java's native `assert` statement:

### Using JVerify.check()
```java

class Example {
    static void foo(int x) {
        precondition(x > 0);
        check(x > 0);  // Verified
        check(x >= 1); // Verified
    }
}
```

### Using Java assert
```java
class Example {
    static void foo(int x) {
        precondition(x > 0);
        assert x > 0;  // Verified
        assert x >= 1; // Verified
    }
}
```

## Test Cases

The following test cases demonstrate assert statement verification:

1. **AssertVsCheck.java** - Demonstrates equivalence of assert and check()


verifier/src/test/java/com/aws/jverify/verifier/tests/AssertVsCheck.java
```diff
@@ -0,0 +1,36 @@
...
+import com.aws.jverify.testengine.JVerifyTest;
+
+import static com.aws.jverify.JVerify.*;
+
+/**
+ * Demonstrates that assert statements and check() calls are equivalent
+ * for verification purposes.
+ */
+@JVerifyTest(exitCode = 0, dafnyVerified = 5, dafnyErrors = 0)
+class AssertVsCheck {
```
keyboardDrummer, line 12: I would put this in `tests/javasupport/statements` since it relates to the assert statement

keyboardDrummer (approved):
An alternative implementation would be to use `InstrumentLower` and `LowerInterceptor` to turn off the transformation of the assert statement. It would allow `assert true` to become like `check(true)`, but the current implementation is fine too.

From a user perspective, does it make sense for `assert` statements to be verified? My impression is that the answer is yes, because the use-case of asserts seems to be to use them like contracts and turn them on during testing.

olivier-aws:
I agree that it makes sense to have the `assert` statements verified like check. I'll move the test into the appropriate folder in a follow-up CR

PR #370, opened 2026-01-02
Split Driver.java into back-end specific and agnostic part

keyboardDrummer:
### What was changed?
- Split `Driver` into DafnyDriver and `Driver` (which is now agnostic to Dafny).
- Add the Strata repository as a submodule
- Generate AST types for minimal Laurel

### How has this been tested?
- Only existing tests for now, although the test expectations have been adapted so they're not Dafny specific.


verifier/src/main/java/com/aws/jverify/verifier/Wrapper.java
```diff
@@ -0,0 +1,17 @@
+package com.aws.jverify.verifier.dafny;
+
+public class Wrapper<T> {
```
fabiomadge, line 3: This doesn't seem specific to Dafny

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -49,152 +32,32 @@ public static int verifyJavaPaths(List<Path> files, VerifierOptions verifierOpti
...
-            if (failedCount > 0) {
-                outputWriter.write(String.format("%sFailed: %s", bullet, failedCount));
-                outputWriter.write('\n');
-            }
+        
+        var verificationResults = results.verificationResults();
+        if (verificationResults != null) {
+            pw.println(String.format("Found %s errors", verificationResults.verificationFailedAssertions()));
+            pw.println(String.format("Verified methods: %s", verificationResults.verificationPassedMethods()));
+            pw.println(String.format("Failed methods: %s", verificationResults.verificationFailedAssertions()));
```
fabiomadge:
```suggestion
            pw.println(String.format("Failed methods: %s", verificationResults.verificationFailedMethods()()));
```

verifier/src/main/java/com/aws/jverify/verifier/dafny/DafnyDriver.java
```diff
@@ -0,0 +1,339 @@
...
+                                if (relativeUri == null) {
+                                    return;
+                                }
+
+                                if (context.get(JavaToDafnyCompiler.class).isContractSource(relativeUri)) {
+                                    return;
+                                }
+
+                                var uriMethods = methodStatusses.get(relativeUri);
+                                var failedJavaMethod = uriMethods.findAtPoint((int) dafnyDiagnostic.getLineNumber());
```
fabiomadge, line 285:
```suggestion
                                var uriMethods = methodStatusses.get(relativeUri);
                                if (uriMethods == null) return;
                                var failedJavaMethod = uriMethods.findAtPoint((int) dafnyDiagnostic.getLineNumber());
```
We should't crash here.

verifier/src/main/java/com/aws/jverify/verifier/dafny/DafnyDriver.java
```diff
@@ -0,0 +1,339 @@
...
+import java.util.List;
+import java.util.Locale;
+import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class DafnyDriver implements Driver {
+
+    public static Context context;
```
fabiomadge: Shouldn't this be an instance field?

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -222,11 +85,24 @@ private static String formatDiagnostic(boolean filePath, Diagnostic<?> diagnosti
...
         }
 
-        sb.append(formatMessage(diagnostic));
+        sb.append(Driver.formatMessage(diagnostic));
         return sb.toString();
     }
 
-    public static String formatMessage(Diagnostic<?> diagnostic) {
+
+    JVerifyResults verifyJavaFiles(
```
fabiomadge, line 92: Shouldn't this be backend agnostic? The same applies to the tests. Different backends may not count assertions the same way.

fabiomadge (changes requested):
You moved from verified assertions to verified methods in the test expectations, but didn't do the same for failed assertions. Do you think this will stand the test of time?

verifier/src/main/java/com/aws/jverify/verifier/dafny/DafnyDriver.java
keyboardDrummer: Why would `uriMethods` be null? I'd rather resolve the underlying issue than fail silently

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
keyboardDrummer: > Shouldn't this be backend agnostic?

What part should be back-end agnostic? If you look at the implementation of this method in DafnyDriver, then you can find Dafny specific parts.

> The same applies to the tests. 

In what way?

> Different backends may not count assertions the same way.

Each back-end should show the same # of errors right?

verifier/src/main/java/com/aws/jverify/verifier/dafny/DafnyDriver.java
fabiomadge: Works for me

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
fabiomadge: My point is about the locations of the structs in the repo and the number of failed assertions. Using them in the tests suggests that they will be uniform with regards to the backend, which is not something I would want to commit myself to.

keyboardDrummer: I've renamed all occurrences of failedAssertions to errorCount, to make it more clear what it is and that it should be unrelated to the choice of back-end.

fabiomadge: Yes, but the tests still expect concrete error counts, irrespective of the backend.

keyboardDrummer: Yes, that seems good to me. The error count should be the same for all backends, should it not?

fabiomadge: I'm not sure it should be. Explicit asserts should probably match, maybe unless there's splitting, but what about implicit ones? That is probably more of an encoding question and not necessarily uniform.

keyboardDrummer: Can you give an example? Easiest would be to give one from the test-suite. Is there a particular file for which you think the error count could vary per back-end?

fabiomadge: Take integer overflows. In Dafny we prove that they can't happen. A different backend could model the overflow.

keyboardDrummer: Could you be more concrete? What is the error in the current tests that you think is X with one back-end but could be Y with another back-end?

fabiomadge (dismissed):

fabiomadge (dismissed):

fabiomadge (approved):

PR #372, opened 2026-01-07
Let the AssertFalse test use both the Dafny and Laurel backend

keyboardDrummer:
### What was changed?
- Let the AssertFalse test use both the Dafny and Laurel backend

### How has this been tested?
- The existing AssertFalse.java test has been updated to use both back-ends
- All tests with failing assertions now expect "assertion does not hold" since the Laurel back-end support proving that assertions fail. When using the Dafny back-end, this expectation is translated to "assertion could not be proved" since it can not prove the existence of bugs


.github/workflows/pr-build.yml
```diff
@@ -45,24 +45,46 @@ jobs:
...
+            unzip z3.zip
+          elif [ "$ARCH" = "x86_64" ]; then
+            ARCH_NAME="x86_64"
+            wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-x64-glibc-2.39.zip
+            ARCHIVE_NAME="z3-4.15.2-x64-glibc-2.39"
+            unzip "${ARCHIVE_NAME}.zip"
+          elif [ "$ARCH" = "aarch64" ] || [ "$ARCH" = "arm64" ]; then
+            ARCH_NAME="arm64"
+            curl -L -o z3.zip https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-arm64-glibc-2.34.zip
+            ARCHIVE_NAME="z3-4.15.2-arm64-win"
```
fabiomadge: Wrong archive name - downloads `arm64-glibc-2.34.zip` but references `arm64-win`
```suggestion
            ARCHIVE_NAME="z3-4.15.2-arm64-glibc-2.34"
```

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
```diff
@@ -0,0 +1,346 @@
...
+
+    public JVerifyResults verifyJavaFiles(
+            List<JavaFileObject> readFiles
+    ) {
+        List<Diagnostic<?>> diagnostics = new ArrayList<>();
+
+        var messages = JavacMessages.instance(context);
+        messages.add("com.aws.jverify.messages");
+
+        var analysisResult = verifierOptions.time("Compiling Java to Dafny",
```
fabiomadge:
```suggestion
        var analysisResult = verifierOptions.time("Compiling Java to Laurel",
```

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
```diff
@@ -0,0 +1,346 @@
...
+        private static final int SEVERITY_ERROR = 1;
+        private static final int SEVERITY_WARNING = 2;
+        private static final int SEVERITY_INFO = 4;
+
+        private final URI uri;
+        private final Range range;
+        private final String message;
+        private final int severity;
+
+        public DafnyDiagnostic.Location location;
```
fabiomadge: Unused field
```suggestion
```

verifier/src/main/java/com/aws/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -0,0 +1,159 @@
...
+
+        private StmtExpr convertStatement(JCTree.JCStatement statement) {
+            return switch (statement) {
+                case JCTree.JCAssert assertStmt ->
+                    new Assert(toSourceRange(assertStmt), convertExpression(assertStmt.cond));
+                case JCTree.JCExpressionStatement exprStmt ->
+                    convertExpression(exprStmt.expr);
+                case JCTree.JCBlock block ->
+                    convertMethodBody(block);
+                default -> null; //throw new RuntimeException("Not supported " + statement.getClass().getName());
```
fabiomadge:
```suggestion
                default -> null; // Unsupported statements skipped
```

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
```diff
@@ -0,0 +1,346 @@
...
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class LaurelDriver implements Driver {
+
+    public final Context context;
```
fabiomadge:
```suggestion
    private final Context context;
```

verifier/src/main/java/com/aws/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -0,0 +1,159 @@
...
+            var name = jverifyMethod.getQualifiedName().toString();
+            if (name.equals("check")) {
+                if (invocation.args.size() != 1) {
+                    throw new JavaViolationException("Check should have a single argument");
+                }
+                return new Assert(toSourceRange(invocation),
+                        convertExpression(invocation.args.getFirst()));
+            } if (name.equals("assume")) {
+                if (invocation.args.size() != 1) {
+                    throw new JavaViolationException("Check should have a single argument");
```
fabiomadge:
```suggestion
                    throw new JavaViolationException("Assume should have a single argument");
```

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
```diff
@@ -0,0 +1,346 @@
...
+                // in order to preserve the order of output and to avoid potential deadlock.
+                var process = processBuilder.redirectErrorStream(true).start();
+                try (var strataStdin = process.getOutputStream();
+                     var writer = IonBinaryWriterBuilder.standard().build(strataStdin)) {
+                    serializedProgram.writeTo(writer);
+                }
+                return parseStrataOutput(filesMap, verifierOptions, nameCompiler, process);
+            } catch (InterruptedException | IOException e) {
+                verifierOptions.outWriter().println("Failed to use Strata at: " + verifierOptions.backendPath());
+                e.printStackTrace();
```
fabiomadge:
```suggestion
                verifierOptions.outWriter().println("Failed to use Strata: " + e.getMessage());
```

fabiomadge (commented):
The major issues are the hardcoded path in `Main.java` and the bad `ARCHIVE_NAME`.

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
```diff
@@ -0,0 +1,346 @@
...
+        if (!checkedVersion) {
+            checkedVersion = true;
+        }
+    }
+
+    public JVerifyResults runVerifier(FilesMap filesMap, NameCompiler nameCompiler, IonValue serializedProgram) {
+        checkVerifierVersion();
+
+        var processBuilder = new ProcessBuilder(
+                "lake", "exe",  "-q", "strata", "laurelAnalyze"
```
fabiomadge:
```suggestion
                "lake", "exe", "-q", "strata", "laurelAnalyze"
```

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
```diff
@@ -0,0 +1,346 @@
...
+
+    private static boolean checkedVersion = false;
+
+    private void checkVerifierVersion() {
+        if (!verifierOptions.testBackendVersion()) {
+            return;
+        }
+
+        if (!checkedVersion) {
+            checkedVersion = true;
```
fabiomadge:
```suggestion
            // TODO: implement Strata version check
            checkedVersion = true;
```

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
```diff
@@ -0,0 +1,346 @@
...
+                return new JVerifyResults(List.of(), -1, null);
+            }
+        });
+    }
+
+    /**
+     * Parses the given {@code dafny verify} output,
+     * adding both diagnostics and the summary verified/error counts to {@code outResults}.
+     * Note that Dafny must be invoked with {@code --json-diagnostics} or else parsing will fail.
+     */
```
fabiomadge: Javadoc references Dafny instead of Strata
```suggestion
    /**
     * Parses Strata output, extracting diagnostics and verification counts.
     */
```

verifier/src/main/java/com/aws/jverify/verifier/Main.java
```diff
@@ -120,22 +120,32 @@ public Integer call() throws IOException {
...
+                var isWindows = System.getProperty("os.name", "").toLowerCase().contains("windows");
 
-            dafnyPath = Path.of(isWindows ? "Dafny.exe" : "dafny");
+                dafnyPath = Path.of(isWindows ? "Dafny.exe" : "dafny");
+            }
+            return dafnyPath;
+        }
+        
+        if (backend == Backend.Laurel) {
+            return Path.of("/Users/rwillems/SourceCode/GradleBased/Strata");
```
fabiomadge: Hardcoded path - breaks for other developers
```java
if (backend == Backend.Laurel) {
    var strataPath = System.getenv("JVERIFY_STRATA");
    return strataPath != null ? Path.of(strataPath) : Path.of("Strata");
}
```

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
```diff
@@ -0,0 +1,346 @@
...
+                        Files.writeString(verifierOptions.printSerializedOutputProgram(), files.toPrettyString());
+                    } catch (IOException e) {
+                        throw new RuntimeException(e);
+                    }
+                }
+                return files;
+            });
+
+            var results = runVerifier(analysisResult.filesMap(), NameCompiler.instance(context), serializedProgram);
+            results.diagnostics().addAll(0, diagnostics);
```
fabiomadge: Mutating record's list after construction. Same issue in DafnyDriver.java:86. Replace with:
```java
var allDiagnostics = new ArrayList<>(diagnostics);
allDiagnostics.addAll(results.diagnostics());
return new JVerifyResults(allDiagnostics, results.exitCode(), results.verificationResults());
```

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
```diff
@@ -0,0 +1,346 @@
...
+                                             NameCompiler nameCompiler,
+                                             Process process) throws IOException, InterruptedException {
+
+        List<Diagnostic<?>> diagnostics = new ArrayList<>();
+        try (var strataOutput = process.inputReader()) {
+
+            Wrapper<Integer> performanceTicks = new Wrapper<>(null);
+            Wrapper<Integer> failedAssertionsCount = new Wrapper<>(0);
+            var annotationCompiler = context.get(VerifyAnnotationCompiler.class);
+            var methodStatusses = annotationCompiler.getMethodStatusPerUri();
```
fabiomadge: Typo: `methodStatusses` → `methodStatuses` (also lines 209, 220)

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    /**
+     * Parses the given {@code dafny verify} output,
+     * adding both diagnostics and the summary verified/error counts to {@code outResults}.
+     * Note that Dafny must be invoked with {@code --json-diagnostics} or else parsing will fail.
+     */
+    private JVerifyResults parseStrataOutput(FilesMap filesMap,
+                                             VerifierOptions options,
+                                             NameCompiler nameCompiler,
```
fabiomadge, line 151: `nameCompiler` parameter unused - remove from signature and call site (line 141)

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
```diff
@@ -0,0 +1,346 @@
...
+                                             Process process) throws IOException, InterruptedException {
+
+        List<Diagnostic<?>> diagnostics = new ArrayList<>();
+        try (var strataOutput = process.inputReader()) {
+
+            Wrapper<Integer> performanceTicks = new Wrapper<>(null);
```
fabiomadge, line 157: `performanceTicks` always null - remove Wrapper, pass `null` directly to `VerificationResults` (line 237)

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
```diff
@@ -0,0 +1,346 @@
...
+import com.aws.jverify.common.Range;
+import com.aws.jverify.laurel.IonSerializer;
+import com.aws.jverify.verifier.*;
+import com.aws.jverify.verifier.compiler.Reporter;
+import com.aws.jverify.verifier.compiler.generator.laurel.JavaToLaurelCompiler;
+import com.aws.jverify.verifier.compiler.generator.laurel.LaurelFile;
+import com.aws.jverify.verifier.compiler.simplifications.NameCompiler;
+import com.aws.jverify.verifier.compiler.simplifications.VerifyAnnotationCompiler;
+import com.aws.jverify.verifier.dafny.*;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
```
fabiomadge: Unused imports: `JsonIgnoreProperties`, `ObjectMapper`, `ByteArrayOutputStream`, `StringWriter` (lines 16, 17, 24, 26)

verifier/src/main/java/com/aws/jverify/verifier/compiler/generator/laurel/LaurelFile.java
```diff
@@ -0,0 +1,8 @@
+package com.aws.jverify.verifier.compiler.generator.laurel;
+
+import com.aws.jverify.laurel.Node;
+
+import java.net.URI;
+import java.util.List;
```
fabiomadge: Unused import: `List`

verifier/src/main/java/com/aws/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -0,0 +1,159 @@
...
+import com.sun.tools.javac.code.Flags;
+import com.sun.tools.javac.code.Symbol;
+import com.sun.tools.javac.code.TypeTag;
+import com.sun.tools.javac.tree.JCTree;
+import com.sun.tools.javac.tree.TreeInfo;
+import com.sun.tools.javac.tree.TreeScanner;
+import com.sun.tools.javac.util.Context;
+
+import javax.tools.JavaFileObject;
+import java.io.IOException;
```
fabiomadge: Unused import: `IOException`

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
```diff
@@ -0,0 +1,346 @@
...
+
+        var processBuilder = new ProcessBuilder(
+                "lake", "exe",  "-q", "strata", "laurelAnalyze"
+        );
+        processBuilder.directory(verifierOptions.backendPath().toFile());
+        return verifierOptions.time("Running Strata", () -> {
+            try {
+                // Redirect stderr into stdout, instead of reading one and then the other,
+                // in order to preserve the order of output and to avoid potential deadlock.
+                var process = processBuilder.redirectErrorStream(true).start();
```
fabiomadge: Process not destroyed on exception - add `process.destroyForcibly()` in catch block

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
```diff
@@ -0,0 +1,346 @@
...
+            return range.start().character();
+        }
+
+        @Override
+        public String getCode() {
+            return "strata.verification.error";
+        }
+
+        @Override
+        public String getMessage(Locale locale) {
```
fabiomadge, line 322: `locale` parameter unused

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -171,57 +174,69 @@ public static void verifyFile(SourceFile sourceFile, JVerifyTest annotation, Lis
...
+                    results.exitCode(),
+                    is(annotation.exitCode())),
+            () -> {
+                if (expectedJavaVerifiedCount != null) {
+                    assertThat("Java verified methods count",
+                            results.verificationResults().verificationPassedMethods(),
+                        is(expectedJavaVerifiedCount));
+                }
+            },
+                () -> {
```
fabiomadge: nit: inconsistent indentation in `assertAll` block

.github/workflows/pr-build.yml
keyboardDrummer: I'll remove this section, given that it seems unused.

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
keyboardDrummer: I'll leave it in since we'll need it soon enough

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
keyboardDrummer: It's part of the base method.

verifier/src/main/java/com/aws/jverify/verifier/Main.java
fabiomadge: The hardcoded path is fixed, but there's a new bug at lines 127-128:

```java
if (System.getenv("JVERIFY_" + backend.toString()) != null) {
    result = Path.of(System.getenv("JVERIFY_DAFNY"));  // ← always reads DAFNY
}
```

Should be:
```java
var envVar = "JVERIFY_" + backend.toString();
if (System.getenv(envVar) != null) {
    result = Path.of(System.getenv(envVar));
}
```

fabiomadge (commented):
Follow-up comments

verifier/src/main/java/com/aws/jverify/verifier/Driver.java
```diff
@@ -94,10 +103,18 @@ JVerifyResults verifyJavaFiles(
...
+        var context = new Context();
+        TypesWithoutErasure.preRegister(context);
+        context.put(VerifierOptions.class, options);
+        var lowerer = new JavaLowerer(context);
+        context.put(JavaLowerer.class, lowerer);
+        return switch (backend) {
+            case Dafny -> new DafnyDriver(context);
+            case Strata -> new LaurelDriver(context);
+            default -> throw new RuntimeException("Unsupported backend: " + backend);
+        };
```
fabiomadge, line 116: Removing default lets the compiler catch missing cases when new backends are added.
```suggestion
            case Strata -> new LaurelDriver(context);
        };
```

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
```diff
@@ -0,0 +1,337 @@
...
+                                failedJavaMethod.setVerificationStatus(JavaMethodVerificationStatus.VerificationStatus.Failed);
+                            }
+                        }
+                    }
+                }
+            }
+
+            for (IntervalTree<Integer, JavaMethodVerificationStatus> uriStatusses : methodStatuses.values()) {
+                var statusses = uriStatusses.streamNodes().toList();
+                for (var methodStatus : statusses) {
```
fabiomadge: Typo: statusses → statuses. Same in DafnyDriver:311-312.
```suggestion
            for (IntervalTree<Integer, JavaMethodVerificationStatus> uriStatuses : methodStatuses.values()) {
                var statuses = uriStatuses.streamNodes().toList();
                for (var methodStatus : statuses) {
```

docs/guide/src/first_verification.md
```diff
@@ -27,7 +27,7 @@ Dafny program verifier finished with 3 verified, 0 errors
 
 This indicates that JVerify successfully verified 3 batches of specifications.
 
-You can remove the need for the `--dafny` option by configuring the environment variable `JVERIFY_DAFNY` instead, or modifying your `PATH` so that `dafny` resolves automatically.
+You can remove the need for the `--verifier` option by configuring the environment variable `JVERIFY_DAFNY` instead, or modifying your `PATH` so that `dafny` resolves automatically.
```
fabiomadge:
```suggestion
You can remove the need for the `--verifier` option by configuring the environment variable `JVERIFY_DAFNY` (for Dafny) or `JVERIFY_STRATA` (for Strata), or modifying your `PATH` so that `dafny` resolves automatically.
```

verifier/src/main/java/com/aws/jverify/verifier/Main.java
```diff
@@ -120,22 +120,24 @@ public Integer call() throws IOException {
...
-        if (dafnyPath == null || !Files.exists(dafnyPath)) {
+        if (result == null || !Files.exists(result)) {
             if (verbose) {
-                output.println("Could not find a file at Dafny path '" + dafnyPath + "'");
+                output.println("Could not find a file at verifier path '" + result + "'");
+            }
+            if (backend == Backend.Dafny) {
+                var isWindows = System.getProperty("os.name", "").toLowerCase().contains("windows");
+                result = Path.of(isWindows ? "Dafny.exe" : "dafny");
             }
```
fabiomadge, line 138: No fallback for Strata backend. Consider adding:
```java
} else if (backend == Backend.Strata) {
    result = Path.of("Strata");
}
```

fabiomadge:
LaurelDriver now has AutoClosingProcessWrapper for process cleanup on exceptions. DafnyDriver (line 132) doesn't have equivalent protection - consider extracting to a shared utility.

fabiomadge:
Consider adding CI to ensure Laurel AST classes stay in sync with Strata grammar:
```yaml
- name: Verify Laurel AST is up to date
  run: |
    make generate-laurel-ast
    git diff --exit-code verifier/src/main/java/com/aws/jverify/laurel/
```

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
fabiomadge: The ArrayList change makes mutation explicit but doesn't fix the root cause. The record is still being mutated after construction (LaurelDriver:102, DafnyDriver:83).

Cleaner approach - construct new list when combining:
```java
var allDiagnostics = new ArrayList<>(diagnostics);
allDiagnostics.addAll(results.diagnostics());
return new JVerifyResults(allDiagnostics, results.exitCode(), results.verificationResults());
```
This also allows keeping the record field as `List<>` instead of exposing ArrayList in the public API.

verifier/src/main/java/com/aws/jverify/verifier/Main.java
keyboardDrummer: How would that work?

keyboardDrummer:
> Consider adding CI to ensure Laurel AST classes stay in sync with Strata grammar:
> 
> ```yaml
> - name: Verify Laurel AST is up to date
>   run: |
>     make generate-laurel-ast
>     git diff --exit-code verifier/src/main/java/com/aws/jverify/laurel/
> ```

I think it's OK if it's out of date, although it would most likely cause the tests to fail.

fabiomadge:
> I think it's OK if it's out of date, although it would most likely cause the tests to fail.

Do you expect a scenario where the mismatch is actually desirable? It seems like this would still make debugging easier.

keyboardDrummer:
> > I think it's OK if it's out of date, although it would most likely cause the tests to fail.
> 
> Do you expect a scenario where the mismatch is actually desirable? It seems like this would still make debugging easier.

If there is a mismatch, then the tests will fail with Strata complaining that you're sending the wrong data. You would likely see this error locally, assuming you're locally running tests, and resolve the issue before running CI, so I'm not expecting much debugging help through it.

That said, I will add your suggestion. It might help prevent confusion indeed.

keyboardDrummer:
> LaurelDriver now has AutoClosingProcessWrapper for process cleanup on exceptions. DafnyDriver (line 132) doesn't have equivalent protection - consider extracting to a shared utility.

I would prefer to leave the Dafny code as is, and this change was for an edge case anyways.

verifier/src/main/java/com/aws/jverify/verifier/Main.java
fabiomadge: You're right - Strata uses `lake exe`, not a standalone executable. Withdrawn.

verifier/src/main/java/com/aws/jverify/verifier/laurel/LaurelDriver.java
keyboardDrummer: Yeah, well it's set up so that if there was a message that increased the ticks, it could be increased. The wrapper is needed because the messages are handled in a lambda.

keyboardDrummer: I think we could remove all the performanceTicks references, both for Dafny and Laurel, since it's unused now, but I kept it since we previously used it for Dafny and I imagine we'll use something like it in the future.

.github/workflows/pr-build.yml
```diff
@@ -45,24 +45,47 @@ jobs:
...
+            ./jverify-collection/Strata
+          build: "true"
+          test: "false"
+          lint: "false"
+      - name: Verify Laurel AST is up to date
         run: |
-          dotnet build ./Source/Dafny.sln
-      - name: Download Z3
+          make generate-laurel-ast
+          git diff --exit-code verifier/src/main/java/com/aws/jverify/laurel/
```
fabiomadge: Windows CI fails due to line ending mismatch (autocrlf converts to CRLF on checkout, generator outputs LF).
```suggestion
          git diff --exit-code --ignore-cr-at-eol verifier/src/main/java/com/aws/jverify/laurel/
```

fabiomadge:
The Laurel AST check passes now, but `make test-generation-nodiff` still fails on Windows for the same reason - it uses `git status` which doesn't have an ignore-line-endings option.

Fix in Makefile:
```makefile
test-generation-nodiff: generate-dafny-ast
	git diff --exit-code --ignore-cr-at-eol \
		|| { echo 'Consider running `make generate-dafny-ast`'; exit 1; }
```

fabiomadge (approved):

PR #380, opened 2026-03-24
Fix name collisions in Strata backend

fabiomadge:
Use class-qualified names (`ClassName_methodName`) in the Laurel backend to avoid collisions with Strata built-ins (`increment`, `select`, `update`) and between classes with same method names.

## Tests

- `StrataNameCollision` — method named `increment` no longer collides with Strata built-in; cross-class collision with `compute` in `StrataNameCollisionHelper`


keyboardDrummer (approved):
Looks good, thanks! For instance methods, once we have them in Laurel, this should be unnecessary, but it's a small idiomaticity thing.

Issue #384, opened 2026-04-10
Investigate how much work it is to use the Laurel/Strata back-end to get all the Dafny based tests to pass


keyboardDrummer:
Investigate how much work it is to use the Laurel/Strata back-end to get all the Dafny based tests to pass

fabiomadge:
Per-test triage results are now available: https://github.com/strata-org/jverify/blob/chore/strata-test-triage/TRIAGE_RESULTS.md

117 tests total, 68 failed, 49 passed. Key findings:

- **30 tests** fail because the Laurel compiler only processes static methods (instance methods silently skipped). `VerifyBooleanOperators` would pass today if its method were `static`.
- **25 tests** crash on unsupported types in static method signatures/bodies (reference types, generics, arrays, float/double).
- **5 tests** fail due to `verifyPrintedDafny = true` trying to run the Strata directory as a Dafny executable — will be removed with Dafny deletion.
- **2 tests** partially succeed but produce wrong error messages/locations (Strata error reporting issue).
- **1 test** crashes because `TreeScanner` recurses into nested classes with unsupported types.
- **5 tests** need unsupported operators (bitwise, shift, ++, +=).

This aligns with the tier breakdown in this issue. Instance method support is the #1 priority — it unblocks 1 test immediately and enables progress on 29 others.

olivier-aws:
@keyboardDrummer-bot Can you create github issues for each of the tasks referenced in https://github.com/strata-org/jverify/blob/chore/strata-test-triage/TRIAGE_RESULTS.md?

PR #387, opened 2026-04-16
chore: Delete Dafny backend, adapt codebase to Strata-only

fabiomadge:
## Summary

Delete the entire Dafny backend and adapt the codebase to be Strata-only.

### Deleted
- `dafny` submodule
- `javaTypesGenerator/` project (Dafny AST code generator)
- 183 generated Dafny AST classes (`verifier/src/main/java/com/aws/jverify/generated/`)
- Dafny backend code: `compiler/generator/dafny/` (20 files, ~3,900 lines) and `verifier/dafny/` (10 files)
- `MethodOrLoopDafnyContract.java`
- `Backend` enum
- `TestVerifier.java` (Dafny-specific integration tests)
- Dafny resources (`additional.dfy`, `dafny.properties`)
- Dafny CLI options (`--print-dafny`, `--print-binary-dafny`, `--backend`)
- `NameCompiler` class (dead code — only used by Dafny compiler)
- `SourceOrigin` class (never instantiated)
- `performanceTicks` field (Dafny resource count, never set by Strata)
- `verifyPrintedDafny` annotation field (Dafny round-trip check)
- Unused dependencies: jackson-databind, jgrapht-core
- 6 stale `requires` and 4 stale `opens` directives from `module-info.java`

### Kept
- `builtin-contracts/` project — backend-agnostic JVerify specifications for JCL classes using the `@Contract` annotation API
- `useBuiltinContracts` annotation field and wiring — controls whether builtin contracts are passed to the verifier

### Moved/Refactored
- `DiagnosticWithRange` → `verifier` package (was in `verifier.dafny`)
- Position tracking classes (`Token`, `TokenRange`, `IOrigin`, `TokenRangeOrigin`) → new `verifier.compiler.position` package (were in `generated`)
- Inlined `ImpureObjectGenerator.IMPURE_OBJECT_NAME` constant

### Simplified
- `Driver.getDriver()` no longer takes a `Backend` parameter
- `Main` no longer has `--backend` option (always Strata)
- `VerifierOptions` removed 5 Dafny-specific fields
- Test engine: removed `Backend` parameter, `verifyPrintedDafny`, `getDafnyInSubmodulePath`
- All test `@JVerifyTest` annotations: removed `BACKENDS`, `verifyPrintedDafny`, `performanceTicks`
- `LaurelDriver`: removed unused `nameCompiler` parameter threading
- `Reporter`: removed dead `getReportingRange` method, simplified `getEntireRange`/`getPositionRange`
- `IOrigin`: converted from abstract class to sealed interface (single permitted class: `TokenRangeOrigin`)
- `Token`/`TokenRange`: removed redundant getter methods (use record accessors)
- CI workflow: removed Dafny build, path setup, getting-started commands, stale `chmod` matrix variable
- Renamed 5 `*Dafny*` test files to drop backend-specific naming
- Fixed `--filter-position` description (`.dfy` typo → `.java`, restored full example text)
- Fixed stale comments referencing `additional.dfy` and Dafny concepts
- Fixed broken `builtin-contracts` reference in `sample-project`
- Renamed `temp.dbin` → `temp.laurel`
- Restored stripped javadoc on `JVerifyTest` and `JVerifyTestEngine`
- Added TODO in `JavaToLaurelCompiler` for overload name disambiguation
- Extracted `NAME_SEPARATOR` constant in `MoveStaticMethodsToStaticType`

Relates to #384


verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/records/RecordsVerified.java
```diff
@@ -154,10 +154,10 @@ record Factor(int value) implements ICoefficient {
 }
 
 /**
- * Records are translated to Dafny datatypes,
- * which don't automatically yield both nullable and non-nullable types (as Dafny classes do).
+ * Records are translated to value types,
+ * which don't automatically yield both nullable and non-nullable types (as classes do).
  * This wrapper class works around JVerify's current lack of automatic handling
- * for values that are of reference types in Java but of value types in Dafny.
+ * for values that are of reference types in Java but of value types in the backend.
```
robin-aws: Would it be accurate to just say "In Strata"?

(There's dozens of comments where that might be true and I'm not going to make you change all of them mind you :)

verifier/src/test/java/com/aws/jverify/verifier/tests/jcl/BigIntegers.java
```diff
@@ -15,7 +15,7 @@
         "OnlyOneElementUsed",
         "StringOperationCanBeSimplified"
 })
-@JVerifyTest(exitCode = 4, useBuiltinContracts = true, methodsVerified = 4, errorCount = 2)
```
robin-aws, line 18: Why remove `useBuildinContracts`? That's independent of backend right?

verifier/src/test/java/com/aws/jverify/verifier/tests/verification/MethodContractsVerification.java
```diff
@@ -10,7 +10,7 @@
 import static com.aws.jverify.JVerify.postcondition;
 import static com.aws.jverify.JVerify.precondition;
 
-@JVerifyTest(methodsVerified = 10, errorCount = 0, verifyPrintedDafny = true)
```
robin-aws, line 13: Doesn't have to happen in this PR, but it might make sense to have an equivalent feature to serialize and check the Strata in a similar way?

build.gradle.kts
```diff
@@ -293,8 +270,6 @@ project(":verifier") {
         implementation(project(":common"))
         implementation(project(":library"))
 
-        implementation("org.jgrapht:jgrapht-core:1.5.2")
```
robin-aws, line 296: Curious why we were able to drop these dependencies, was it related to removing Dafny?

verifier/src/main/java/module-info.java
```diff
@@ -2,25 +2,13 @@
...
-    requires com.fasterxml.jackson.databind;
-    requires org.jgrapht.core;
-    requires java.naming;
     requires net.bytebuddy;
     requires java.instrument;
     requires net.bytebuddy.agent;
     requires com.amazon.ion;
 
     opens com.aws.jverify.verifier to info.picocli;
-    opens com.aws.jverify.verifier.compiler.simplifications to info.picocli;
```
robin-aws, line 20: I might be missing why dropping Dafny lets us drop all these dependencies

verifier/src/main/java/com/aws/jverify/verifier/Main.java
```diff
@@ -34,31 +33,25 @@ class AppCommand implements Callable<Integer> {
     @Parameters(description = "Java files to verify", arity = "1..*")
     private List<Path> inputs;
     
-    @Option(names = "--print-binary-dafny", description = "Given a filepath, prints the binary Dafny code that is generated from Java")
```
robin-aws, line 37: Worth having the strata equivalent too I think (but can be added later)

verifier/src/main/java/com/aws/jverify/verifier/Main.java
```diff
@@ -34,31 +33,25 @@ class AppCommand implements Callable<Integer> {
...
-    private Path printDafny;
-    
     @Option(names = "--show-ranges", description = "Show source location ranges in diagnostics")
     private boolean showRanges;
     
     @Option(names = "--paths", description = "Show file paths instead of names")
     private boolean paths;
     
-    @Option(names = "--filter-position", description = "Filter what gets verified based on a source location. The location is specified as a file path suffix, optionally followed by a colon and a line number or line range. For example, `jverify verify source1.java source2.java --filter-position=source1.dfy:5-7` will only verify things that between (and including) line 5 and 7 in the file `source1.dfy`. You can also use `:5`, `:5-`, `:-5` to specify individual lines or open ranges.")
+    @Option(names = "--filter-position", description = "Filter what gets verified based on a source location. The location is specified as a file path suffix, optionally followed by a colon and a line number or line range.")
```
robin-aws: Looks like `.dfy` in the original was just a typo, we should fix that and keep the more detailed message.

verifier/src/main/java/com/aws/jverify/verifier/Main.java
```diff
@@ -34,31 +33,25 @@ class AppCommand implements Callable<Integer> {
...
     private String filterPositionString;
     
     @Option(names = "--include-filter-dependencies", description = "When filtering on just a file, also verify its transitive dependencies.")
     private boolean includeFilterDependencies;
     
-    @Option(names = "--verifier", description = "Location of the verifier to use. Overrides environment variable JVERIFY_DAFNY or JVERIFY_STRATA, depending on the back-end used.")
+    @Option(names = "--verifier", description = "Location of the Strata verifier to use. Overrides environment variable JVERIFY_STRATA.")
```
robin-aws:
```suggestion
    @Option(names = "--verifier", description = "Location of the Strata CLI to use. Overrides environment variable JVERIFY_STRATA.")
```

verifier/src/main/java/com/aws/jverify/verifier/Main.java
```diff
@@ -69,24 +62,14 @@ class AppCommand implements Callable<Integer> {
...
-    @Option(names = "--backend", description = "")
-    private Backend backend = Backend.Dafny;
 
     @Override
     public Integer call() throws IOException {
         var writer = new PrintWriter(spec.commandLine().getOut());
 
-        // In the future we'll have to add an argument to specify jar files for dependencies of the input sources,
-        // And those will include the JVerify library jar.
-        // But right now we manually find the JVerify library jar and include it in the compilation.
```
robin-aws, line 82: This is still true I think?

verifier/src/main/java/com/aws/jverify/verifier/Main.java
```diff
@@ -99,21 +82,16 @@ public Integer call() throws IOException {
...
-        var testDafnyVersion = customVerifier != null;
         var workingDirectory = Path.of(System.getProperty("user.dir"));
         var positionFilter = PositionFilter.getPositionFilter(includeFilterDependencies, filterPositionString);
-        String[] additionalDafnyArguments = { 
-                //"--wait-for-debugger"
-        };
-        var verifierOptions = new VerifierOptions(writer, workingDirectory, dafnyPath, classpathEntries,
-                tempFile.toPath(), testDafnyVersion,
-                printDafny, printBinaryDafny, showRanges, contractPathEntries,
-                paths, additionalDafnyArguments, verifyByDefault, false,
```
robin-aws, line 111: Is `additionalDafnyArguments` still defined above? I assume we want to remove it entirely (even if we might want the Strata equivalent later)

verifier/src/main/java/com/aws/jverify/verifier/Main.java
```diff
@@ -34,31 +33,25 @@ class AppCommand implements Callable<Integer> {
...
+    @Option(names = "--filter-position", description = "Filter what gets verified based on a source location. The location is specified as a file path suffix, optionally followed by a colon and a line number or line range.")
     private String filterPositionString;
     
     @Option(names = "--include-filter-dependencies", description = "When filtering on just a file, also verify its transitive dependencies.")
     private boolean includeFilterDependencies;
     
-    @Option(names = "--verifier", description = "Location of the verifier to use. Overrides environment variable JVERIFY_DAFNY or JVERIFY_STRATA, depending on the back-end used.")
+    @Option(names = "--verifier", description = "Location of the Strata verifier to use. Overrides environment variable JVERIFY_STRATA.")
     private Path customVerifier;
```
robin-aws, line 55: Could we just rename this to `--strata` instead?

verifier/src/main/java/com/aws/jverify/verifier/Main.java
```diff
@@ -123,21 +101,15 @@ public Integer call() throws IOException {
...
             }
         }
         if (result == null || !Files.exists(result)) {
             if (verbose) {
-                output.println("Could not find a file at verifier path '" + result + "'");
-            }
-            if (backend == Backend.Dafny) {
-                var isWindows = System.getProperty("os.name", "").toLowerCase().contains("windows");
-                result = Path.of(isWindows ? "Dafny.exe" : "dafny");
+                output.println("Could not find Strata at verifier path '" + result + "'");
```
robin-aws:
```suggestion
                output.println("Could not find Strata at path '" + result + "'");
```

.github/workflows/pr-build.yml
```diff
@@ -16,18 +16,10 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-24.04, windows-2022 ]
-        include:
-          - os: ubuntu-24.04
-            chmod: true
-          - os: windows-2022
-            chmod: false
```
robin-aws, line 23: Just to check, this was unrelated dead configuration right?

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTest.java
```diff
@@ -37,42 +36,26 @@
      * Mark the test as skipped for the given reason (if not null or empty).
      */
     String skip() default "";
-
     /**
      * What to pass for the --verify-by-default verifier option.
      */
     boolean verifyByDefault() default true;
-
-    boolean useBuiltinContracts() default false;
```
robin-aws, line 46: Commented elsewhere as well, but I think we want to keep this configuration for whenever we do support the builtin contracts on Strata. Otherwise we'll have to re-do the work to figure out which skipped tests need the option.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -106,16 +105,12 @@ public EngineExecutionContext execute(EngineExecutionContext context, DynamicTes
         {
             LOGGER.fine(() -> "Executing test for class: " + testClass);
 
-            // We assume that the working directory is the root of the current Gradle project (as is default),
-            // and that the test sources are in the standard "src/test/java" subdirectory of the project root.
```
robin-aws, line 110: Still true

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
```diff
@@ -106,16 +105,12 @@ public EngineExecutionContext execute(EngineExecutionContext context, DynamicTes
...
             var sourceRoot = Path.of(System.getProperty("user.dir") , "src", "test", "java");
             if (!Files.exists(sourceRoot) || !Files.isDirectory(sourceRoot)) {
                 Assertions.fail(() ->
                         "Test sources root directory for %s not found at %s".formatted(testClass, sourceRoot));
             }
 
-            // We assume the class's source file is at the path corresponding to its package and class name,
-            // just as Java requires for public classes.
```
robin-aws, line 118: Still true

verifier/src/main/java/com/aws/jverify/verifier/compiler/position/IOrigin.java
robin-aws, line 1: I assume these four files were actually a move from somewhere else?

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/NameCompiler.java
robin-aws, line 1: I think we'll still need something similar for Strata as well? Might be better to strip out the Dafny references but keep the logic.

robin-aws (changes requested):
WHEW :)

Main high-level feedback is I think we should keep the builtin contracts configuration, even though it's not supported for Strata yet. It'll save a bunch of work when we do get to that.

verifier/src/test/java/com/aws/jverify/verifier/tests/javasupport/records/RecordsVerified.java
fabiomadge: Updated to "In Strata, records are translated to value types..."

verifier/src/test/java/com/aws/jverify/verifier/tests/jcl/BigIntegers.java
fabiomadge: You're right, this is backend-agnostic. Restored `useBuiltinContracts` along with the `builtin-contracts` project.

verifier/src/test/java/com/aws/jverify/verifier/tests/verification/MethodContractsVerification.java
fabiomadge: Agreed, a round-trip check for Strata output would be valuable. Not in scope for this PR but noted as a follow-up.

build.gradle.kts
fabiomadge: Yes — jackson was used for parsing Dafny's JSON output (`DafnyOutputDeserializer`, `DafnyDiagnostic`, `StatusMessage`), jgrapht for type hierarchy ordering in `BaseDafnyGenerator`. No usages remain in the Strata/Laurel code.

verifier/src/main/java/module-info.java
fabiomadge: `com.fasterxml.jackson.*` and `org.jgrapht.core` were only used by deleted Dafny code. `java.sql` and `java.naming` were module-path workarounds needed when jackson-databind was present. `jdk.jshell` had no direct usage — likely added to suppress a module resolution warning. The 4 removed `opens` directives were always unnecessary — none of those packages contain picocli-annotated classes. Only `com.aws.jverify.verifier` (which contains `AppCommand`) needs to be opened to picocli, and that one is kept.

verifier/src/main/java/com/aws/jverify/verifier/Main.java
fabiomadge: Agreed. A `--print-laurel` option would be useful for debugging. Can be added in a follow-up.

verifier/src/main/java/com/aws/jverify/verifier/Main.java
fabiomadge: Good catch — the `.dfy` was always a typo. Restored the full description with it fixed to `.java`.

verifier/src/main/java/com/aws/jverify/verifier/Main.java
fabiomadge: Applied.

verifier/src/main/java/com/aws/jverify/verifier/Main.java
fabiomadge: Restored.

verifier/src/main/java/com/aws/jverify/verifier/Main.java
fabiomadge: Confirmed fully removed — no references remain.

verifier/src/main/java/com/aws/jverify/verifier/Main.java
fabiomadge: `--verifier` is more generic and wouldn't need renaming if the backend ever changes. Happy to go either way — what's your preference?

verifier/src/main/java/com/aws/jverify/verifier/Main.java
fabiomadge: Applied.

.github/workflows/pr-build.yml
fabiomadge: Correct — it was defined but never referenced in any step.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTest.java
fabiomadge: Restored, along with all the test annotations that had it and the wiring in `getVerifierOptions`.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
fabiomadge: Restored.

test-engine/src/main/java/com/aws/jverify/testengine/JVerifyTestEngine.java
fabiomadge: Restored.

verifier/src/main/java/com/aws/jverify/verifier/compiler/position/IOrigin.java
fabiomadge: Yes — moved from `com.aws.jverify.generated`. `Token` and `TokenRange` were simplified to plain records, `IOrigin` was converted from abstract class to sealed interface, and `SourceOrigin` was deleted (never instantiated). Git shows them as new files because the content changed significantly.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/NameCompiler.java
fabiomadge: The Laurel compiler currently passes Java names through verbatim — `JavaToLaurelCompiler` uses `method.name.toString()` directly. This works because Strata doesn't support overloaded methods yet. When it does, we'll need disambiguation, but the old `NameCompiler` was tightly coupled to the Dafny AST and its reserved words, so it couldn't be reused as-is. Added a TODO in `JavaToLaurelCompiler` for when overload support is added.

verifier/src/main/java/com/aws/jverify/verifier/Main.java
fabiomadge: Renamed to `--strata` with description "Location of the Strata binary to use."

robin-aws: Thanks. I think it's better this way so a user doesn't have to understand that "verifier" actually just means "strata" right now.

verifier/src/main/java/com/aws/jverify/verifier/compiler/simplifications/NameCompiler.java
robin-aws: Okay fair enough, not worth keeping if it's going to change a lot.

robin-aws (approved):
🚀 🚀 🚀

Issue #414, opened 2026-05-19
Laurel compiler: char primitive type support


fabiomadge:
The Laurel compiler does not handle Java's `char` primitive type. `JavaToLaurelCompiler.translateType` (`:113-120`) covers `INT`, `SHORT`, `BYTE`, `LONG`, `BOOLEAN` but has no `CHAR` arm — `char` falls through to the `default` throw "Unsupported type".

**Affected tests:** `Switches.switchExprChar`/`switchExprCharBad`, `PrimitiveTypes` (line 28: `char c = 'q';`), `Strings.charAt` paths.

**Work required:**
- Register `char` once via `makeConstrainedType("char", 0L, 65535L)` (mirrors the `int8`/`int16`/`int32`/`int64` registration at `JavaToLaurelCompiler:90-93,97-104`).
- Add `case CHAR -> compositeType("char")` to `translateType` (`:113-120`).
- Add `case CHAR -> literalInt(...)` to `convertLiteral` (`:340-345`).
- Reject `@Nat char` with `notSupported` — the `0..65535` constraint already implies non-negative.

**Notes:**
- Strata has no `char`/`char16` prelude type (verified via grep of `Strata/Languages/Laurel/`); inlining as a constrained int is the only path. Dafny used a named `char16` type; this is a forced divergence.
- ~5 lines of plumbing.

**Depends on:** none.

PR #416, opened 2026-05-19
Skip translator-rejected methods gracefully and support char primitive

fabiomadge:
Step 1 of the JVerify completion plan (PLAN.md §3.1).

## Resolves

- #398 — graceful skip of unsupported methods (catch `JavaViolationException`)
- #414 — `char` primitive support

## Summary

- Wrap per-method translation in `JavaToLaurelCompiler.visitMethodDef` in a `try/catch` on `JavaViolationException`. On catch: emit a `translatorError` diagnostic and demote the method's `IntervalTree` entry from `Verified` to `Skipped`.
- The trailing `super.visitMethodDef` stays outside the catch so nested classes are still visited even when the parent method skipped.
- Catch scope is narrow — `JavaViolationException` only. Broader catches would mask translator bugs (NPEs, "Could not find line map for...") and `OOM`/`StackOverflow` from Strata recursion.
- Add `char` as a constrained int (`0..65535`, full UTF-16 range) registered alongside `int8`/`int16`/`int32`/`int64`.

## Why bundle char with #398

5-line type-plumbing tweak — too small for its own step. Mechanically aligned with the safety-net edits (both touch `translateType` / `convertLiteral`), and unblocks Step 2's `switchExprChar`/`switchExprCharBad` and Step 5's char-using methods in `PrimitiveTypes` without leaving them gated on a follow-up.

## Implementation notes

- `VerifyAnnotationCompiler.markSkipped(JCCompilationUnit, JCMethodDecl)` — new helper. Scans the URI's interval-tree by `methodTree == methodDecl` identity rather than recomputing line position, so it doesn't depend on `PositionCalculator` semantics matching `addMethodToIntervalTree` exactly. No-op when the method has no entry (synthetic methods, methods without an implementation body) — those are already invisible to the counting layer.
- Strata has no `char`/`char16` prelude type (verified via grep of `Strata/Languages/Laurel/`); inlining as a constrained int is the only path. Dafny used a named `char16` type — a forced divergence.
- `translatorError` (new `messages.properties` key) instead of `notSupported`. `JavaViolationException`'s javadoc states "Indicates a JVerify compiler bug" — so it's a translator gap surfaced to the user, not a clean refusal of an explicitly-unsupported feature. Future Step 6 throws → `reportError` migrations will use `notSupported` for the genuine "user feature not supported" cases.

## Test plan

- [x] `./gradlew :verifier:test` passes — 113 tests / 0 failed / 72 skipped.
- [x] No new uncaught `JavaViolationException` stack traces in test output.
- [x] No test transitions green → red.
- [x] `TranslatorSkip.java` covers the catch wrapper end-to-end. Asserts three properties simultaneously: (1) sibling translation continues past a throwing method, (2) the translator-level error reaches the driver as a diagnostic, (3) the throwing method's status flips `Verified → Skipped` (the silent-overcount bug §3.1 calls out). Trigger is the multi-init `for` refusal at `convertStatement`, chosen for stability — the plan refuses this form rather than supporting it.
- [x] `CharPrimitive.java` is the regression test for #414. Pairs the happy path (`translateType` + `convertLiteral` round-trip) with `rangeBoundsBad` that fails when `c == 65535` to assert the `0..65535` constraint is actually enforced rather than dropped.

### DIFF
--- a/verifier/src/test/java/org/strata/jverify/verifier/tests/verification/TranslatorSkip.java
+++ b/verifier/src/test/java/org/strata/jverify/verifier/tests/verification/TranslatorSkip.java
@@ -1,0 +1,34 @@
+package org.strata.jverify.verifier.tests.verification;
+
+import org.strata.jverify.testengine.JVerifyTest;
+
+import static org.strata.jverify.JVerify.check;
+
+/**
+ * Regression test for #398. Asserts that a translator-rejected method
+ * produces a diagnostic, flips its status to Skipped, and leaves siblings
+ * verifiable. The trigger is the multi-init for refusal — if that form
+ * gains support, swap the trigger; the test covers the catch wrapper, not
+ * the construct.
+ */
+@JVerifyTest(
+        continueOnErrors = true,
+        exitCode = 0,
+        methodsVerified = 2,
+        methodsSkipped = 1,
+        errorCount = 0
+)
+class TranslatorSkip {
+    static int unsupported(int n) {
+//             ^ error: Multi-init or multi-step for loops are not supported
+        var sum = 0;
+        for (int i = 0, j = n; i < j; i++, j--) {
+            sum = sum + i;
+        }
+        return sum;
+    }
+
+    static void supported(int x) {
+        check(x == x);
+    }
+}
--- a/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
+++ b/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
@@ -4,10 +4,12 @@ import org.strata.jverify.common.Position;
 import org.strata.jverify.laurel.*;
 import org.strata.jverify.verifier.VerifierOptions;
 import org.strata.jverify.verifier.compiler.JavaViolationException;
+import org.strata.jverify.verifier.compiler.Reporter;
 import org.strata.jverify.verifier.compiler.frontend.JavaLowerer;
 import org.strata.jverify.verifier.compiler.simplifications.JVerifyUtils;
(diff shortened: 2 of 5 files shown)


verifier/src/test/java/org/strata/jverify/verifier/tests/verification/TranslatorSkip.java
```diff
@@ -0,0 +1,34 @@
...
+import static org.strata.jverify.JVerify.check;
+
+/**
+ * Regression test for #398. Asserts that a translator-rejected method
+ * produces a diagnostic, flips its status to Skipped, and leaves siblings
+ * verifiable. The trigger is the multi-init for refusal — if that form
+ * gains support, swap the trigger; the test covers the catch wrapper, not
+ * the construct.
+ */
+@JVerifyTest(
```
keyboardDrummer, line 14: Interesting! I have trouble evaluating the UX but it seems great. Would JVerify previously just crash if it ran into something that was not supported?

keyboardDrummer (approved):
Nice!

verifier/src/test/java/org/strata/jverify/verifier/tests/verification/TranslatorSkip.java
fabiomadge: Yes. `JavaViolationException` is a `RuntimeException` and there's no catch between `JavaToLaurelCompiler` and the test engine / CLI. Pre-#416, an unsupported construct in any method aborted translation of the entire compilation unit (the `TreeScanner` recursion exits early), and the exception propagated out as an unhandled stack trace.

Step 1 wraps each method's translation in `visitMethodDef` so the failure is contained to that one method: sibling methods still translate, the user gets a `Reporter` diagnostic with the right source range, and the method's status is updated from `Verified` to `Skipped` (otherwise it'd silently count as `Verified` despite producing no Laurel).

PR #418, opened 2026-05-20
Add break/continue/labeled statements support

### LINKED ISSUE #407: Laurel compiler: break, continue, labeled statements, and decreases clauses
The Laurel IR already has `StmtExpr.LabelledBlock` and `StmtExpr.Exit`, and `MethodOrLoopContractCompiler` (lines 316-322) already parses `decreases(...)` into `MethodOrLoopContract.decreases`. But:

- `JCBreak`, `JCContinue`, `JCLabeledStatement`, `JCDoWhileLoop` all hit the `default` arm at `JavaToLaurelCompiler.java:265` and throw `JavaViolationException`.
- `visitMethodDef` uses only `preconditions()` + `postconditions()`; `extractLoopParts` uses only `loopInvariants()`. **The collected `decreases` list is dropped.**

Fixes all land in `JavaToLaurelCompiler`:
- New `convertStatement` arms for break/continue/labeled/do-while, wrapping loop bodies in `labelledBlock` for break-and-continue targets.
- Emit `decreases` on procedures and loops (may need a small IR tweak to add a `decreases` slot on `Procedure` / `while` / `forLoop` if absent).

Unblocks `VerifyStatements` (19 methods) and `Termination`.

Note: issue #384's claim that break/continue is "blocked on Strata grammar — needs Exit node" is outdated; the Exit node exists.

olivier-aws:
Resolves #407 (break/continue/labeled; do-while and decreases deferred).

## Design

- Label stack for break/continue resolution
- Only wrap loops in labelledBlock when break/continue is present (tree scan)
- When break/continue present, desugar for-loops to while manually matching PythonToLaurel pattern
- Labeled statements pass label to next loop via pendingLabel

**Not handled:** do-while loops and enhanced-for loops (`for (T x : collection)`) are not yet supported.

### Simple for loop (no break/continue) — uses `forLoop` IR node directly

```java
for (int i = 0; i < 5; i = i + 1) {
    invariant(i >= 0 && i < 5);
    sum = sum + 1;
}
```
→
```
forLoop(
  var i: int32 = 0,
  i < 5,
  i = i + 1,
  [invariant(i >= 0 && i < 5)],
  { sum = sum + 1 }
)
```

### For loop with break — desugars to while with labelled blocks

```java
for (int i = 0; i < 10; i = i + 1) {
    invariant(i >= 0 && i <= 5);
    if (i == 5) { break; }
}
```
→
```
labelledBlock(loop_break_0, {
  var i: int32 = 0;
  while (i < 10)
    invariants: [i >= 0 && i <= 5]
  {
    labelledBlock(loop_continue_1, {
      if (i == 5) { exit loop_break_0; }
    });
    i = i + 1;
  }
})
```

### For loop with continue

```java
for (int i = 0; i < 5; i = i + 1) {
    if (i == 2) { continue; }
    x = x + i;
}
```
→
```
labelledBlock(loop_break_0, {
  var i: int32 = 0;
  while (i < 5)
    invariants: [...]
  {
    labelledBlock(loop_continue_1, {
      if (i == 2) { exit loop_continue_1; }
      x = x + i;
    });
    i = i + 1;
  }
})
```

Note: `continue` exits the inner labelled block (skipping the rest of the body) but the step `i = i + 1` still executes because it is outside the continue-labelled block but inside the while body.

### While loop with break

```java
while (x < 10) {
    if (x == 5) { break; }
    x = x + 1;
}
```
→
```
labelledBlock(loop_break_0, {
  while (x < 10) {
    labelledBlock(loop_continue_1, {
      if (x == 5) { exit loop_break_0; }
      x = x + 1;
    });
  }
})
```
## Test plan

`VerifyBreakContinue` test with 17 methods: for-break, while-break, infinite-while-break, nested-break (inner only), for-continue (single skip and range), while-continue (single skip and range), multiple loops with break+continue, break+continue in same loop, labeled break outer, return inside loop, and one intentionally failing assertion. All existing tests pass.

### DIFF
--- a/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
+++ b/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
@@ -4,10 +4,12 @@ import org.strata.jverify.common.Position;
 import org.strata.jverify.laurel.*;
 import org.strata.jverify.verifier.VerifierOptions;
 import org.strata.jverify.verifier.compiler.JavaViolationException;
+import org.strata.jverify.verifier.compiler.Reporter;
 import org.strata.jverify.verifier.compiler.frontend.JavaLowerer;
 import org.strata.jverify.verifier.compiler.simplifications.JVerifyUtils;
 import org.strata.jverify.verifier.compiler.simplifications.MethodOrLoopContract;
 import org.strata.jverify.verifier.compiler.simplifications.MethodOrLoopContractCompiler;
+import org.strata.jverify.verifier.compiler.simplifications.VerifyAnnotationCompiler;
 import org.strata.jverify.verifier.laurel.FilesMap;
 import com.sun.tools.javac.code.Flags;
 import com.sun.tools.javac.code.Symbol;
@@ -27,12 +29,16 @@ public class JavaToLaurelCompiler {
     private final JavaLowerer lowerer;
     private final MethodOrLoopContractCompiler contractCompiler;
     private final JVerifyUtils jverifyUtils;
+    private final Reporter reporter;
+    private final VerifyAnnotationCompiler annotationCompiler;
     JCTree.JCCompilationUnit currentCompilationUnit;
 
     public JavaToLaurelCompiler(Context context) {
         lowerer = context.get(JavaLowerer.class);
         contractCompiler = MethodOrLoopContractCompiler.instance(context);
         jverifyUtils = JVerifyUtils.instance(context);
+        reporter = Reporter.instance(context);
+        annotationCompiler = VerifyAnnotationCompiler.instance(context);
     }
 
     public record AnalysisResult(List<LaurelFile> files, FilesMap filesMap) {}
@@ -48,6 +54,7 @@ public class JavaToLaurelCompiler {
                 continue;
             }
             currentCompilationUnit = compilationUnit;
+            reporter.compilationUnit = compilationUnit;
             var visitor = new StaticMethodCollector();
             compilationUnit.accept(visitor);
             List<Command> commands = new ArrayList<>();
@@ -90,7 +97,8 @@ public class JavaToLaurelCompiler {
             makeConstrainedType("int8", -128L, 127L),
             makeConstrainedType("int16", -32768L, 32767L),
             makeConstrainedType("int32", -2147483648L, 2147483647L),
-            makeConstrainedType("int64", -9223372036854775808L, 9223372036854775807L)
+            makeConstrainedType("int64", -9223372036854775808L, 9223372036854775807L),
+            makeConstrainedType("char", 0L, 65535L)
         );
     }
 
@@ -116,6 +124,7 @@ public class JavaToLaurelCompiler {
             case SHORT -> compositeType("int16");
             case BYTE -> compositeType("int8");
             case LONG -> compositeType("int64");
+            case CHAR -> compositeType("char");
             case BOOLEAN -> boolType();
             default -> throw new JavaViolationException("Unsupported type: " + type);
         };
@@ -196,77 +205,86 @@ public class JavaToLaurelCompiler {
         @Override
         public void visitMethodDef(JCTree.JCMethodDecl method) {
             if ((method.mods.flags & Flags.STATIC) != 0) {
-                // TODO: when overloaded methods are supported, disambiguate names
-                //  (e.g. by appending parameter type suffixes) to avoid duplicate procedure names in Laurel.
-                String methodName = qualifiedMethodName(method.sym);
-
-                List<Parameter> params = new ArrayList<>();
-                for (var param : method.params) {
-                    params.add(parameter(param.name.toString(), translateType(param.type)));
+                try {
+                    translateStaticMethod(method);
+                } catch (JavaViolationException e) {
+                    reporter.reportError(method, "translatorError", e.getMessage());
+                    annotationCompiler.markSkipped(currentCompilationUnit, method);
                 }
-
-                Optional<ReturnType> retType = Optional.empty();
-                if (method.restype != null && method.restype.type != null
-                        && method.restype.type.getTag() != TypeTag.VOID) {
-                    retType = Optional.of(returnType(translateType(method.restype.type)));
-                }
-
-                List<RequiresClause> requires = new ArrayList<>();
-                List<EnsuresClause> ensures = new ArrayList<>();
-                StmtExpr methodBody = null;
-
-                if (method.body != null) {
-                    MethodOrLoopContract contract = contractCompiler.getContract(method.body);
-                    for (var pre : contract.preconditions()) {
-                        requires.add(requiresClause(convertExpression(pre.get()), Optional.empty()));
-                    }
-                    for (var post : contract.postconditions()) {
-                        var postExpr = post.get();
-                        if (postExpr instanceof JCTree.JCLambda lambda && lambda.params.size() == 1) {
-                            var paramName = lambda.params.getFirst().name.toString();
-                            var renames = Map.of(paramName, "result");
-                            ensures.add(ensuresClause(convertLambdaBody(lambda, renames), Optional.empty()));
-                        } else {
-                            ensures.add(ensuresClause(convertExpression(postExpr), Optional.empty()));
-                        }
-                    }
-
-                    var implStatements = MethodOrLoopContractCompiler.getImplementationStatements(method.body);
-                    if (!implStatements.isEmpty()) {
-                        List<StmtExpr> stmts = new ArrayList<>();
-                        for (var statement : implStatements) {
-                            StmtExpr converted = convertStatement(statement);
-                            if (converted != null) {
-                                stmts.add(converted);
-                            }
-                        }
-                        methodBody = block(toSourceRange(method.body), stmts);
-                    }
-                }
-
-                Optional<Body> optBody = methodBody != null
-                        ? Optional.of(body(methodBody))
-                        : Optional.empty();
-
-                boolean isPure = jverifyUtils.isPure(method.sym);
-                // Strata rejects transparent (visible-body) procedures unless they're functional;
-                // emit an OpaqueSpec to mark the body opaque otherwise. Pure functions stay
-                // transparent only when they have no ensures clauses (the schema can't carry
-                // ensures without an OpaqueSpec wrapper).
-                // modifies is always empty: jverify doesn't yet emit modifies clauses.
-                boolean canStayTransparent = isPure && ensures.isEmpty();
-                Optional<OpaqueSpec> optSpec = canStayTransparent
-                        ? Optional.empty()
-                        : Optional.of(opaqueSpec(ensures, List.of()));
-
-                Procedure proc = isPure
-                        ? function(toSourceRange(method), methodName, params,
-                                retType, Optional.empty(), requires, Optional.empty(), optSpec, optBody)
-                        : procedure(toSourceRange(method), methodName, params,
-                                retType, Optional.empty(), requires, Optional.empty(), optSpec, optBody);
-                procedures.add(proc);
             }
             super.visitMethodDef(method);
+        }
+
+        private void translateStaticMethod(JCTree.JCMethodDecl method) {
+            // TODO: when overloaded methods are supported, disambiguate names
+            //  (e.g. by appending parameter type suffixes) to avoid duplicate procedure names in Laurel.
+            String methodName = qualifiedMethodName(method.sym);
+
+            List<Parameter> params = new ArrayList<>();
+            for (var param : method.params) {
+                params.add(parameter(param.name.toString(), translateType(param.type)));
+            }
+
+            Optional<ReturnType> retType = Optional.empty();
+            if (method.restype != null && method.restype.type != null
+                    && method.restype.type.getTag() != TypeTag.VOID) {
+                retType = Optional.of(returnType(translateType(method.restype.type)));
+            }
+
+            List<RequiresClause> requires = new ArrayList<>();
+            List<EnsuresClause> ensures = new ArrayList<>();
+            StmtExpr methodBody = null;
+
+            if (method.body != null) {
+                MethodOrLoopContract contract = contractCompiler.getContract(method.body);
+                for (var pre : contract.preconditions()) {
+                    requires.add(requiresClause(convertExpression(pre.get()), Optional.empty()));
+                }
+                for (var post : contract.postconditions()) {
+                    var postExpr = post.get();
+                    if (postExpr instanceof JCTree.JCLambda lambda && lambda.params.size() == 1) {
+                        var paramName = lambda.params.getFirst().name.toString();
+                        var renames = Map.of(paramName, "result");
+                        ensures.add(ensuresClause(convertLambdaBody(lambda, renames), Optional.empty()));
+                    } else {
+                        ensures.add(ensuresClause(convertExpression(postExpr), Optional.empty()));
+                    }
+                }
+
+                var implStatements = MethodOrLoopContractCompiler.getImplementationStatements(method.body);
+                if (!implStatements.isEmpty()) {
+                    List<StmtExpr> stmts = new ArrayList<>();
+                    for (var statement : implStatements) {
+                        StmtExpr converted = convertStatement(statement);
+                        if (converted != null) {
+                            stmts.add(converted);
+                        }
+                    }
+                    methodBody = block(toSourceRange(method.body), stmts);
+                }
+            }
+
+            Optional<Body> optBody = methodBody != null
+                    ? Optional.of(body(methodBody))
+                    : Optional.empty();
+
+            boolean isPure = jverifyUtils.isPure(method.sym);
+            // Strata rejects transparent (visible-body) procedures unless they're functional;
+            // emit an OpaqueSpec to mark the body opaque otherwise. Pure functions stay
+            // transparent only when they have no ensures clauses (the schema can't carry
+            // ensures without an OpaqueSpec wrapper).
+            // modifies is always empty: jverify doesn't yet emit modifies clauses.
+            boolean canStayTransparent = isPure && ensures.isEmpty();
+            Optional<OpaqueSpec> optSpec = canStayTransparent
+                    ? Optional.empty()
+                    : Optional.of(opaqueSpec(ensures, List.of()));
+
+            Procedure proc = isPure
+                    ? function(toSourceRange(method), methodName, params,
+                            retType, Optional.empty(), requires, Optional.empty(), optSpec, optBody)
+                    : procedure(toSourceRange(method), methodName, params,
+                            retType, Optional.empty(), requires, Optional.empty(), optSpec, optBody);
+            procedures.add(proc);
         }
 
         private StmtExpr convertBlock(JCTree.JCBlock blk) {
@@ -526,6 +544,7 @@ public class JavaToLaurelCompiler {
                     long val = ((Number) literal.value).longValue();
                     yield longLiteral(sr, val);
                 }
+                case CHAR -> longLiteral(sr, ((Number) literal.value).longValue());
                 default -> throw new JavaViolationException("Unsupported literal type: " + literal.typetag);
             };
         }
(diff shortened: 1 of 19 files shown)


verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -136,6 +136,71 @@ private static String qualifiedMethodName(Symbol.MethodSymbol sym) {
 
     private class StaticMethodCollector extends TreeScanner {
         final List<Procedure> procedures = new ArrayList<>();
+        private int labelCounter = 0;
+
+        /** Label stack entry for break/continue resolution. */
+        private record LabelEntry(String javaLabel, String breakLabel, String continueLabel) {}
+        private final Deque<LabelEntry> labelStack = new ArrayDeque<>();
```
fabiomadge, line 143: The assumption is wrong for labeled break/continue: `break L` from an inner loop CAN target us (the outer labeled loop). When it does, this scan returns false, breakLbl/continueLbl are null, and the resolve returns null — see top-level Bug 1. One-line fix:

    boolean needsLabels =
        pendingLabel != null || containsBreakOrContinue(loopBody);

i.e., always emit labels when this loop is Java-labeled, since labeled break/continue from inside might target us.

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -233,6 +298,64 @@ private StmtExpr convertBlock(JCTree.JCBlock blk) {
             return block(toSourceRange(blk), statements);
         }
 
+        /**
+         * Emit a while loop with optional break/continue label wrapping.
+         *
+         * When labels are present, produces (for a for-loop with break/continue):
+         * <pre>
+         * labelledBlock(breakLbl, {
```
fabiomadge, line 305: Stale — `needsLabels` was inlined as `breakLbl != null` and is no longer a parameter. Drop the line.

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -266,34 +389,77 @@ yield varDecl(toSourceRange(varDecl), varDecl.name.toString(),
...
-                            : literalBool(toSourceRange(forLoop), true);
-                    StmtExpr step = forLoop.step.isEmpty() ? block(toSourceRange(forLoop), List.of())
-                            : convertStatement(forLoop.step.getFirst());
-                    List<InvariantClause> invariants = List.of();
-                    StmtExpr loopBody;
-                    if (forLoop.body instanceof JCTree.JCBlock loopBlock) {
-                        var parts = extractLoopParts(loopBlock);
-                        invariants = parts.invariants;
-                        loopBody = parts.body;
+                    var sr = toSourceRange(forLoop);
```
fabiomadge, line 401: The codebase convention for translator-synthesized names is `$` — Suspenders.java uses `$caseMatch` and `$assertMarker`. `__first_N` is the new outlier, and `__`-prefixed names aren't reserved by Java, so a user `int __first_0` would collide. Suggest `$first_N`.

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -266,34 +389,77 @@ yield varDecl(toSourceRange(varDecl), varDecl.name.toString(),
...
+                    yield withLoopLabels(forLoop.body, (breakLbl, continueLbl) -> {
+                        StmtExpr init = forLoop.init.isEmpty() ? block(sr, List.of())
+                                : convertStatement(forLoop.init.getFirst());
+                        StmtExpr cond = forLoop.cond != null
```
fabiomadge, line 405: The `__first || cond` shape weakens the loop guard for modular invariant reasoning — see top-level Bug 2. Whatever fix lands, this PR needs at least one static do-while + numeric-invariant test to demonstrate the behavior; the existing test (VerifyStatements.doWhileLoop) is vacuous per #420.

verifier/src/test/java/org/strata/jverify/verifier/tests/javasupport/statements/VerifyBreakContinue.java
```diff
@@ -0,0 +1,162 @@
+package org.strata.jverify.verifier.tests.javasupport.statements;
+
+import org.strata.jverify.testengine.JVerifyTest;
+
+import static org.strata.jverify.JVerify.*;
+
+@SuppressWarnings({"ConditionalBreakInInfiniteLoop", "ConstantValue"})
+@JVerifyTest(methodsVerified = 11, errorCount = 0)
```
fabiomadge: The existing tests cover unlabeled break/continue solidly. Missing scenarios that the PR title claims support for:

- Labeled break (the only labeled test, nestedForLoopContinue, is non-static and vacuous per #420). Adding a static version surfaces Bug 1 from the top-level review.
- Do-while + break.
- Do-while + continue (would surface Bug 2).
- `return` inside a labelled-block-wrapped loop.
- A negative case (e.g. `breakBad` with `check(i == 6)` where i ends at 5) — cheap insurance against a silent miscompile to no-op.

fabiomadge (commented):
Implementation matches PLAN.md §3.4's labelledBlock shape. Two real bugs surface when you write the missing tests as static methods.

**Bug 1 — labeled break to outer loop without own break/continue: crashes with exit code 1.** `containsBreakOrContinue` skips into nested loops, so a labeled outer with no break of its own gets `LabelEntry("X", null, null)`. Inner `break X` resolves to that null. Strata sees `(ident null null.string)`. Repro:
```java
static void labeledBreakOuter() {
    int x = 0;
    outerLoop:
    for (int i = 0; i < 5; i = i + 1) {
        invariant(x == i * 5); invariant(i <= 2);
        for (int j = 0; j < 5; j = j + 1) {
            invariant(x == j + i * 5); invariant(i == 2 ? j <= 2 : true);
            if (i == 2 && j == 2) { check(x == 12); break outerLoop; }
            x = x + 1;
        }
    }
    check(x == 12);
}
```

**Bug 2 — do-while desugaring breaks invariant preservation.** `while (__first || cond)` weakens the guard. Strata's preservation rule is `I ∧ guard → wp(body, I)`; with `__first = true` the guard carries no info about cond, so any invariant whose preservation needs the loop bound becomes unprovable. Same body verifies as `while`, fails as `do-while`:
```java
do {
    invariant(0 <= x && x < 100);
    x = x + 1;
} while (x < 5);
```
Hits ~every numeric do-while since int32 overflow checks need bounds and the natural bound is the guard. Possible fix: emit `body; while(cond) body` — labelledBlock makes the break case clean, continue needs two targets but is solvable. Or file a Strata-side ask for a real do-while.

**Why CI didn't catch these.** `VerifyStatements` is non-static. Per #420, non-static methods get `Verified` for free without translation, so `nestedForLoopContinue` and `doWhileLoop` are vacuous. The static-method test set in this PR has no labeled break, no do-while, no `return`-from-loop, no negatives.

**Scope vs PLAN.md §3.4.** The PR descopes #81 (enhanced-for) without mentioning it. PLAN.md prescribes a regression test for arrays + a `notSupported` refusal for `Iterable<T>` — neither lands. Either implement (likely 1 test + 1 line) or split out and note in the PR description.

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -266,34 +388,77 @@ yield varDecl(toSourceRange(varDecl), varDecl.name.toString(),
...
+                }
+                case JCTree.JCDoWhileLoop doWhileStmt -> {
+                    var sr = toSourceRange(doWhileStmt);
+                    yield withLoopLabels(doWhileStmt.body, (breakLbl, continueLbl) -> {
+                        // Desugar: var $first_N = true; while($first_N || cond) { $first_N = false; body }
+                        String sentinel = "$first_" + labelCounter;
+                        StmtExpr sentinelDecl = varDecl(sr, sentinel,
+                                Optional.of(typeAnnotation(boolType())),
+                                Optional.of(initializer(literalBool(sr, true))));
+                        StmtExpr cond = orElse(sr, identifier(sr, sentinel), convertExpression(doWhileStmt.cond));
```
fabiomadge: Bug 2 still reproduces here on 19173575 with the original repro. Either fix the desugaring or refuse do-while with `notSupported` and follow up on a Strata-side ask.

verifier/src/test/java/org/strata/jverify/verifier/tests/javasupport/statements/VerifyBreakContinue.java
```diff
@@ -0,0 +1,215 @@
+package org.strata.jverify.verifier.tests.javasupport.statements;
+
+import org.strata.jverify.testengine.JVerifyTest;
+
+import static org.strata.jverify.JVerify.*;
+
+@SuppressWarnings({"ConditionalBreakInInfiniteLoop", "ConstantValue"})
+@JVerifyTest(exitCode = 4, methodsVerified = 14, methodsInvalid = 1, errorCount = 1)
```
fabiomadge: Two test gaps remain blocking:
- Do-while + continue (parallel to the new doWhileWithBreak; the interaction with the sentinel-reset and labelled-block wrapping is not currently exercised).
- Either an enhanced-for-array regression test, or a `notSupported` refusal site for Iterable<T> with a follow-up issue (#81).

Smaller note on the existing additions:
- `labeledBreakOuter` could use a `check(x == 5)` after the loop. Without it the test asserts only "translator produces valid IR" — a regression that broke to the wrong loop would still pass.
- `returnInsideLoop`'s postcondition line is tab-indented (rest of file uses spaces).

fabiomadge (changes requested):
Thanks for addressing #1, #2, #3, and 4 of 5 inline test items from the prior review — all clean. Three items remain blocking:

**1. Bug 2 (do-while desugaring weakens invariant preservation) — still reproduces.** I re-ran the original repro on 19173575:
```java
do {
    invariant(0 <= x && x < 100);
    x = x + 1;
} while (x < 5);
```
Strata reports "assertion could not be proved" (preservation: with `__first = true`, the guard `__first || x < 5` carries no info about x, so `x = 99` is admissible at body entry → x+1 = 100 violates I). The added `doWhileWithBreak` doesn't exercise this — its conditional `break` absorbs the boundary. The bug fires on the common case `do { x++; ... } while (cond);`. Notably, `VerifyStatements.doWhileLoop` (invariant `x <= 5`, body `x = x + 1`) will start failing verification once #392/Step 3 drops the static gate — counterexample `x = 5, __first = true`. Either fix the desugaring (emit `body; while(cond) body` with a labelled-block break target — works for break; continue needs two targets but is solvable) or refuse do-while with `notSupported` and file a follow-up.

**2. Do-while + continue — no test.** The added do-while test pairs with break; continue is missing. The continue-form is likely affected by the same Bug 2; adding the test either confirms that or surfaces a separate continue-specific issue.

**3. Enhanced-for (#81) — still descoped without acknowledgment.** PLAN.md §3.4 bundles #81 with this step: regression test for arrays + `notSupported "enhanced for loop over Iterable; use array or index-based loop"` for `Iterable<T>`. Neither lands. Either implement (1 test + 1 diagnostic line) or split out and update the PR description.

Smaller items spotted while re-reviewing the new tests:

- `labeledBreakOuter` has no post-loop `check(x == 5)` — verifies translator emits valid IR but doesn't assert break exited the right loop. A regression that breaks to the wrong target would still pass.
- `returnInsideLoop`'s `postcondition(...)` line is indented with a tab while the rest of the file uses spaces. Cosmetic but odd.

olivier-aws:
@fabiomadge thanks for your review and comments. I changed how do-while loops are lowered and improved the tests.

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -266,34 +388,117 @@ yield varDecl(toSourceRange(varDecl), varDecl.name.toString(),
...
+                            // extractLoopParts is called twice: once for the inlined first iteration,
+                            // once for the while body. This mirrors the else branch where two calls are
+                            // required because each compilation uses a different label context (different
+                            // continue targets). Here the results are structurally identical (no labels),
+                            // but we keep the same pattern for clarity.
+                            var firstParts = extractLoopParts(doWhileStmt.body);
+                            StmtExpr cond = convertExpression(doWhileStmt.cond);
+                            var whileParts = extractLoopParts(doWhileStmt.body);
+                            StmtExpr whileNode = while_(sr, cond, whileParts.invariants, whileParts.body);
+                            yield block(sr, List.of(firstParts.body, whileNode));
```
fabiomadge: Soundness gap (Bug 3): `firstParts.body` runs WITHOUT the loop invariant being asserted at its entry. Strata only checks the invariant at while-entry (after the first body) and on preservation. If the invariant is false at the do-while's syntactic entry but the first body restores it, the verifier silently accepts a false invariant claim — see top-level review for repro. Same concern applies to the labels branch at line 466.

Fix: prepend an `assert(I)` for each invariant before `firstParts.body`. The block then becomes:

    block(sr, [
        assert(I_1), ..., assert(I_n),
        firstParts.body,
        whileNode
    ])

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -266,34 +388,117 @@ yield varDecl(toSourceRange(varDecl), varDecl.name.toString(),
...
+                        StmtExpr result;
+                        try {
+                            StmtExpr cond = convertExpression(doWhileStmt.cond);
+                            var whileParts = extractLoopParts(doWhileStmt.body);
+                            StmtExpr wrappedWhileBody = labelledBlock(sr, List.of(whileParts.body), continueLbl);
+                            StmtExpr whileNode = while_(sr, cond, whileParts.invariants, wrappedWhileBody);
+                            List<StmtExpr> outerStmts = new ArrayList<>();
+                            outerStmts.add(firstBody);
+                            outerStmts.add(whileNode);
+                            result = labelledBlock(sr, outerStmts, breakLbl);
```
fabiomadge: Same soundness gap as line 436 — see top-level Bug 3. The labels-branch firstBody (line 449) also runs without an entry invariant check. The asserts need to go before firstBody, inside the outer labelled block, so a break in the firstBody still exits to breakLbl.

verifier/src/test/java/org/strata/jverify/verifier/tests/javasupport/statements/VerifyBreakContinue.java
```diff
@@ -0,0 +1,246 @@
+package org.strata.jverify.verifier.tests.javasupport.statements;
+
+import org.strata.jverify.testengine.JVerifyTest;
+
+import static org.strata.jverify.JVerify.*;
+
+@SuppressWarnings({"ConditionalBreakInInfiniteLoop", "ConstantValue"})
+@JVerifyTest(exitCode = 4, methodsVerified = 16, methodsInvalid = 1, errorCount = 1)
```
fabiomadge: Two test gaps remain:

- A negative do-while regression test for Bug 3, e.g.:
  ```java
      int x = -1;
      do {
          invariant(x >= 0);
  //        ^^^^^^^^^^^^^^^^ Error: assertion does not hold
          x = x + 1;
      } while (x < 5);
  }
  ```

- Either an enhanced-for-array regression test, or a `notSupported` refusal site for `Iterable<T>` with a follow-up issue (#81).

fabiomadge (changes requested):
Bug 2 (do-while desugar) is fixed at the right level — body duplication is the correct shape, and `doWhileNumericInvariant` is a clean regression test. But the new desugaring trades the previous completeness bug for a narrow soundness gap. Two items still open from the prior review.

**Bug 3 — body-duplicated do-while loses the first-iteration invariant check.** The desugaring is `firstBody; while(cond) invariants:[I] body`. Strata checks `I` at while-entry (i.e. after the first body) and on preservation, but never before the first body. If `I` is false at the do-while's syntactic entry but the first body restores it, Strata accepts the program even though the user's `invariant(I)` claim is false where they wrote it.

Minimal repro (passes verification on e5d3bd5c, reject expected):
```java
int x = -1;
do {
    invariant(x >= 0);   // FALSE at first iteration entry (x = -1)
    x = x + 1;
} while (x < 5);
```
The analogous while-form `while (x < 6) { invariant(x >= 0); x = x + 1; }` correctly rejects with "assertion does not hold" at the invariant. The asymmetry is the soundness gap.

Severity: narrow (fires only when initial state violates `I` and first body restores it), but Strata is a verifier — soundness over completeness. Worth fixing.

Fix: synthesize an `assert(I)` for each invariant before `firstBody`, in both the no-labels and labels branches. The invariant expressions can be re-extracted from the loop body's contract (or returned alongside the existing `LoopParts.invariants`).

**Still open from the prior review:**

- **Enhanced-for (#81).** PLAN.md §3.4 bundles #81 with this step: regression test for arrays + `notSupported "enhanced for loop over Iterable; use array or index-based loop"` for `Iterable<T>`. Neither lands. Either implement (1 test + 1 diagnostic line) or split out and update the PR description / file a follow-up.
- **`returnInsideLoop`'s `postcondition(...)` line is tab-indented** while the rest of the file uses spaces. Cosmetic, but trivial to fix while you're here.

fabiomadge:
**Update on Bug 3 fix:** I prototyped an alternative do-while shape in [`prototype/do-while-while-true-exit`](https://github.com/strata-org/jverify/compare/feat-break-continue-label...prototype/do-while-while-true-exit) (commit 58a137d1):

```
do { I; body } while (cond)  →
  labelledBlock(breakLbl, [
    while (true) invariants:[I] {
      labelledBlock(continueLbl, [body]);
      if (!cond) exit(breakLbl);
    }
  ])
```

Trade-offs vs. `body-duplication + assert(I) before firstBody`:

- ✅ Fixes Bug 3 by construction — the while node's invariants are checked at while-entry before every iteration including the first.
- ✅ No body duplication, single `extractLoopParts` call.
- ✅ All existing tests pass (104 / 0 / 63, same as e5d3bd5c).
- ✅ Bug 3 repro (`int x = -1; do { invariant(x >= 0); x = x + 1; } while (x < 5);`) is correctly rejected with "assertion does not hold".
- ✅ Net diff: −24 lines vs. e5d3bd5c.

Implementation gotcha I hit while writing this: the cond-exit must live OUTSIDE the continue-labelled block (so a `continue` falls through to the cond check). My first cut had it inside, which silently skipped the cond check on continue paths and broke `doWhileWithContinue`. Comment in the prototype calls this out.

Either approach (body-duplication + assert, or this) addresses Bug 3. Posting this as a sketch in case the simpler shape is preferable; happy to defer to whichever you prefer.

olivier-aws:
@fabiomadge , thanks for the suggested alternative for the do-while loop desugaring. I believe this deserves a proper thinking about design, the body duplication approach is not great (because hard to duplicate invariants and assertions), your approach seems simpler.
I updated this PR to remove support of do-while loops, updated the PR description as well. I would prefer to defer this support in a follow-up.

fabiomadge (approved):
All my prior review items are resolved:

| Item | Resolution |
|---|---|
| Bug 1 (labeled-break crash) | Fixed in 98222a63 |
| Stale `@param needsLabels` | Fixed in 616b298c |
| Sentinel naming `__first_N` → `$first_N` | Made moot by descope |
| Missing test scenarios | Added in 0d11493f (labeledBreakOuter, returnInsideLoop, breakBad) |
| `labeledBreakOuter` post-loop check | Added in e5d3bd5c |
| Tab indentation on `returnInsideLoop` | Fixed in bf85240f |
| Bug 2 (sentinel weakens guard) | Removed by descope |
| Bug 3 (invariant before first iter) | Removed by descope |
| Do-while + continue test | Removed by descope |
| Enhanced-for (#81) | Acknowledged in PR description |

Do-while tracked separately in #421.

Approving.

PR #426, opened 2026-06-04
Support @Nat and @Unbounded annotations in the Laurel compiler (#402)

### LINKED ISSUE #402: Laurel compiler: @Nat and @Unbounded annotation support
## Context

From the [Strata test triage](https://github.com/strata-org/jverify/blob/chore/strata-test-triage/TRIAGE_RESULTS.md) (Priority 11).

## Problem

The Laurel compiler doesn't process `@Nat` and `@Unbounded` constrained type annotations on parameters and return types. These require generating subset type constraints in the Laurel output.

## Impact

~3 tests unblocked

## Effort

Medium

## Blocked tests

PrimitiveTypes, Termination, FibonacciInvalid (all also blocked by instance method support)

Parent issue: #384

olivier-aws:
## What

Implements #402 — `@Nat` and `@Unbounded` type-use annotation support in the Java→Laurel compiler. These annotations now generate subset type constraints in the Laurel output, which Strata's existing `ConstrainedTypeElim` pass turns into the `requires`/`ensures`/`assert` checks.

## Changes

`JavaToLaurelCompiler`:

- **New predefined subset types**: `nat7`/`nat15`/`nat31`/`nat63` — natural-number (`x >= 0`) counterparts of `int8`/`int16`/`int32`/`int64` with the same upper bounds — and an unbounded `nat` (`x >= 0`, no upper bound). `makeConstrainedType` now takes nullable `min`/`max` bounds so a side can be left unconstrained.
- **`translateType` reads the annotations**: an overload takes the declaration's `JCModifiers` and checks for `@Nat`/`@Unbounded` on both the resolved javac `Type` (works for parameters) and the modifiers (necessary for local variables, where javac strips the type-use annotation off the resolved type/tree). Mapping:
  - default → `int32`/`int16`/…
  - `@Nat` → `nat31`/`nat15`/…
  - `@Unbounded` → unbounded `int`
  - `@Unbounded @Nat` → unbounded `nat`
- Wired into method parameters, local variable declarations, and quantifier lambda params.

## Test

Adds `NatAndUnbounded` — a static-method test exercising `@Nat` on params/locals/reassignment/short/long and `@Unbounded` disabling the overflow check.

## Notes on the issue's "blocked tests"

The three tests named in #402 (PrimitiveTypes, Termination, FibonacciInvalid) remain `skip`-annotated: per the [triage doc](https://github.com/strata-org/jverify/blob/chore/strata-test-triage/TRIAGE_RESULTS.md) they are *also* blocked by instance-method support (the compiler only translates static methods) and a separate Strata-side bug (priority 12) — Strata currently emits `"assertion does not hold"` rather than `"value does not satisfy the subset constraints of '<name>'"`. The new `NatAndUnbounded` test uses the message Strata actually emits, mirroring the existing `CharPrimitive` test.

## Verification

`./gradlew :verifier:test` and `:examples:test` both pass.


verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
kadirayk, line 331: Should `translateType()` also get `method.mods` here?

fabiomadge (dismissed):
Right level and sound — composes the existing `ConstrainedType` machinery (new `nat*` types + annotation reading), no new Laurel node or Strata change. Verified `@Nat` is genuinely enforced (negative to a `@Nat` param rejected at the call site; upper bound holds) and `@Unbounded` drops the overflow bound.

Tests cover the core (`@Nat` lower-bound on param/local/reassign/short/long, `@Unbounded` add). A few wired-but-untested paths worth adding (all work — I checked — so coverage, not correctness): `@Unbounded @Nat` (defined + documented but untested), the `@Nat` upper bound (distinguishes `nat31` from unbounded `nat`), and `@Nat` on a quantifier param.

Minor: `@Nat`/`@Unbounded` on `char`/`boolean` silently no-ops; `makeConstrainedType`'s `reduce().orElse(true)` is dead; return type uses single-arg `translateType` vs `(type, mods)`. LGTM.

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
olivier-aws: Looking into it, good call.

olivier-aws: Not needed, because unlike local variables, javac keeps the @TYPE_USE annotation on the
  resolved return Type, so JVerifyUtils.isAnnotated(type, Nat.class) picks it up via the type-metadata path.

kadirayk (approved):
LGTM!

PR #427, opened 2026-06-08
do-while desugaring

### LINKED ISSUE #421: Laurel compiler: do-while loop support (deferred from #418)
## Context

PR #418 (break/continue/labeled statements support) originally included do-while desugaring, but it was deferred during review. Two desugaring shapes were prototyped during the review:

1. **Body duplication** (in PR #418, commit e5d3bd5c, then reverted in bf85240f): emit `firstBody; while(cond) invariants:[I] body`. Issues:
   - Body duplication doubles translation work (two `extractLoopParts` calls with different label-stack entries to handle `continue`).
   - Soundness gap: invariant is only asserted at while-entry (after the first body), not before the first iteration. Concrete repro:
     ```java
     int x = -1;
     do {
         invariant(x >= 0);   // FALSE at first iteration entry
         x = x + 1;
     } while (x < 5);
     ```
     This passes verification under body-duplication despite the invariant being false where the user wrote it. (The analogous while-form is correctly rejected.)

2. **`while(true)` + `exit(!cond)`** ([prototype branch](https://github.com/strata-org/jverify/tree/prototype/do-while-while-true-exit), commit 58a137d1):
   ```
   labelledBlock(breakLbl, [
     while (true) invariants:[I] {
       labelledBlock(continueLbl, [body]);
       if (!cond) exit(breakLbl);
     }
   ])
   ```
   Properties:
   - Invariant checked at while-entry before EVERY iteration including the first — fixes the soundness gap.
   - No body duplication, single `extractLoopParts` call.
   - Net diff: -24 lines vs body-duplication.
   - All 104 existing tests pass; the soundness repro above is correctly rejected with "assertion does not hold".
   - Implementation gotcha (calls out in the prototype): the `exit(!cond)` must live OUTSIDE the continue-labelled block so a `continue` falls through to the cond check. First draft had it inside, which silently skipped the cond check on continue paths.

## What's needed

Pick a desugaring (the prototype, body-duplication with an entry-invariant assert, or something else), implement it, and re-add the do-while tests that were removed in bf85240f:

- `doWhileWithBreak` (do-while with `break` exiting the loop)
- `doWhileWithContinue` (do-while with `continue` re-evaluating the cond)
- `doWhileNumericInvariant` (numeric loop variable bounded by the invariant — exercises the Bug 2 sentinel-weakening case from the original review)
- A **negative regression test** for the Bug 3 unsoundness above:
  ```java
      int x = -1;
      do {
          invariant(x >= 0);
  //        ^^^^^^^^^^^^^^^^ Error: assertion does not hold
          x = x + 1;
      } while (x < 5);
  }
  ```

## Related

- Parent: #407 (Laurel compiler: break, continue, labeled statements, and decreases clauses).
- PR #418 review thread for the full design discussion.
- Prototype branch: `prototype/do-while-while-true-exit`.

kadirayk:
### What was changed?
Implements do-while loop desugaring as described in #421, i.e., the while(true) + exit(!cond) approach from the prototype branch over the body-duplication approach that was reverted in #418.

Desugaring shape:

```
labelledBlock(breakLbl, [
    while (true) invariants:[I] {
        labelledBlock(continueLbl, [body]);
        if (!cond) exit(breakLbl);
    }
])
```

Why this approach (per the design discussion in #418 and #421):

Invariant is checked at while-entry before every iteration including the first — fixes the soundness gap where body-duplication silently passed invariants that were false at first iteration (Bug 3 in #421)
No body duplication, single extractLoopParts call
Correct continue semantics: `exit(continueLbl)` skips the rest of the body but falls through to the condition check — the if `(!cond) exit(breakLbl)` is outside the continue-labelled block (as called out in #421)

####  Changes: 

`JavaToLaurelCompiler.java`: Added case `JCTree.JCDoWhileLoop` in `convertStatement`, and added `visitDoLoop` override in `containsBreakOrContinue`
`VerifyDoWhile.java:` New test file with the 4 test cases requested in #421
Resolves #421. Parent: #407.

### How has this been tested?
New test file `VerifyDoWhile.java` with the tests specified in #421:

`doWhileWithBreak` — do-while with break exiting the loop
`doWhileWithContinue` — do-while with continue re-evaluating the condition
`doWhileNumericInvariant` — numeric loop variable bounded by the invariant with decreases clause (the Bug 2 sentinel-weakening case from #418)
`doWhileBadInitialInvariant` — negative regression test for the Bug 3 unsoundness (invariant false at first iteration, correctly rejected with "assertion could not be proved")
Additionally:

Existing `cVerifyStatements.doWhileLoop()` passes
Full `./gradlew :verifier:test` suite passes

**Follow-up**: When a do-while invariant fails, the error range covers the entire do-while statement (e.g., do { ... } while (cond);) rather than pointing at just the invariant(...) expression. So `doWhileBadInitialInvariant` in `VerifyDoWhile` uses a single-line do-while — it's the only way the test annotation can match the reported range. I believe this should be improved to provide more a fine-grained error message. The follow up issue: #437

### DIFF
--- a/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
+++ b/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
@@ -152,6 +152,7 @@ public class JavaToLaurelCompiler {
                 // Don't descend into nested loops — their break/continue don't target us
                 @Override public void visitWhileLoop(JCTree.JCWhileLoop tree) {}
                 @Override public void visitForLoop(JCTree.JCForLoop tree) {}
+                @Override public void visitDoLoop(JCTree.JCDoWhileLoop tree) {}
             });
             return found[0];
         }
@@ -415,6 +416,38 @@ public class JavaToLaurelCompiler {
                         return emitLoop(sr, cond, parts.invariants, parts.body, step, List.of(init),
                                 breakLbl, continueLbl);
                     });
+                }
+                case JCTree.JCDoWhileLoop doWhile -> {
+                    // Desugar do-while using while(true) + exit(!cond):
+                    //   labelledBlock(breakLbl, [
+                    //     while (true) invariants:[I] {
+                    //       labelledBlock(continueLbl, [ body ]);
+                    //       if (!cond) exit(breakLbl);
+                    //     }
+                    //   ])
+                    // The exit(!cond) is OUTSIDE the continueLbl block so that
+                    // continue re-evaluates the condition (doesn't skip it).
+                    // Invariant is checked at while-entry before every iteration
+                    // including the first — no soundness gap as discussed in #418.
+                    var sr = toSourceRange(doWhile);
+                    // Do-while always needs labels: the desugaring itself uses exit(breakLbl)
+                    var breakLbl = freshLabel("loop_break");
+                    var continueLbl = freshLabel("loop_continue");
+                    var javaLabel = pendingLabel;
+                    pendingLabel = null;
+                    labelStack.push(new LabelEntry(javaLabel, breakLbl, continueLbl));
+                    try {
+                        var cond = convertExpression(doWhile.cond, renames);
+                        var parts = extractLoopParts(doWhile.body, renames);
+                        var exitIfDone = ifThenElse(sr, not(sr, cond),
+                                exit(sr, breakLbl), Optional.empty());
+                        var wrappedBody = labelledBlock(sr, List.of(parts.body), continueLbl);
+                        var whileBody = block(sr, List.of(wrappedBody, exitIfDone));
+                        var whileNode = while_(sr, literalBool(sr, true), parts.invariants, whileBody);
+                        yield labelledBlock(sr, List.of(whileNode), breakLbl);
+                    } finally {
+                        labelStack.pop();
+                    }
                 }
                 case JCTree.JCBreak breakStmt -> {
                     yield exit(toSourceRange(breakStmt), resolveBreakLabel(breakStmt));
--- a/verifier/src/test/java/org/strata/jverify/verifier/tests/javasupport/statements/VerifyDoWhile.java
+++ b/verifier/src/test/java/org/strata/jverify/verifier/tests/javasupport/statements/VerifyDoWhile.java
@@ -1,0 +1,69 @@
+package org.strata.jverify.verifier.tests.javasupport.statements;
+
+import org.strata.jverify.testengine.JVerifyTest;
+
+import static org.strata.jverify.JVerify.*;
+
+@SuppressWarnings({"ConstantValue"})
+@JVerifyTest(exitCode = 4, methodsVerified = 4, errorCount = 1)
+class VerifyDoWhile {
+    /**
+     * do-while with break exiting the loop early.
+     */
+    static void doWhileWithBreak() {
+        int x = 0;
+        do {
+            invariant(x >= 0 && x <= 3);
+            if (x == 3) {
+                break;
+            }
+            x = x + 1;
+        } while (x < 10);
+        check(x == 3);
+    }
+
+    /**
+     * do-while with continue re-evaluating the condition.
+     * When continue is hit, execution should skip the rest of the body
+     * but still check the while-condition.
+     */
+    static void doWhileWithContinue() {
+        int x = 0;
+        int sum = 0;
+        do {
+            invariant(x >= 0 && x <= 5);
+            invariant(x <= 2 ? sum == x : sum == 2);
+            if (x >= 2) {
+                x = x + 1;
+                continue;
+            }
+            sum = sum + 1;
+            x = x + 1;
+        } while (x < 5);
+        check(sum == 2);
+    }
+
+    /**
+     * Numeric loop variable bounded by the invariant.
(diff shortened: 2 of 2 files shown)


olivier-aws (approved):
Nice work!

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -416,6 +417,38 @@ yield withLoopLabels(forLoop.body, (breakLbl, continueLbl) -> {
...
+                    pendingLabel = null;
+                    labelStack.push(new LabelEntry(javaLabel, breakLbl, continueLbl));
+                    try {
+                        var cond = convertExpression(doWhile.cond, renames);
+                        var parts = extractLoopParts(doWhile.body, renames);
+                        var exitIfDone = ifThenElse(sr, not(sr, cond),
+                                exit(sr, breakLbl), Optional.empty());
+                        var wrappedBody = labelledBlock(sr, List.of(parts.body), continueLbl);
+                        var whileBody = block(sr, List.of(wrappedBody, exitIfDone));
+                        var whileNode = while_(sr, literalBool(sr, true), parts.invariants, whileBody);
```
fabiomadge, line 446: The constant-`true` guard is the root of the completeness gap (finding #1): the exit `assume(¬G)` in LoopElim degenerates to `assume(false)`, so `¬cond` never reaches the post-loop assertion. A head-tested `while(cond)` here would leave `I ∧ ¬cond` available at exit — but that requires peeling the first iteration (body duplication) to keep do-while semantics, which doesn't scale under nesting. Hence the Laurel-level follow-up (strata-org/Strata#1350).

fabiomadge (commented):
Nice work, and the soundness story is solid: the `while(true) + exit(!cond)` shape checks the invariant before every iteration, and the `doWhileBadInitialInvariant` negative test confirms the Bug-3 gap stays closed. The level is right too — Laurel has no do-while node, so front-end desugaring composing existing nodes is the correct approach.

I built and ran this locally; it surfaced one significant completeness gap plus a few test/diagnostic gaps. None are soundness issues.

**1. Completeness gap (main finding).** Because the modeled guard is constant `true`, the verifier never learns `¬cond` at the *normal* condition-exit, so postconditions that depend on the exit condition can't be proved. Concretely, this fails:
```java
int x = 0;
do { invariant(0 <= x && x <= 3); x = x + 1; } while (x < 3);
check(x == 3);   // "assertion could not be proved"
```
The identical head-tested `while` proves `x == 3`. It's not generally rescuable with a stronger invariant. I prototyped the alternative (peel the first iteration, then a head-tested `while`): it closes the gap and stays sound, but duplicates the body → 2ⁿ blow-up under nesting. Since neither front-end encoding is ideal, I filed the clean fix (a body-tested loop encoding) at the Laurel level: **strata-org/Strata#1350**. Suggestion: keep this encoding, add a test that documents the limitation, and link #1350 as the follow-up.

**2. Diagnostic range (the #437 follow-up).** Root cause is in `extractLoopParts`: invariants are built with the single-arg `invariantClause(convertExpression(...))`, which uses `SourceRange.NONE`. Passing `invariantClause(toSourceRange(inv.get()), ...)` fixes the range for **all** loops (while/for too), and would let `doWhileBadInitialInvariant` stop relying on a single-line body.

**3. Unrelated pre-existing NPE found while testing.** `if (cond) return;` (a valueless return as the sole then-branch) NPEs the verifier — common inside do-while bodies. Not introduced by this PR; filed separately as **strata-org/jverify#443**.

Overall: approach is sound and shippable; the asks are the limitation test + the cheap `SourceRange` fix. Inline notes below.

verifier/src/test/java/org/strata/jverify/verifier/tests/javasupport/statements/VerifyDoWhile.java
```diff
@@ -0,0 +1,69 @@
+package org.strata.jverify.verifier.tests.javasupport.statements;
+
+import org.strata.jverify.testengine.JVerifyTest;
+
+import static org.strata.jverify.JVerify.*;
+
+@SuppressWarnings({"ConstantValue"})
+@JVerifyTest(exitCode = 4, methodsVerified = 4, errorCount = 1)
+class VerifyDoWhile {
```
fabiomadge, line 9: These four tests all use postconditions provable **without** the exit guard (break-based exits, or `check(sum == 2)` from the invariant), so they pass despite the completeness gap in finding #1. Missing coverage worth adding:
- a case that exercises the completeness limit (e.g. `do { invariant(0<=x&&x<=3); x=x+1; } while (x<3); check(x==3);`) — as a documented known-limitation or, ideally, after the Laurel fix;
- **nested** do-while and **labeled** do-while (`continue`/`break` to an outer label) — the `javaLabel = pendingLabel` capture in the desugaring is currently untested.

fabiomadge:
Correction to finding #2 (diagnostic range): I'd called it a cheap front-end fix (`toSourceRange(inv.get())` in `extractLoopParts`). **Wrong — built it, clean rebuild confirms no effect.** Strata discards the per-invariant range: `LaurelToCoreTranslator` translates `While` invariants as `("", translateExpr i)`, and `LoopElim` stamps each invariant assert with the *loop's* metadata. So it's a Strata fix, not front-end — and it's general (while/for mislocate to the loop head too, not just do-while), so #437 is really a Strata-side loop-invariant-range issue.

PR #430, opened 2026-06-09
JVerify front-end: set synthetic constructor type to fix javac Lower NPE

tautschnig:
LambdaToAnonymousClassCompiler.createConstructor missed setting result.type, causing javac's Lower.visitMethodDef:2664 to NPE on tree.type.getReturnType() when walking the synthetic anonymous class. Mirror what createImplementationMethod already does.

### DIFF
--- a/verifier/src/main/java/org/strata/jverify/verifier/compiler/simplifications/LambdaToAnonymousClassCompiler.java
+++ b/verifier/src/main/java/org/strata/jverify/verifier/compiler/simplifications/LambdaToAnonymousClassCompiler.java
@@ -215,6 +215,13 @@ public class LambdaToAnonymousClassCompiler extends TreeTranslator {
                 null
         );
         result.sym = methodSymbol;
+        // Without setting result.type, javac's Lower.visitMethodDef
+        // (Lower.java:2664) NPEs on tree.type.getReturnType() when
+        // walking the synthetic anonymous class.
+        // createImplementationMethod sets it for the SAM method; mirror
+        // that here for the synthetic constructor so both methods of the
+        // lambda's anonymous class survive Lower.
+        result.type = methodSymbol.type;
         return result;
     }
 


keyboardDrummer (dismissed):
Can you add a test?

tautschnig:
> Can you add a test?

Test added in a second commit.

fabiomadge (dismissed):
Verified locally: the new test fails on `main` (synthetic constructor left with null `type`) and passes with the fix; `*Lambda*` suite clean and CI is green on the full suite (ubuntu + windows). The fix is exactly symmetric — `createImplementationMethod` already does `result.type = methodSymbol.type`; `createConstructor` was the one synthesized decl missing it, and since these decls are built after javac's attribution they need `.type` set manually. The build.gradle test-export is correct and necessary (the test touches `com.sun.tools.javac.*` from the unnamed module; mirrors the existing `ALL-UNNAMED` exports).

Worth noting the NPE is Java-25-specific (latent on the current Java 23 build), so this is hardening ahead of the #429 migration rather than a crash seen today — but it's a real invariant violation, and the JDK-independent test (running the lowering pass in isolation and asserting the well-formedness invariant directly) is a nice way to lock it. LGTM.

verifier/src/test/java/org/strata/jverify/verifier/compiler/simplifications/LambdaConstructorTypeTest.java
```diff
@@ -0,0 +1,86 @@
+package org.strata.jverify.verifier.compiler.simplifications;
```
keyboardDrummer: Instead of this test, could you add something to `verifier/src/test/java/org/strata/jverify/verifier/tests/javasupport/lambdas/Lambdas.java` that exposes the bug?

The reason is that E2E tests that only test from the users perspective, instead of referencing internals, like Lambdas.java, are less brittle and easier to read.

tautschnig: The problem is that 1) `Lambdas.java` is `@JVerifyTest(skip = "Strata: not yet supported", …)` and 2) this would only become visible when using Java 25. With Java 25 we wouldn't even need any new test - the existing E2E tests already expose the problem.

So this approach was my attempt to have a test exposing the problem even with Java 23. But I do agree that it's brittle and indeed it might just be better to merge this one without tests, merge #429, and thereby have this tested?

keyboardDrummer: > But I do agree that it's brittle and indeed it might just be better to merge this one without tests, merge https://github.com/strata-org/jverify/pull/429, and thereby have this tested?

Sounds good! Feel free to merge after removing the test.

keyboardDrummer (approved):

PR #433, opened 2026-06-09
JVerify front-end: handle loop bodies without a contract block

tautschnig:
### What was changed?

Defensive hardening around contract-bearing loop/method blocks.

`MethodOrLoopContractCompiler.getContractBlock` casts a loop body's first statement to `JCBlock`. If a loop body **without** the contract-bearing shape ever reached it, that cast would throw a `ClassCastException`.

- Adds a `hasContractStructure()` predicate and has `getContractBlock` validate the shape up front, throwing a descriptive `JavaViolationException` (which the per-method handler catches, so the method is skipped gracefully) instead of letting the unchecked cast fail.
- Guards `extractLoopParts` in `JavaToLaurelCompiler` so a loop body without the contract-bearing shape is processed as pure implementation statements with no invariants (threading the caller's `renames` map, like the contract and non-block branches).

### Note: forward-looking, not a currently-reachable crash

This is defensive hardening, **not** a fix for a crash reachable on the current tree. Hand-written contract-free loops are wrapped by `MethodOrLoopContractCompiler` and so already have the expected structure; the only unwrapped loop form is enhanced-for, which needs array/`Iterable` support that errors out first (`Unsupported type: int[]`). For that reason **no regression test is constructible yet**. Soundness is fine: the empty-invariant fallback is conservative, and real `invariant()` clauses are always wrapped.


verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -469,6 +469,22 @@ private record LoopParts(List<InvariantClause> invariants, StmtExpr body) {}
...
             if (body instanceof JCTree.JCBlock loopBlock) {
+                // Only loops authored with an explicit invariant
+                // block (the contract-bearing shape produced by
+                // JVerify's contract authoring convention) have the
+                // structure getContract expects. Loop bodies
+                // without the wrapping contract block — including
+                // javac-synthesized while loops from enhanced-for
+                // desugaring — are processed as pure implementation
+                // statements with no invariants.
+                if (!MethodOrLoopContractCompiler.hasContractStructure(loopBlock)) {
```
fabiomadge, line 570: Can't reach this branch today: the only unwrapped loop is enhanced-for, which needs arrays/`Iterable`, and those error first. If no regression test is constructible yet, worth noting in the description that this is forward-looking hardening rather than a current-crash fix.

verifier/src/main/java/org/strata/jverify/verifier/compiler/simplifications/MethodOrLoopContractCompiler.java
```diff
@@ -268,9 +268,31 @@ public JCTree.JCBlock getDefaultBody() {
     }
 
     public static JCTree.JCBlock getContractBlock(JCTree.JCStatement outerBlock) {
+        if (!hasContractStructure(outerBlock)) {
+            throw new IllegalArgumentException(
```
fabiomadge: Minor: `IllegalArgumentException` isn't a `JavaViolationException`, so it won't be caught by the per-method `catch (JavaViolationException)`. Not reachable in practice (method bodies are always wrapped), but the catchable type would be more consistent.

fabiomadge (commented):
Sound defensive fix; the guard + descriptive exception beat a bare CCE.

Couldn't reproduce the CCE on the current tree — hand-written contract-free loops all get wrapped by `MethodOrLoopContractCompiler`, so they have the expected structure. The only unwrapped case is enhanced-for (no `visitForeachLoop`), which needs array/`Iterable` support, and those error first (`Unsupported type: int[]`). So it's latent hardening, not a currently-reachable crash. Soundness is fine (empty-invariant fallback is conservative; real `invariant()`s are always wrapped).

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
tautschnig: Right, description updated, this needs the array support to be merged.

verifier/src/main/java/org/strata/jverify/verifier/compiler/simplifications/MethodOrLoopContractCompiler.java
tautschnig: Fixed in [95c15cb](https://github.com/strata-org/jverify/pull/433/commits/95c15cb5be89115d31da1f810b3251c81444fc25).

fabiomadge (approved):
Re-checked `95c15cb`: both points addressed (`getContractBlock` throws catchable `JavaViolationException`; unreachable branch documented). Full suite passes, fallback is sound. No regression test yet (not constructible until arrays land — fine as forward-hardening); needs a rebase (`BEHIND`). LGTM.

PR #434, opened 2026-06-09
verifier: add --emit-laurel <path> CLI flag

tautschnig:
### What was changed?

Adds a `--emit-laurel <FILE>` CLI flag that writes the serialized Laurel program (the result of the Java-to-Laurel translation) to `FILE` and **exits without running verification**.

This separates the two halves of the pipeline — compilation (Java → Laurel IR) and verification (Laurel IR → SMT via Strata) — so the emitted Laurel Ion can be fed to Strata independently. (Use case from the discussion: running compilation and verification as separate steps.)

Implementation notes:

- `VerifierOptions` already had an optional `printSerializedOutputProgram` path that `LaurelDriver` writes the Ion to. This wires the picocli flag to that path and adds an `emitLaurelOnly` flag so the driver stops after writing rather than continuing on to the backend.
- The test harness still writes its debug dump to `build/temp.laurel`, but with `emitLaurelOnly = false`, so existing tests verify as before.

### How has this been tested?

- `EmitLaurelWorkflowTest` (new) is an end-to-end test of the split workflow: it emits the Laurel for a small program and asserts that verification did **not** run, then re-reads the emitted Ion and verifies it **separately** through Strata, asserting it verifies cleanly.
- `./gradlew :verifier:test` builds; the new test passes.


keyboardDrummer (approved):

keyboardDrummer (dismissed):
Can I ask what your specific use-case is? Given that the emitting does not interrupt the verification, it seems like more of a debugging feature at this point. I would think options for debugging should be a little more hidden than regular options.

tautschnig:
> Can I ask what your specific use-case is? Given that the emitting does not interrupt the verification, it seems like more of a debugging feature at this point. I would think options for debugging should be a little more hidden than regular options.

The use case is one that would suggest that we _should_ interrupt verification, so I'm happy to make that further change: I'd like to be able to separate compilation (from Java to a Strata IR) from verification (Strata IR all the way to SMT).

tautschnig:
@keyboardDrummer Please re-review - the behaviour is now changed.

fabiomadge (approved):
Verified end-to-end: built the jar, ran `--emit-laurel out.laurel` — writes valid Laurel Ion and exits without invoking Strata. `EmitLaurelWorkflowTest` passes; all three `VerifierOptions` call sites updated; the `emitLaurelOnly` early-exit is well-placed. Clean.

Re the open thread: interrupting verification turns this into a real compile/verify split rather than a passive debug dump, so a visible option seems right. (No `hidden` options exist in `Main` today, so no convention either way — `hidden = true` is there if you want it de-emphasized.)

PR #435, opened 2026-06-09
Support 0-argument precondition/postcondition lambdas (e.g. for void methods)

tautschnig:
### What was changed?

Adds the `postcondition(BooleanSupplier)` overload (the 0-argument form, for void methods or any method whose return value is not needed) **and makes the 0-argument pre/postcondition forms actually work** in the Java → Laurel front end.

Previously the API existed (`postcondition(BooleanSupplier)` is new here; `precondition(BooleanSupplier)` already existed) but **neither was functional**: `translateStaticMethod` only special-cased a 1-parameter postcondition lambda and converted preconditions via `convertExpression`, which has no `JCLambda` case — so a 0-argument `() -> …` raised `Unsupported expression: JCLambda`.

Now:
- **preconditions** — a lambda argument is converted via `convertLambdaBody` (no rename); a plain expression behaves as before.
- **postconditions** — any lambda is converted via `convertLambdaBody`, binding the return value to Laurel's result binding (named via the `LAUREL_RESULT_BINDING` constant) only for the 1-parameter form; the 0-parameter form captures the enclosing scope directly.

### How has this been tested?

Adds a 0-argument pre/postcondition case to `MethodContractsVerification` (rather than a separate `@JVerifyTest`, to avoid the per-test overhead). `./gradlew :verifier:test` passes.

### Known follow-up

A *failing* 0-argument postcondition currently reports at `1:1` (the lambda body's source range is not yet threaded into the diagnostic). Tracked as a follow-up; it does not affect verification outcomes.

### DIFF
--- a/library/src/main/java/org/strata/jverify/JVerify.java
+++ b/library/src/main/java/org/strata/jverify/JVerify.java
@@ -80,6 +80,18 @@ public class JVerify {
     }
 
     /**
+     * Declare a postcondition for a void method (or any method whose return
+     * value is not needed). The {@link BooleanSupplier} takes no arguments and
+     * captures the enclosing scope directly, e.g.
+     * {@code postcondition(() -> x >= 0)}.
+     * <p>
+     * Like the other JVerify contract primitives, this is a marker interpreted
+     * by the verifier and has no effect at runtime.
+     */
+    public static void postcondition(BooleanSupplier predicate) {
+    }
+
+    /**
      * Return the precondition of the given expression
      * Currently only works when the expression is a method call
      * and only works for method calls that return a value.
--- a/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
+++ b/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
@@ -26,6 +26,11 @@ import java.util.*;
 import static org.strata.jverify.laurel.Laurel.*;
 
 public class JavaToLaurelCompiler {
+    /// The name Laurel binds a procedure's return value to inside ensures
+    /// clauses. A 1-parameter postcondition lambda's parameter is renamed
+    /// to this so the clause refers to the return value correctly.
+    private static final String LAUREL_RESULT_BINDING = "result";
+
     private final JavaLowerer lowerer;
     private final MethodOrLoopContractCompiler contractCompiler;
     private final JVerifyUtils jverifyUtils;
@@ -238,13 +243,24 @@ public class JavaToLaurelCompiler {
             if (method.body != null) {
                 MethodOrLoopContract contract = contractCompiler.getContract(method.body);
                 for (var pre : contract.preconditions()) {
-                    requires.add(requiresClause(convertExpression(pre.get()), Optional.empty()));
+                    var preExpr = pre.get();
+                    StmtExpr converted = (preExpr instanceof JCTree.JCLambda lambda)
+                            ? convertLambdaBody(lambda, Map.of())
+                            : convertExpression(preExpr);
+                    requires.add(requiresClause(converted, Optional.empty()));
                 }
                 for (var post : contract.postconditions()) {
                     var postExpr = post.get();
-                    if (postExpr instanceof JCTree.JCLambda lambda && lambda.params.size() == 1) {
-                        var paramName = lambda.params.getFirst().name.toString();
-                        var renames = Map.of(paramName, "result");
+                    if (postExpr instanceof JCTree.JCLambda lambda) {
+                        // A 1-param postcondition lambda binds the return
+                        // value (renamed to Laurel's canonical
+                        // LAUREL_RESULT_BINDING); a 0-param lambda
+                        // (postcondition(BooleanSupplier), e.g. on a void
+                        // method) captures the enclosing scope directly and
+                        // needs no rename.
+                        Map<String, String> renames = lambda.params.size() == 1
+                                ? Map.of(lambda.params.getFirst().name.toString(), LAUREL_RESULT_BINDING)
+                                : Map.of();
                         ensures.add(ensuresClause(convertLambdaBody(lambda, renames), Optional.empty()));
                     } else {
                         ensures.add(ensuresClause(convertExpression(postExpr), Optional.empty()));
--- a/verifier/src/test/java/org/strata/jverify/verifier/tests/verification/MethodContractsVerification.java
+++ b/verifier/src/test/java/org/strata/jverify/verifier/tests/verification/MethodContractsVerification.java
@@ -10,7 +10,7 @@ import java.util.function.IntPredicate;
 import static org.strata.jverify.JVerify.postcondition;
 import static org.strata.jverify.JVerify.precondition;
 
-@JVerifyTest(methodsVerified = 10, errorCount = 0)
+@JVerifyTest(methodsVerified = 11, errorCount = 0)
 public class MethodContractsVerification {
 
     private int y;
@@ -56,6 +56,15 @@ public class MethodContractsVerification {
         int res;
         res = x;
         return res;
+    }
+
+    // 0-argument postcondition (postcondition(BooleanSupplier)): the lambda
+    // takes no parameters and captures the enclosing scope directly rather
+    // than binding the return value. Paired with a 0-argument precondition.
+    int zeroArgPostcondition(int x) {
+        precondition(() -> x > 0);
+        postcondition(() -> x > 0);
+        return x;
     }
 
     /**


keyboardDrummer (dismissed):
Can you add a test?

library/src/main/java/org/strata/jverify/JVerify.java
```diff
@@ -79,6 +79,18 @@ public static void postcondition(BooleanPredicate predicate) {
...
+    /**
+     * Declare a postcondition for a void method (or any method whose return
+     * value is not needed). The {@link BooleanSupplier} takes no arguments and
+     * captures the enclosing scope directly, e.g.
+     * {@code postcondition(() -> counter == old(counter) + 1)}.
+     * <p>
+     * Like the other JVerify contract primitives, this is a marker interpreted
+     * by the verifier and has no effect at runtime.
+     */
+    public static void postcondition(BooleanSupplier predicate) {
```
fabiomadge, line 91: Compiles, but the verifier rejects the 0-arg lambda: `postcondition(() -> ...)` → `Unsupported expression: JCLambda` (extraction only handles 1-param lambdas; 0-param falls to `convertExpression`, which has no `JCLambda` case). `precondition(() -> ...)` same gap. Fix: handle the lambda in pre/postcondition extraction (`convertLambdaBody`, `result` rename only for 1-param).

fabiomadge (commented):
API shape is fine — `BooleanSupplier` is imported and `precondition(BooleanSupplier)` exists. But I built and ran it: `postcondition(() -> ...)` errors `Unsupported expression: JCLambda`, so the feature doesn't work. `precondition(() -> ...)` has the same gap (the existing overload is also non-functional).

Root cause in `translateStaticMethod`: postcondition extraction only special-cases a 1-param lambda; a 0-param lambda falls to `convertExpression`, which has no `JCLambda` case. Preconditions call `convertExpression` directly with no lambda handling at all.

Fix (verified locally — a bounded 0-arg postcondition then verifies):
- precondition: if the arg is a lambda, `convertLambdaBody(lambda, Map.of())`, else `convertExpression`.
- postcondition: extend the lambda branch to 0 params too, keeping the `result` rename only when `params.size() == 1`.

Two caveats: a failing 0-arg postcondition currently reports at `1:1` (the lambda body carries no source range — worth fixing), and the description's `old()`/instance-field example isn't supported yet, so a test should use a static-method example. Needs the front-end change + a verifying test before merge.

library/src/main/java/org/strata/jverify/JVerify.java
tautschnig: Addressed in [e2e532d](https://github.com/strata-org/jverify/pull/435/commits/e2e532dc213fd8f6d97f6d56e6a27d6ccbac32b4).

fabiomadge (dismissed):
The rework matches what I had in mind, and `VoidPostcondition` covers @keyboardDrummer's test ask. Re-checked locally: full suite passes; 0-arg pre/postconditions verify and are properly enforced (`postcondition(() -> false)` fails with a precise location); 2-param lambdas are compile-rejected, so the 0-vs-1-param branch can't be mis-hit.

Worth noting it also repairs the pre-existing `precondition(BooleanSupplier)`, which was silently broken (errored `Unsupported expression: JCLambda`). Present-day uses are limited — without `old()`/fields/`this`, a 0-arg postcondition on a static method can only restate inputs — but it's the right primitive for void-method postconditions once state lands, and it makes an already-advertised API actually work. LGTM.

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -238,13 +238,23 @@ private void translateStaticMethod(JCTree.JCMethodDecl method) {
...
-                        var paramName = lambda.params.getFirst().name.toString();
-                        var renames = Map.of(paramName, "result");
+                    if (postExpr instanceof JCTree.JCLambda lambda) {
+                        // A 1-param postcondition lambda binds the return
+                        // value (renamed to Strata's canonical `result`);
+                        // a 0-param lambda (postcondition(BooleanSupplier),
+                        // e.g. on a void method) captures the enclosing
+                        // scope directly and needs no rename.
+                        Map<String, String> renames = lambda.params.size() == 1
+                                ? Map.of(lambda.params.getFirst().name.toString(), "result")
```
keyboardDrummer: "result" is a magic string. Does the string matter?

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -238,13 +238,23 @@ private void translateStaticMethod(JCTree.JCMethodDecl method) {
...
+                    requires.add(requiresClause(converted, Optional.empty()));
                 }
                 for (var post : contract.postconditions()) {
                     var postExpr = post.get();
-                    if (postExpr instanceof JCTree.JCLambda lambda && lambda.params.size() == 1) {
-                        var paramName = lambda.params.getFirst().name.toString();
-                        var renames = Map.of(paramName, "result");
+                    if (postExpr instanceof JCTree.JCLambda lambda) {
+                        // A 1-param postcondition lambda binds the return
+                        // value (renamed to Strata's canonical `result`);
```
keyboardDrummer: Changing `Strata` into `Laurel` would be more specific.

However, I don't think Laurel should have a canonical `result` and also that JVerify should not rely on something like that. Is this code relying on that?

verifier/src/test/java/org/strata/jverify/verifier/tests/verification/VoidPostcondition.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Exercises the 0-argument {@code precondition(BooleanSupplier)} /
+ * {@code postcondition(BooleanSupplier)} forms (used e.g. for void methods),
+ * which capture the enclosing scope directly rather than binding a return
+ * value. Uses a static method with a parameter reference, since {@code old(...)}
+ * and instance fields are not yet supported.
+ */
+@JVerifyTest(methodsVerified = 2, errorCount = 0)
+class VoidPostcondition {
```
keyboardDrummer: There is some testing time overhead to every `@JVerifyTest`. I would prefer having this test as a test-case in `verifier/src/test/java/org/strata/jverify/verifier/tests/verification/MethodContractsVerification.java`

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
tautschnig: Renamed `Strata` → `Laurel` in the comment, and extracted the bareword into a documented `LAUREL_RESULT_BINDING` constant (1722853).

On the substantive question — yes, this relies on it: a 1-parameter postcondition lambda's parameter is renamed to `result`, and that resolves only because Laurel binds a procedure's return value to `result` inside `ensures` clauses (there's no explicit return-parameter declaration on our side). That coupling is pre-existing — it's how the 1-param postcondition form already worked on `main`; this PR only adds the 0-parameter form, which captures the enclosing scope directly and needs no rename.

I agree it's not ideal for JVerify to depend on a magic name. Making it explicit — e.g. emitting a named return parameter and referencing that instead of the bareword — would remove the reliance, but it changes the existing 1-param mechanism, so I'd suggest doing it as a separate follow-up rather than folding it into this "add the 0-arg form" change. Happy to file/take that follow-up if you'd like.

verifier/src/test/java/org/strata/jverify/verifier/tests/verification/VoidPostcondition.java
tautschnig: Done in [1722853](https://github.com/strata-org/jverify/pull/435/commits/1722853922424d2e70cd42d1550f33a82aebc087).

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
tautschnig: See https://github.com/strata-org/jverify/pull/435#discussion_r3405572738.

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
tautschnig: See also #431 - so we should eventually fix this on the Laurel side, but I'd argue to first get this PR in and then rework that small piece once the Laurel change is done.

fabiomadge (approved):
Re-approving after the push — the change vs my earlier (dismissed) approval is just the `LAUREL_RESULT_BINDING` constant extraction and folding the test into `MethodContractsVerification` (10→11). Substance unchanged; re-verified it compiles and the contracts suite passes.

PR #436, opened 2026-06-09
Add an assigns(...) write-frame declaration to the JVerify contract API

tautschnig:
Introduce JVerify.assigns(Object... targets) alongside the existing contract-declaration primitives (decreases / reads / modifies). Each argument names a target the method is permitted to write — a field, an array slot, a static, and so on. The intended semantics, per the Javadoc, is that at a modular call site only the named targets are havocked while everything else is preserved, mirroring CBMC's __CPROVER_assigns(...).

Like the other primitives in this class the method has an empty body: it is a marker recognized as a contract annotation, not executed at runtime. This commit adds only the public method and its Javadoc to the library module; no other files change.


tautschnig:
We already have `modifies`. Closing.

PR #438, opened 2026-06-09
JavaToLaurelCompiler: collapse reference comparison against null to a constant

tautschnig:
In JVerify's array-as-map model an array-typed value is never null, so `arr == null` / `arr != null` (detected via a BOT-typed null-literal operand) is folded to a boolean constant — `false` / `true` respectively — before operand conversion. This keeps the BOT-typed null literal off the `convertLiteral` path.

The fold is **restricted to comparisons whose non-null operand is array-typed** (map-backed), so it does not silently mask the nullability of ordinary object references — those fall through to the normal conversion path. The degenerate `null == null` / `null != null` case folds to `true` / `false` respectively.

### DIFF
--- a/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
+++ b/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
@@ -558,6 +558,36 @@ public class JavaToLaurelCompiler {
         }
 
         private StmtExpr convertBinary(JCTree.JCBinary binary, Map<String, String> renames) {
+            // Handle comparisons against the `null` literal before
+            // recursing into operand conversion. In JVerify's
+            // array-as-map model an array-typed value is never null, so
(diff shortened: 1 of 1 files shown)


fabiomadge (commented):
This is unreachable until array support lands — array-typed operands error at `translateType` today (verified), so `arr == null` never reaches this fold. #441 (array support) is what makes it reachable, and notably #441 doesn't need this — it doesn't touch `arr == null` handling at all.

Suggestion: punt this until arrays land, and fold it into #441 (or a follow-up on top of it). Reasons: (1) it can't be tested in isolation today; (2) the "arrays are never null" assumption is only meaningful once #441's `mapType(int, int)` model exists — at which point it's verifiable; (3) the `instanceof Type.ArrayType` check would also match `@Nullable int[]`, harmless now (`@Nullable` is a no-op — `isNullable` has zero call sites) but worth settling alongside the real array+null model rather than ahead of it.

The logic itself looks reasonable; it's just landing ahead of the thing it depends on.

PR #439, opened 2026-06-09
JavaToLaurelCompiler: fall back to a 1:1 position for unknown line maps

tautschnig:
### What was changed?

When Strata reports a diagnostic for a node that carries no source location, it serializes it against a synthesized `<unknown>` source path, which has no entry in our line map. Previously a missing line map threw, masking the real Strata error in the user's verdict.

This PR addresses that at the root: **`JavaToLaurelCompiler` now threads a source range from the originating Java tree into every node it synthesizes** — method parameters, return types, and the `requires` / `ensures` / loop-invariant clauses (all previously constructed with the *no-source* builder overloads, i.e. `SourceRange.NONE`). The loop-invariant ranges pair with Strata's per-invariant source ranges (#1361). So a Strata diagnostic on any node jverify constructs now points at the user's code instead of `<unknown>`.

For the residual case where a diagnostic still lands on a node jverify does **not** construct (e.g. Strata's implicit `result` *return* binding — see the test below), the `<unknown>` fallback is kept as defense-in-depth but is now **loud**: it emits a `[jverify] internal: … source-less synthesized node …` warning naming it a synthesis bug, instead of silently returning a misleading `1:1`. Any *other* unmapped URI keeps the previous fail-fast behaviour, since that indicates a real line-map collection bug we do not want to hide.

### How has this been tested?

`UnknownLineMapFallbackTest` drives the verifier on a method with a parameter named `result` (whose Strata `result` collision is reported against the synthetic `<unknown>` URI) and asserts the diagnostic **surfaces** — rather than crashing with `Could not find line map` — **and** that the residual fallback is flagged loudly. Full verifier suite green (131 run, 0 failures). Verified it fails (with the crash) when the fallback is removed.

### Note: coupling with #431

The `result`-parameter collision used as the trigger is the same one **PR #431** fixes (by renaming the user parameter). The two are complementary — this one makes the error *visible* (and points jverify-synthesized diagnostics at real locations), #431 *removes* the collision. Consequently the test relies on pre-#431 behaviour to reach the synthetic-URI path.

### DIFF
--- a/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
+++ b/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
@@ -97,21 +97,25 @@ public class JavaToLaurelCompiler {
         FilesMap filesMap = (uri, offset) -> {
             var lineMap = lineMaps.get(uri);
             if (lineMap == null) {
-                // Strata reports diagnostics for trees injected by a
-                // simplification pass against a synthesized "<unknown>"
-                // source path (e.g. ArrayCompiler's lowering of an
-                // initializer-list array). For that specific synthetic
-                // URI, return a 1:1 fallback Position so the diagnostic
-                // still threads through and surfaces as a user-visible
-                // Verifier error instead of being masked by an
-                // exception (the URI is still reported, so a developer
-                // inspecting raw_stderr can see the synthetic origin).
+                // jverify threads a source range from the generating Java
+                // tree into every synthesized clause/binding (see the
+                // requires/ensures/invariant construction above), so a
+                // diagnostic should normally carry a real user-source
+                // location. If one still arrives against Strata's synthetic
+                // "<unknown>" path, some synthesis site failed to thread its
+                // source -- a jverify bug. Surface it LOUDLY (so it is not
+                // hidden behind a misleading location) but do not crash: a
+                // 1:1 fallback keeps the underlying Strata diagnostic visible
+                // to the user instead of masking it with an exception.
                 //
-                // For any other unmapped URI we keep failing fast: that
-                // indicates a real line-map collection bug, which we do
-                // not want to hide behind a misleading 1:1 location.
+                // Any OTHER unmapped URI indicates a real line-map collection
+                // bug; keep failing fast there.
                 String path = uri.getPath();
                 if (path != null && path.endsWith(SYNTHETIC_UNKNOWN_PATH)) {
+                    System.err.println("[jverify] internal: a Strata diagnostic was reported "
+                            + "against a source-less synthesized node (" + uri + "); a synthesis "
+                            + "pass did not thread a source range from its generating context. "
+                            + "Falling back to 1:1 -- please report this as a jverify bug.");
                     return new Position(1, 1);
                 }
                 throw new RuntimeException("Could not find line map for " + uri);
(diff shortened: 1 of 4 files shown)


fabiomadge (commented):
Correct and well-scoped — narrowing the fallback to the synthetic `/<unknown>` URI (and keeping fail-fast for any other unmapped URI) is the right call; a blanket 1:1 fallback would hide real line-map collection bugs.

Verified it does what it claims: applied it alone and re-ran a case that previously crashed with `Could not find line map` (a method with a `result` parameter, from #431). It now surfaces the real Strata diagnostic — `Duplicate definition 'result' is already defined in this scope` — instead of throwing. So this also unmasks the underlying error in #431's scenario; the two are complementary (this makes it visible, #431 fixes the collision).

Two small things: (1) no test — since the whole point is "diagnostic instead of crash," a test that hits the synthetic path and asserts a user-visible error would lock it (the `result`-param case reliably triggers it); (2) `endsWith("/<unknown>")` couples to javac's synthetic filename — fine, but a brittle string match.

tautschnig:
> Two small things: (1) no test — since the whole point is "diagnostic instead of crash," a test that hits the synthetic path and asserts a user-visible error would lock it (the `result`-param case reliably triggers it); (2) `endsWith("/<unknown>")` couples to javac's synthetic filename — fine, but a brittle string match.

Both addressed in [193d5da](https://github.com/strata-org/jverify/pull/439/commits/193d5da27495bd6fc284173c449451631b310762).

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -30,6 +30,10 @@ public class JavaToLaurelCompiler {
     /// clauses. A 1-parameter postcondition lambda's parameter is renamed
     /// to this so the clause refers to the return value correctly.
     private static final String LAUREL_RESULT_BINDING = "result";
+    /// javac's synthetic source path for trees injected by a
+    /// simplification pass; Strata reports diagnostics for them against
```
keyboardDrummer, line 34: For context, I think that reporting a diagnostic on a generated node indicates a compiler bug. Do you feel the same?

Secondly, I think that despite the above characterization of a bug, that for debugging purposes, it's better to give all synthesis nodes a source location. That source location should originate from the context that is generating the node. Getting a (1,1) diagnostic at any point is quite frustrating for debugging. What do you think?

tautschnig: Agreed on both counts.

(1) Yes — a user-facing diagnostic on a source-less generated node is a defect on our side. Here it was mostly that `JavaToLaurelCompiler` built parameters, return types, and the requires/ensures/invariant clauses through the *no-source* builder overloads (`SourceRange.NONE`) despite the originating Java tree having a real range.

(2) Agreed — fixed in 204bd2c: it now threads `toSourceRange(<originating tree>)` into everything it synthesizes (parameters, return types, requires/ensures/loop-invariant clauses; the invariant ranges pair with #1361). So a diagnostic on any node jverify constructs now points at the user's code.

Honest caveat: this PR's `result`-collision trigger isn't reported against a node jverify builds — Strata raises it against its own implicit `result` *return* binding, which we never construct, so there's no source to thread. For that residual class I kept the fallback but made it **loud** (a `[jverify] internal: … source-less synthesized node …` warning naming it a synthesis bug) rather than a silent 1:1 — surfacing (1) instead of hiding it. (#431 removes this collision anyway -- I'll return to that PR next, trying to address it at the Laurel source as suggested in #431.)

verifier/src/test/java/org/strata/jverify/verifier/UnknownLineMapFallbackTest.java
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * <p>Note: the underlying collision is removed by #431 (which renames the user
+ * parameter), so this test relies on pre-#431 behaviour to reach the
+ * synthetic-URI path.
+ */
+public class UnknownLineMapFallbackTest {
+
+    @Test
+    public void syntheticUnknownUriSurfacesDiagnosticLoudlyInsteadOfCrashing() throws Exception {
+        String source = """
```
keyboardDrummer, line 41: I'm guessing this test will break when we solve the Laurel/Strata bug, right?

keyboardDrummer (approved):
Looks good, thanks!

verifier/src/test/java/org/strata/jverify/verifier/UnknownLineMapFallbackTest.java
tautschnig: It actually should, yes.

PR #444, opened 2026-06-10
Test cases for documenting the limitation of do-while desugaring and for nested/labeled control flow

kadirayk:
#### 1. Document Completeness Gap 
Added two paired tests to `VerifyDoWhile` that document the known completeness gap in the current `while(true) + exit(!cond)` do-while desugaring (the encoding shipped in #427 as identified in [the review](https://github.com/strata-org/jverify/pull/427#pullrequestreview-4462430268)).

Because the desugaring models the loop guard as the constant `true`, the verifier never learns `!cond` at the normal condition-exit. Any postcondition that depends on the exit condition is therefore unprovable, even though it holds for the equivalent head-tested `while`.

- `doWhileForgetsExitCondition`: a `do { invariant(0 <= x && x <= 3); x = x + 1; } while (x < 3);` followed by `check(x == 3)`. The check is correctly proved for a head-tested loop but is **rejected** here ("assertion could not be proved"). This is not rescuable with a stronger invariant. The test pins the current (rejected) behavior so the limitation stays visible and the eventual fix is detectable.
- `whileKeepsExitCondition`: the head-tested `while` counterpart with the identical body, invariant, and postcondition. The same `check(x == 3)` **verifies**, because the modeled guard `x < 3` leaves `x >= 3` available at exit, which with the invariant `x <= 3` gives `x == 3`. This acts as the positive control demonstrating exactly what the do-while encoding loses.

Both tests reference the Laurel-level follow-up (a body-tested loop encoding) tracked in strata-org/Strata#1350, which would close the gap.

#### 2. Nested and Labeled Control Flow
Extended `VerifyDoWhile` with three new verification tests that exercise the do-while desugaring under nested and labeled control flow:

- `nestedDoWhileBreak`: two stacked do-while loops, each exited via an unlabeled `break`. Confirms the inner `break` resolves to the inner loop's break label and the outer `break` to the outer loop's, leaving each loop variable at its break-time value. Postconditions derive from the break guards, so they're independent of the exit-condition gap.
- `labeledDoWhileBreakToOuter`: `break outer` from inside a nested loop. Exercises the `javaLabel = pendingLabel` capture in the desugaring: the outer loop's label frame must carry `"outer"` so the break targets the outer break label rather than the inner one. The break fires on a known state (`i == 2`), so the post-loop assertion is provable.
- `labeledDoWhileContinueToOuter`: `continue outer` from inside a nested loop, re-evaluating the outer condition rather than the inner one. 

### How has this been tested?

- `./gradlew :verifier:test --tests "*VerifyDoWhile*"` passes. the `do-while` check is reported as "assertion could not be proved", the `while` check verifies clean. Nested and labeled control flow tests pass as well.

### DIFF
--- a/verifier/src/test/java/org/strata/jverify/verifier/tests/javasupport/statements/VerifyDoWhile.java
+++ b/verifier/src/test/java/org/strata/jverify/verifier/tests/javasupport/statements/VerifyDoWhile.java
@@ -5,7 +5,7 @@ import org.strata.jverify.testengine.JVerifyTest;
 import static org.strata.jverify.JVerify.*;
 
 @SuppressWarnings({"ConstantValue"})
-@JVerifyTest(exitCode = 4, methodsVerified = 5, errorCount = 2)
+@JVerifyTest(exitCode = 4, methodsVerified = 8, errorCount = 2)
 class VerifyDoWhile {
     /**
      * do-while with break exiting the loop early.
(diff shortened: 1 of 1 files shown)


fabiomadge (approved):
Verified against merged #427: all pass. Nice — covers the completeness gap (`doWhileForgetsExitCondition` rejected / `whileKeepsExitCondition` proved) plus the nested and labeled break/continue-to-outer cases. Note `doWhileForgetsExitCondition` intentionally pins a *rejected* outcome, so Strata#1350 will need it flipped to a positive assertion when it lands. Only early `return` from a do-while body is still uncovered (blocked by the bare-return NPE, aws/jverify#443 / Strata#1353) — fine as a follow-up. LGTM.

PR #446, opened 2026-06-14
builtin-contracts: add pure List.subList and List.equals contracts

tautschnig:
Extends ListContract with two pure methods:

  subList(from, to): precondition 0<=from<=to<=size; postcondition
    result.size()==to-from and forall i. result.get(i)==get(from+i).
  equals(o): extensional — same length and pointwise-equal elements.

Both are @Pure and reuse the existing get/size contracts. No behavioural change to the already-present of/get/size/isEmpty/ contains contracts.

### How has this been tested?

### DIFF
--- a/builtin-contracts/src/main/java/org/strata/jverify/builtin/ListContract.java
+++ b/builtin-contracts/src/main/java/org/strata/jverify/builtin/ListContract.java
@@ -76,4 +76,39 @@ abstract class ListContract<E> implements List<E> {
                         0 <= i && i < size() && get(i).equals(o)));
         throw new ContractException();
     }
+
+    @Override
+    @Pure
+    public List<E> subList(@Nat int fromIndex, @Nat int toIndex) {
+        reads(everything());
+        decreases(0);
+        precondition(0 <= fromIndex && fromIndex <= toIndex && toIndex <= size());
+        // The sublist has length toIndex-fromIndex and its element i is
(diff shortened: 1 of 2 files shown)


fabiomadge (approved):
Approving. Both specs are correct by inspection: `subList`'s forall bounds guard + `r.size() == toIndex - fromIndex` keeps each `get(fromIndex + i)` in-bounds, and reference `==` for elements is right (a view -> same references), matching the `of(...)` contracts. `equals` is exactly `List#equals` -- `other.size() == size()` before the forall via `&&` short-circuit, and `jequals` (not `==`) for elements.

They're inert today, but not because of this PR: no built-in contract's postconditions produce verification facts at the current Strata pin. Standalone, `s.size() == 99` on a 3-element list verifies clean, as does `"abc".length() == 99` -- contracted methods return unconstrained values, so properties over them pass vacuously. Contracts still provide translation (without `--contract-path`, `List.of(...)` errors `Unsupported type: E`); only the value facts don't flow. Hence the whole `jcl/` suite is uniformly `skip = "Strata: not yet supported"`.

So these are exactly as verifiable as the existing `get`/`contains`/`of` contracts here -- correct, and dormant until Strata's contract-fact support lands. Landing them now is fine; the `skip` stays accurate, and lifting it later (suite-wide) will need cases that genuinely fail today plus updated counts.

PR #448, opened 2026-06-14
JVerify: emit @Pure functions with loop bodies as uninterpreted

tautschnig:
A Strata function body may contain only let-bindings and a final if-then-else -- no loops. A `@Pure` Java helper whose body contains a `for`/`while`/`foreach` loop was emitted as a transparent function body, which the Laurel-to-Core translator then rejected with "block expression should have been lowered in a separate pass".

**Fix:** detect a loop in a `@Pure` function body (`bodyContainsLoop`) and drop the body, emitting the function uninterpreted -- a sound over-approximation (congruent, usable in spec positions) instead of an untranslatable block.

Dropping the body is only sound when the function has no postcondition: an uninterpreted function cannot be checked against an `ensures`, so a dropped body would silently discard the postcondition (a contradictory postcondition would vacuously "verify"). The translator now refuses `@Pure` + loop + postcondition with a clear "not yet supported" diagnostic instead, matching the existing non-loop `@Pure` + postcondition behaviour.

### Testing

Two runnable (non-skipped) regression tests:

- `PureMethodWithLoop`: three `@Pure` helpers using `for`/`while`/`do` loops translate without crashing, their uninterpreted results are congruent (provable), and the interpreted loop result is correctly not provable. The helpers return `boolean` to avoid the unrelated "constrained return types on functions" limitation.
- `PureLoopWithPostcondition`: a `@Pure` loop helper with a contradictory postcondition is reported as not-yet-supported rather than vacuously verifying.

The `foreach` arm of `bodyContainsLoop` is future-proofing: it is not yet reachable (no supported iterable type, and no enhanced-for translation case), so it cannot be exercised by a test yet.

### DIFF
--- a/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
+++ b/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
@@ -257,6 +257,24 @@ public class JavaToLaurelCompiler {
             return found[0];
         }
 
+        /** Check if a method body contains any loop. A @Pure function with
+         *  a loop cannot be a transparent (pure-expression) function body --
+         *  Strata function bodies allow only let-bindings and a final
+         *  if-then-else -- so we emit such a function uninterpreted. */
+        private boolean bodyContainsLoop(JCTree.JCBlock body) {
+            boolean[] found = {false};
+            body.accept(new TreeScanner() {
+                @Override public void visitForLoop(JCTree.JCForLoop tree) { found[0] = true; }
+                // Foreach is future-proofing: it is not yet reachable here
+                // (there is no supported iterable type and no enhanced-for
+                // translation case), so it cannot be exercised by a test yet.
+                @Override public void visitForeachLoop(JCTree.JCEnhancedForLoop tree) { found[0] = true; }
+                @Override public void visitWhileLoop(JCTree.JCWhileLoop tree) { found[0] = true; }
+                @Override public void visitDoLoop(JCTree.JCDoWhileLoop tree) { found[0] = true; }
+            });
+            return found[0];
+        }
+
         private String freshLabel(String prefix) {
             return prefix + "_" + (labelCounter++);
         }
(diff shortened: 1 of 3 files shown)


verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
```diff
@@ -321,11 +336,20 @@ private void translateStaticMethod(JCTree.JCMethodDecl method) {
...
 
+            boolean isPure = jverifyUtils.isPure(method.sym);
+            // A @Pure function whose body contains a loop cannot be a
+            // transparent (pure-expression) function body, so drop the
+            // body and emit it uninterpreted (a sound over-approximation)
+            // rather than producing an untranslatable block expression.
+            if (isPure && method.body != null && methodBody != null
+                    && bodyContainsLoop(method.body)) {
+                methodBody = null;
+            }
```
fabiomadge, line 411: `ensures` is in scope here, so refuse that case and reuse the translatorError-skip path:

```suggestion
            boolean isPure = jverifyUtils.isPure(method.sym);
            // A @Pure function whose body contains a loop cannot be a
            // transparent (pure-expression) function body, so drop the
            // body and emit it uninterpreted (a sound over-approximation)
            // rather than producing an untranslatable block expression.
            if (isPure && method.body != null && methodBody != null
                    && bodyContainsLoop(method.body)) {
                if (!ensures.isEmpty()) {
                    // Dropping the body would silently discard the
                    // postcondition (it cannot be checked against an
                    // uninterpreted function), so refuse instead.
                    throw new JavaViolationException(
                        "@Pure function with a loop and a postcondition is not yet supported");
                }
                methodBody = null;
            }
```

fabiomadge (changes requested):
Core approach is sound and the test is solid — uninterpreted is a valid over-approximation, and `PureMethodWithLoop` validates it end-to-end.

One gap before merge: a `@Pure` looping function that **also has a postcondition** drops its body to `.Opaque _ none _`, hitting the silent `_ => pure none` arm in `LaurelToCoreTranslator` (684) — the postcondition is **silently discarded**. Repro: `@Pure` + loop + `postcondition((boolean r) -> r == true && r == false)` verifies clean (exit 0); without the loop it reports `functions with postconditions are not yet supported`. Callers are safe (they don't assume the bogus `ensures`), but the author gets no signal it was dropped. (Verified the inline fix locally: it compiles, reports instead of silently passing, and `PureMethodWithLoop` still passes.)

Also: a regression test for the postcondition case, and one for enhanced-for (`bodyContainsLoop` handles it but it's untested).

verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
tautschnig: Thanks Fabio — applied your suggestion in 42fed277: `@Pure` + loop + postcondition now throws the `JavaViolationException` and is reported via the translatorError-skip path instead of silently dropping the `ensures`.

Added `PureLoopWithPostcondition` as a runnable regression test — the contradictory-postcondition case now reports "not yet supported" (exit 2) rather than vacuously verifying.

On enhanced-for: I left it untested, because it isn't reachable yet. There's no supported iterable type (`int[]` reports `Unsupported type: int[]`, collections are unsupported) and `convertStatement` has no enhanced-for case, so the body fails to translate before the drop is consulted. I kept the `visitForeachLoop` arm as future-proofing, with a comment noting this.

fabiomadge (approved):
Re-approving — my requested changes are addressed. The guard is applied (`if (!ensures.isEmpty()) throw … "not yet supported"`), and `PureLoopWithPostcondition` is a sound regression test: a contradictory postcondition that would vacuously verify if silently dropped, now asserted to refuse with `exitCode = 2` (correct — a translatorError without `continueOnErrors` returns USAGE/2). The foreach case is honestly documented as not-yet-reachable rather than tested. CI green on both platforms; the PR touches only the compiler + two test files (no Strata pin / generated AST).

Issue #443, opened 2026-06-09
NPE on `if (cond) return;` — null statement passed unguarded into a Laurel node


fabiomadge:
## Summary

An `if` whose then-branch is a valueless `return;` crashes the verifier during Laurel serialization:

```
java.lang.NullPointerException: Cannot invoke
"org.strata.jverify.laurel.Node.toIon(...)" because "node" is null
```

This is a crash on valid, already-supported Java.

## Reproduce

Each of these triggers the NPE (confirmed against `main`):

```java
static void m() { int x = 5; if (x == 5) return; check(false); }          // if-then bare return
static void m() { int x = 5; lbl: return; }                                // labeled bare return
static void m() { int x = 0; do { if (x == 2) return; x = x + 1; } while (x < 3); }  // inside a loop
```

Confirmed **not** to trigger (delimits the bug):

```java
static void m() { int x = 5; if (x == 5) { return; } else { x = 0; } check(x == 0); }  // else-branch is null-guarded
static void m() { int x = 1; check(x == 1); return; }                                  // trailing bare return: filtered by callers
static void m() { int x = 0; while (x < 3) return; check(true); }                      // loop body IS the bare return: verifies cleanly
```

## Root cause

In `verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java`, `StaticMethodCollector.convertStatement` yields `null` for a valueless `JCReturn` (L392). That `null` is an intentional "drop this statement" signal, and four call sites honor it by filtering (`if (converted != null)`): method body (L259), `convertBlock` (L294), `extractLoopParts` (L514), `LetExpr` (L562).

Two sites instead pass the converted statement **directly into a Laurel node constructor with no null guard**:

- **`JCIf` then-branch** (L378 → L386 `ifThenElse(cond, thenBranch, elseB)`). Note the *else* branch immediately above (L382) *is* null-guarded — the then-branch isn't.
- **labeled non-loop statement** (L473 → L474 `labelledBlock(..., List.of(body), breakLbl)`).

The resulting null node then NPEs when `toIon` is called during serialization.

## Suggested fix

[...]

PR #449, opened 2026-06-15
Fix NPE on valueless return in if-then and labeled statements

kadirayk:
### What was changed?

Fixed a `NullPointerException` (#443) thrown during Laurel serialization when a method contains a valueless `return;` reached through control flow.

A valueless `return;` is converted to `null` in `convertStatement` as a "drop this statement" signal. Most call sites honor that contract by filtering out the `null`, but two sites passed it straight into a Laurel node constructor:

- the `JCIf` then-branch (`ifThenElse(...)`), and
- the non-loop labeled statement body (`labelledBlock(...)`).

The resulting `null` node then NPE'd when `toIon` was called during serialization. Both sites are now null-guarded by substituting an empty `block(...)`, mirroring the existing else-branch guard.

This affects valid, already-supported Java such as `if (cond) return;`, `lbl: return;`, and a bare `return;` inside a loop. Note: because Laurel cannot yet express a valueless return ([Strata#1353](https://github.com/strata-org/Strata/issues/1353)), the return is dropped rather than modeled as control flow.

Files:
- `JavaToLaurelCompiler.java` — null-guard the two node-bound sites.
- `VerifyBareReturn.java` — new regression tests.

### How has this been tested?

Added `VerifyBareReturn` regression tests covering the three previously-crashing shapes: `if (cond) return;` at method level, a labeled bare return, and a bare return inside a do-while loop. None existed before.

- The new tests pass (`./gradlew :verifier:test --tests "...VerifyBareReturn"`), confirming translation no longer crashes and the methods verify cleanly.
- The surrounding statement suites (`VerifyDoWhile`, `VerifyBreakContinue`, `VerifyStatements`, `Switches`, `SwitchDesugaring`, `AssertVsCheck`, `TranslationErrors`) still pass, confirming no regressions.

### DIFF
--- a/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
+++ b/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
@@ -433,6 +433,12 @@ public class JavaToLaurelCompiler {
                 case JCTree.JCIf ifStmt -> {
                     StmtExpr cond = convertExpression(ifStmt.cond, renames);
                     StmtExpr thenBranch = convertStatement(ifStmt.thenpart, renames);
+                    // convertStatement returns null for droppable statements (e.g. a
+                    // valueless `return;`). The Laurel node requires a non-null then-branch,
+                    // so substitute an empty block, mirroring the else-branch guard below.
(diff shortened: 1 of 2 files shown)


fabiomadge:
We already merged https://github.com/strata-org/Strata/pull/1354, so we should probably just use it instead.

PR #451, opened 2026-06-15
Guard getContractBlock against malformed method bodies

fabiomadge:
`getContractBlock`'s unguarded double `(JCBlock)` cast threw a `ClassCastException` on a non-nested-block body (Lower-generated / raw — instance methods, some enums). It escaped the `JavaViolationException`-only catch in `visitMethodDef` and aborted the whole run instead of skipping the one method.

Now guards the shape with `instanceof` and throws `JavaViolationException`, so the method skips gracefully. Prerequisite for #452. Suite green.


fabiomadge:
Superseded by #433, which landed the identical guard on `getContractBlock` (throw a catchable `JavaViolationException` on a non-contract-shaped body instead of letting a `ClassCastException` escape). #433's version is the better implementation — it extracts a reusable `hasContractStructure()` predicate and also guards the `extractLoopParts` caller side. #452 now picks up that guard from main directly; closing this as redundant.

PR #452, opened 2026-06-15
Translate instance methods via static-call-with-self encoding

### LINKED ISSUE #420: Non-static methods skipped by translator are silently counted as Verified
## Problem

`JavaToLaurelCompiler.visitMethodDef` at `verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java:198` gates all translation behind `(method.mods.flags & Flags.STATIC) != 0`. Non-static methods produce no Laurel output. The translator does **not** report a diagnostic, **not** call `markSkipped`, and **not** otherwise update the method's status.

Meanwhile `VerifyAnnotationCompiler.addMethodToIntervalTree` (at `…/simplifications/VerifyAnnotationCompiler.java:102-125`) adds every method (static or not) to the interval tree with status `Verified` by default (when `@Verify(true)`).

`LaurelDriver` only demotes a method's status from `Verified` to `Failed` when a Strata diagnostic lands on its source range (`LaurelDriver.java:201-207`). Since non-static methods never reach Strata, no diagnostic is emitted, and they remain `Verified`.

Net effect: `verifiedCount` includes non-static methods that were never actually translated or verified. The test annotation `@JVerifyTest(methodsVerified = N)` is satisfied trivially.

## Repro

`verifier/src/test/java/org/strata/jverify/verifier/tests/javasupport/statements/VerifyStatements.java` is annotated `methodsVerified = 13` and passes CI. All 13 methods are non-static. Running the test produces zero Laurel procedures (verified empirically — `temp.laurel` contains only the predefined int-range commands, no `procedureCommand` entries for any `VerifyStatements_*`).

## Scope of impact

Searched `verifier/src/test/java/org/strata/jverify/verifier/tests/`:

- `VerifyStatements.java`: 0 / 11 methods static — **fully vacuous**.
- Most files in `verification/`, `javasupport/lambdas/`, `javasupport/generics/`, `javasupport/contracts/`, `examples/` — partial or full non-static. Each such method is silently `Verified`.

Mixed-static files have partially honest counts; full-non-static files have completely vacuous counts.

## Why it matters

The verdict for the *static* methods is honest. The accounting summary is misleading. Concretely:

- Test passes don't prove what the test purports to prove.
- A real verifier regression on a non-static test would not surface as a failure.
- When PLAN.md Step 3 (issue #392) drops the `Flags.STATIC` gate, every previously-vacuous test starts translating for real — and any pre-existing translator/desugar bugs surface in the existing test corpus. Notably the do-while desugaring in PR #418 has a verifier-completeness bug with the sentinel-OR guard form that won't fire until then (#418 review thread for details).

## Proposed fix

Two parts:

1. **Fix the silent skip.** In the static-gate else branch at `JavaToLaurelCompiler.java:198`, call `annotationCompiler.markSkipped(currentCompilationUnit, method)`. Reuses the helper added by #416. One-line change.

2. **Sweep test annotations.** Every file with non-static methods will need its `methodsVerified` count reduced and `methodsSkipped` count introduced. ~25-30 files. Mostly mechanical via `JVERIFY_UPDATE_TEST_ANNOTATIONS=1` once #411 lands.

## Sequencing

Recommend doing this *after* Step 3 (#392) lands, because Step 3 dissolves the gap entirely (every method gets translated, no silent skips). Doing it before Step 3 is a temporary half-measure that requires the test-annotation sweep, only to be reverted when Step 3 makes everything `Verified` again.

**Fix order:**
- Land #392 (Step 3) — drops static gate, every method translates.
- Sweep test annotations to reflect honest counts (likely many tests start failing — those are real regressions to fix).
- Close this issue.

If we discover a non-static test producing wrong verdicts before Step 3 lands, revisit.

## Related

- #411 — `updateTestAnnotation` regex bug (needs to land first to make the sweep mechanical).
- #392 — instance method support (Step 3) — fixes the underlying gate.
- #416 — Step 1 safety net, introduces `markSkipped` helper used by the proposed fix.

fabiomadge:
Translate **instance methods** via static-call-with-self: drop the `Flags.STATIC` gate, add an explicit `self` first parameter, map `this` → `self`, prepend the receiver at call sites — all lowering to Laurel `.StaticCall` (never `.InstanceCall`/`.This`).

**Soundness.** `self` is an unconstrained, never-dereferenced opaque composite (instance-state access is refused), so it only strengthens obligations, never discharges a false postcondition. Calls are refused unless provably monomorphic (callee/class `final` or `private`): static mangling binds the *static* type, so an overridable call would verify the wrong contract. `PolyDispatchSoundness` guards this (bisection-proven); `InstanceMethodVerifies`/`InstanceMethodContractViolated` show real contracts verify and violations are caught.

**No false Verified.** A skipped member emits no Laurel but keeps its default `Verified` entry — so a skipped user constructor or anon/local method with a real contract would read `Verified` unchecked. These are now demoted to `Skipped` (implicit constructors excepted: no body, vacuously verified). Guarded by `SkippedConstructorContractNotVerified`.

**Refused (graceful skip):** constructors, anon/local classes, instance fields, polymorphic dispatch, allocation in a contract, overloads. A transitive-emittability fixpoint drops callers of un-emitted methods (across files) so no emitted call is unresolved.

**Mangling.** Names join class and method with `?` (illegal in Java identifiers), so mangling is injective — collisions are impossible and the old collision pass is gone (`MangledNameSeparation`).

**@Pure.** Emitted as an inlinable `function`, or an opaque `procedure` when it has a postcondition or constrained return (Strata rejects both on a `function`). A plain `@Unbounded` return stays a function; its result is usable directly (`PureCallResultUsedDirectly`) or via an `@Unbounded` local (`PureCallResultViaUnboundedLocal`).

**Tests.** Affected fixtures re-baselined to honest `methodsVerified`/`methodsSkipped` counts. `VerifyStatements` split (loop-invariant methods → skipped `VerifyStatementsLoops`, pending a Strata loop-elim fix). Plus `StaticOverloadRefusal`, `NewReceiverRefusal`, `NewInLoopInvariantRefusal`, `TransitiveSkip`, `CrossFileTransitiveSkip`.

**Workarounds to undo:** `grep -rn "TODO(strata-gap-" verifier/src/main/` → gap-1 (#1172 instance methods), gap-2 (#1173 function postconditions), gap-3 (#1174 dispatch). Constructor/anon translation and instance fields skip cleanly meanwhile.

Rebased onto `main` (#451 crash-guard superseded by #433, already on main). Clean build green, all modules.

Closes #420 — translator-skipped members count `Skipped`, not silently `Verified`. The broader Verified-by-default design stays tracked in #455.


kadirayk:
Nice Work. Soundness reasoning holds: the polymorphic-dispatch refusal is keyed correctly, `self` is inert with fields refused, and skipped members are now honestly `Skipped` instead of false `Verified`. LGTM.

**Scope:** this covers only stateless, monomorphically-reached instance methods. Field access, dispatch through overridable types, constructors, `super`/`new T().m()`, and overloads are all refused (graceful skip) — as intended for this step.

**Nit:** - `PureCallResultUsedDirectly.java`: broken `@link PureCallResultViaLocal` → `PureCallResultViaUnboundedLocal`.

kadirayk (approved):

PR #423, opened 2026-06-03
feat: add increment/decrement operator support (++x, --x, x++, x--)

olivier-aws:
## Summary

Adds support for Java prefix/postfix increment and decrement operators (`++x`, `--x`, `x++`, `x--`) by mapping them to the new Laurel `preIncr`/`preDecr`/`postIncr`/`postDecr` nodes introduced in [Strata PR #1320](https://github.com/strata-org/Strata/pull/1320).

The `EliminateIncrDecr` pass in Strata handles lowering to assignments, preserving Java semantics (prefix yields new value, postfix yields old value).

## Changes

- Update Strata submodule to `laurel-incr-decr` branch
- Add `PreIncr`/`PreDecr`/`PostIncr`/`PostDecr` records to `StmtExpr.java`
- Add factory methods to `Laurel.java`
- Handle `PREINC`/`PREDEC`/`POSTINC`/`POSTDEC` in `JavaToLaurelCompiler`

## Tests

Mirrors Strata's `T23_IncrDecr.lean` and `T23b_IncrDecrField.lean`:

- **IncrementDecrement.java**: statement form, pre/post expression semantics, repeated incr, mixed pre+post, for-loop step, if-condition, short-circuit, function argument
- **IncrementDecrementErrors.java**: intentionally failing assertions confirming correct error reporting
- **IncrementDecrementField.java**: field-based tests (skipped until class/instance support lands)

## Dependencies

Requires [Strata PR #1320](https://github.com/strata-org/Strata/pull/1320) to be merged first (submodule points to that branch).

PR #453, opened 2026-06-17
Bump Strata to 24b8f04e (source range fix for invariants), and update do/while bad invariant tests

### LINKED ISSUE #437: Inconsistency of source range for failing invariant checks in while and do-while
## Summary

When an invariant inside a `while` or `do-while` loop fails verification, we see two different behaviors of showing the source range.

### While

`JavaToLaurelCompiler.toSourceRange()` produces a zero-width source range for `while` loops because `TreeInfo.getEndPos()` returns `-1` for `JCWhileLoop` AST nodes. This causes verification error diagnostics to be reported at a single point rather than spanning the loop construct. So it's not possible to write matching test annotations using the caret (`^`) system.

```java
static void whileBadInvariant() {
    var x = 0;
    while (x < 10) {
        invariant(x >= 0 && x <= 5);
        x = x + 1;
    }
}
```

### Do-While

In this case, the error diagnostic spans the **entire** do-while statement. 

```java
    int x = -1;
    do { invariant(x >= 0); x = x + 1; } while (x < 5);
//  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Error: assertion could not be proved
}
```

## Design Question
What should be the correct way of informing the users about the failing invariants in both cases?

kadirayk:
### What was changed?

Bumped the Strata submodule from `e115b8ec` to `24b8f04e` (nightly, "Per-invariant source ranges for loop-invariant diagnostics", [#1361](https://github.com/strata-org/Strata/pull/1361)) and brought JVerify in line with the updated Laurel dialect.

- **Strata submodule**: updated pointer to `24b8f04e`.
- **`VerifyDoWhile` test**: updated the expected diagnostic for `doWhileBadInitialInvariant` to the new, tighter per-invariant range (points at the invariant condition rather than the whole do-while statement); added a negative regression case `whileBadInvariant` (invariant holds on entry but is not preserved by the loop body) with a matching diagnostic marker and doc comment; updated the `@JVerifyTest` counts to enforce the new expectations (`methodsInvalid = 3, errorCount = 3`).

This is not a user-visible behavior change for JVerify itself; it tracks the upstream Strata improvement to invariant diagnostic source ranges.

### How has this been tested?
- `./gradlew build` passes (compile + full test suite)
- `./gradlew :verifier:test --tests "*VerifyDoWhile*" --rerun-tasks` passes, exercising the updated and newly added invariant diagnostic cases.

Fixes #437

### DIFF
--- a/verifier/src/main/java/org/strata/jverify/laurel/Laurel.java
+++ b/verifier/src/main/java/org/strata/jverify/laurel/Laurel.java
@@ -40,6 +40,9 @@ public class Laurel {
     public static StmtExpr string(SourceRange sourceRange, java.lang.String s) { return new StmtExpr.String_(sourceRange, s); }
     public static StmtExpr string(java.lang.String s) { return new StmtExpr.String_(SourceRange.NONE, s); }
 
+    public static StmtExpr bvLiteral(SourceRange sourceRange, long value, long width) { if (value < 0) throw new IllegalArgumentException("value must be non-negative"); if (width < 0) throw new IllegalArgumentException("width must be non-negative"); return new StmtExpr.BvLiteral(sourceRange, java.math.BigInteger.valueOf(value), java.math.BigInteger.valueOf(width)); }
+    public static StmtExpr bvLiteral(long value, long width) { if (value < 0) throw new IllegalArgumentException("value must be non-negative"); if (width < 0) throw new IllegalArgumentException("width must be non-negative"); return new StmtExpr.BvLiteral(SourceRange.NONE, java.math.BigInteger.valueOf(value), java.math.BigInteger.valueOf(width)); }
+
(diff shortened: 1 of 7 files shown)


fabiomadge (approved):
Approving. Reviewed the full Strata delta: with #423 on main, this advances the pin by exactly one commit (`8abb4630` + #1361 = `24b8f04e`), plus the test updates. You've merged main in (head `49adb8b8`), so it's cleanly mergeable; CI green on both platforms.

Substance holds: the per-invariant ranges work end-to-end — the `^^^^^^` carets are enforced assertions (the harness compares full ranges, not just counts), so this genuinely tests #1361. `whileBadInvariant` is a good negative regression, and the new `methodsInvalid = 3` is it plus the two pre-existing failures. #428's StrataCLI wiring still holds.

PR #454, opened 2026-06-17
Emit valueless return; instead of dropping it

### LINKED ISSUE #443: NPE on `if (cond) return;` — null statement passed unguarded into a Laurel node

fabiomadge:
Closes #443.

### What was changed?

A bare `return;` was converted to `null` ("drop this statement"). Most consumers filtered it, but two crashed during serialization (#443) by passing it into a Laurel node constructor:

- the `JCIf` then-branch — `if (c) return;`
- the labeled non-loop body — `lbl: return;`

Where the null *was* filtered, the `return;` silently vanished, losing early-exit semantics — unsound (code after an early return became unconditionally reachable).

Strata now accepts a valueless return end-to-end (strata-org/Strata#1354; `Return none` lowers to a procedure-body exit), so this emits `return_(Optional.empty())` instead of null — fixing both the crash and the silent drop. No generated AST changes.

### How has this been tested?

- New `VerifyValuelessReturn`: the three #443 crash shapes, a soundness probe (early return guards the rest), and a trailing bare return. Fails without the fix (NPE), passes with it.
- `./gradlew :verifier:test` — full suite green.

### DIFF
--- a/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
+++ b/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
@@ -497,10 +497,10 @@ public class JavaToLaurelCompiler {
                     yield ifThenElse(toSourceRange(ifStmt), cond, thenBranch, elseB);
                 }
                 case JCTree.JCReturn retStmt -> {
-                    if (retStmt.expr != null) {
-                        yield return_(toSourceRange(retStmt), Optional.of(convertExpression(retStmt.expr, renames)));
-                    }
-                    yield null;
+                    Optional<StmtExpr> value = retStmt.expr != null
+                            ? Optional.of(convertExpression(retStmt.expr, renames))
+                            : Optional.empty();
+                    yield return_(toSourceRange(retStmt), value);
                 }
                 case JCTree.JCWhileLoop whileStmt -> {
                     var sr = toSourceRange(whileStmt);
--- a/verifier/src/test/java/org/strata/jverify/verifier/tests/javasupport/statements/VerifyValuelessReturn.java
+++ b/verifier/src/test/java/org/strata/jverify/verifier/tests/javasupport/statements/VerifyValuelessReturn.java
@@ -1,0 +1,47 @@
+package org.strata.jverify.verifier.tests.javasupport.statements;
+
+import org.strata.jverify.testengine.JVerifyTest;
+
+import static org.strata.jverify.JVerify.*;
+
+/** Valueless {@code return;} (#443): once dropped as {@code null}, so it crashed or lost early-exit semantics. */
+@SuppressWarnings({"ConstantValue"})
+@JVerifyTest(exitCode = 0, methodsVerified = 6, errorCount = 0)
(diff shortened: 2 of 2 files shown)


kadirayk (approved):
Nice minimal, root-cause fix.

`JCReturn` with no expr previously yielded `null`, which was causing:

- **Crashes** when `null` reached a node constructor: `if (c) return;` (`ifThenElse` then-branch) and `lbl: return;` (`List.of(null)` in `labelledBlock`).
- **Silent unsoundness** where `null` was filtered by `convertBlock`: the early-exit was dropped, making code after the return unconditionally reachable.

The fix always emits `return_(..., Optional.empty())` for valueless returns, relying on Strata's `Return none` lowering (already pinned via the submodule on `main`).

PR #459, opened 2026-06-25
JVerify: tolerate non-ClassSymbol owner chains in method-symbol resolution

tautschnig:
Some method invocations reach the front end with a symbol whose owner chain contains a non-ClassSymbol — typically record-component accessors of types declared inside an anonymous class, or built-in methods on synthetic Symtab entries (e.g. Symtab$6). `Symbol.outermostClass()` casts the owner chain up to ClassSymbol and throws ClassCastException on those, aborting translation of the whole source file.

Guard the three affected sites so such symbols degrade gracefully instead of crashing; for the common case (regular static methods on a top-level class) every path produces identical results:

- JVerifyUtils.getJVerifyMethod: the method-select symbol may not be a MethodSymbol at all — treat that as definitely-not-a-JVerify-call.
- JVerifyUtils.fromJVerify: catch the outermostClass() cast and return false (such a symbol cannot be the JVerify class).
- JavaToLaurelCompiler.qualifiedMethodName: catch the cast and fall back to the immediate owner's name.

### DIFF
--- a/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
+++ b/verifier/src/main/java/org/strata/jverify/verifier/compiler/generator/laurel/JavaToLaurelCompiler.java
@@ -227,11 +227,28 @@ public class JavaToLaurelCompiler {
     }
 
     private static String qualifiedMethodName(Symbol.MethodSymbol sym) {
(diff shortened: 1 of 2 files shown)


keyboardDrummer (approved):
Nice, thanks!

